PR #24, opened 2020-02-19
Fix/dockerimmugw

mmeloni:
In dockerfile is missing immugw.
Remove useless host param

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,11 @@
 FROM golang:1.13-stretch as build
 WORKDIR /src
 COPY . .
-RUN GOOS=linux GOARCH=amd64 make immu-static immud-static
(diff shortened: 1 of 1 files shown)


cmd/immugw/immugw.go
```diff
@@ -124,6 +119,6 @@ func serve(cmd *cobra.Command, args []string) error {
 			logger.Infof(fmt.Sprintf("Immudb is listening at %s:%s", immudhost, immudport))
 		}
 	}
-	logger.Infof("Starting immugw at %s:%s", host, port)
-	return http.ListenAndServe(host+":"+port, handler)
```
leogr: Why do we want to prevent the user from setting the "host" value?

Issue #34, opened 2020-03-27
Golang SDK immudb


mmeloni:
Actually immugw rest proxy exposes simplify api that wrap immudb one's.
All safe methods exposes by immugw (SafeSet, SafeGet, SafeReference, SafeZAdd) doesn't require for root and index because this data is managed internally.
The purpose of this task is to create a golang SDK to offer the same capabilities but in pure GO.
It should be fine that immugw wil be upgraded in order to use the newly sdk.
In order to do this you could use a new .proto schema that extends https://github.com/codenotary/immudb/blob/master/pkg/api/schema/schema.proto 
Ex:
actual SafeGet service in immudb schema.proto
```
message SafeGetOptions {
	Key key = 1;
	Index rootIndex = 2;
}
rpc SafeGet(SafeGetOptions) returns (SafeItem){
		option (google.api.http) = {
            post: "/v1/immurestproxy/item/safe/get"
            body: "*"
        };
	};
```
New SafeSet in immu.proto
```
rpc SafeGet(Key) returns (VeridiedItem){};
```
eg: VerifiedItem contains item and a bool in which we put verification results.

Flow:
- The sdk will intercept SafeGet,
- will retrieve rootIndex from client.RootService package
- will forward the call to immudb (with SafeGet(SafeGetOptions) )
- will make checks on result
- will return to the user the VerifiedItem

padurean:
Merged to master.

PR #40, opened 2020-03-31
#34 - Go SDK - Add SafeGet, SafeSet, SafeReference, SafeAdd, Backup and Restore

padurean:


pkg/client/client.go
```diff
@@ -86,6 +89,56 @@ func (c *ImmuClient) Get(keyReader io.Reader) (*schema.Item, error) {
...
+	var metadata runtime.ServerMetadata
+	safeItem, err := c.serviceClient.SafeGet(
+		context.Background(),
+		opts,
+		grpc.Header(&metadata.HeaderMD),
+		grpc.Trailer(&metadata.TrailerMD))
+
+	verified := safeItem.Proof.Verify(safeItem.Item.Hash(), *root)
+	if verified {
+		// saving a fresh root
```
padurean: @mmeloni Do we need to do this or not?

pkg/client/client.go
```diff
@@ -86,6 +89,56 @@ func (c *ImmuClient) Get(keyReader io.Reader) (*schema.Item, error) {
 	return result, err
 }
 
+// SafeGet ...
+func (c *ImmuClient) SafeGet(keyReader io.Reader) (*schema.Item, bool, error) {
```
padurean: @mmeloni Do you think it is ok to return an `Item` or should we create a dedicated struct (for simple Get an `Item` is returned, so probably if we do not want to expose it we need to change everywhere in this file).

pkg/client/client.go
```diff
@@ -86,6 +91,70 @@ func (c *ImmuClient) Get(keyReader io.Reader) (*schema.Item, error) {
...
+	start := time.Now()
+	if !c.isConnected() {
+		return nil, ErrNotConnected
+	}
+	key, err := ioutil.ReadAll(keyReader)
+	if err != nil {
+		return nil, err
+	}
+
+	rs := NewRootService(c.serviceClient, cache.NewFileCache())
```
mmeloni, line 113: Can you use dependency injection for RootService?

pkg/client/client.go
```diff
@@ -245,6 +403,71 @@ func (c *ImmuClient) Reference(keyReader io.Reader, valueReader io.Reader) (*sch
...
+	root, err := rs.GetRoot(context.Background())
+	if err != nil {
+		return nil, err
+	}
+
+	opts := &schema.SafeReferenceOptions{
+		Ro: &schema.ReferenceOptions{
+			Reference: &schema.Key{Key: reference},
```
mmeloni, line 429: In the future We should semplify this model inside .proto file and use a more simple
```
message ReferenceOptions {
	bytes reference = 1;
	bytes key = 2;
}
```

pkg/client/client.go
padurean: @mmeloni Please see my last commit : https://github.com/codenotary/immudb/pull/40/commits/fb49cb630987c7e5178f8b83ee146fdf6cedb461

and let me know what you think. Thanks.

mmeloni: ok!

PR #45, opened 2020-04-01
Fix/immugw handling conn

mmeloni:


cmd/immugw/immugw.go
```diff
@@ -32,17 +14,18 @@ func main() {
 		Use:   "immugw",
 		Short: "Immu gateway",
 		Long:  `Immu gateway is an intelligent proxy for immudb. It expose all gRPC methods with a rest interface and wrap SAFESET and SAFEGET endpoints with a verification service.`,
-		Run: func(cmd *cobra.Command, args []string) {
-			serve(cmd, args)
+		RunE: func(cmd *cobra.Command, args []string) error {
```
padurean, line 17: @mmeloni is the trailing `E` (i.e. `RunE`) intentional?

mmeloni: Yes. I would like to use same method than cmd/immud

PR #47, opened 2020-04-03
implemented issue#25, added embedded swagger ui to immugw

nredko:


Makefile
```diff
@@ -10,6 +10,9 @@ STRIP = strip
 .PHONY: all
 all: immu immud immugw
 
+.PHONE: rebuild
```
padurean, line 13: Isn't this a typo? I suppose it should be `.PHONY`

cmd/immu/immu.go
```diff
@@ -96,7 +97,11 @@ func main() {
 				if err != nil {
 					return err
 				}
-				printItem([]byte(args[0]), value, response)
+
+				response, err := immuClient.Connected(func() (interface{}, error) {
```
padurean, line 101: Just to confirm/double-check: this was added intentionally, so that `set` will return the complete item (with `get`), not just the index returned by `Set(SV)` (as it was before), right?

Makefile
nredko: fixed

cmd/immu/immu.go
nredko: Since timestamp is generated inside the Client, but according requirements Set methods should return only index. So in immu CLI we should perform Get after Set to read generated timestamp.

pkg/client/convert.go
```diff
@@ -0,0 +1,64 @@
...
+		Index: item.Index,
+		Key:   item.Key,
+		Value: &schema.Content{
+			Payload:   payload,
+			Timestamp: ts,
+		},
+	}
+}
+
+func ToKV(skv *schema.StructuredKeyValue) *schema.KeyValue {
```
mmeloni, line 59: In order to get an elegant usage of ToSItem and ToKV we have to do something similar
```
result := item.ToSItem()

func (i *Item)ToSItem() *StructuredItem {
***
}
```

nredko:
> * All SV  methods need to be removed in order to avoid useless duplication

@mmeloni In my understanding there was requirement to keep RPC server as-is (so gRPC clients/drivers will process timestamps as part of bytestream) and implement separate methods for GW to allow REST clients to use structured JSON values with timestamp as separate field. 

So I've made (Safe)GetSV/SetSV methods to be handled by immugw REST server, and (Safe)Get/Set to be handled by RPC server. 

Please correct me if I'm wrong and immugw also should receive timestamp as part of payload bytestream.
Thanks.

mmeloni:
Thanks @nredko I have understand now. Ok on my side

mmeloni:
>clients/drivers will process timestamps as part of bytestream) and implement separate methods for GW to allow REST clients to use structured JSON values with timestamp as separate field.

Thanks @nredko I have understand now. Ok on my side

nredko:
@mmeloni please review.

go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/prometheus/client_golang v1.2.1
+	github.com/rakyll/statik v0.1.7
```
mmeloni, line 15: If this is a dependency of swagger-ui please remove it.

nredko:
@mmeloni @padurean Guys, please let me know correct scenario to test new functions of immu CLI.

My scenario (for empty DB):
```
./immu set 1 1
 - added OK, index 0
./immu safeset 22 22
 - returns 'verified: false'
./immu get 22
 - result is OK
./immu safeget 22
 - returns 'verified: false'
```
Another scenario:
```
./immu set 1 1
 - added OK, index 0
./immu set 2 2
 - added OK, index 1
./immu safeset 22 22
 - returns 'verified: false'
./immu get 22
 - result is OK
./immu safeget 22
 - returns 'verified: false'
```

In previous build I was testing "safe" functions with immugw, it required "root index" and worked well.

nredko:
Arghhhh. I've spent about 2 days until found that file `.root` should be removed as well as `./immudb` folder.

nredko:
@mmelony, please review. Note that tests dont work due to this comment from @padurean 
https://github.com/codenotary/immudb/blob/ab59216ae3b37314b8e4e6abb5b3172c69bd9aa9/pkg/client/client_test.go#L211

```
func TestRestore(t *testing.T) {
	cleanup()
	defer cleanup()

	ctx := context.Background()

	// this only succeeds if only this test function is run
```

padurean:
@nredko Tests work fine on the master branch. That comment refers only to the following commented lines: it's those commented lines that don't work, but they are commented, so they can't be the cause of failing.

nredko:
@padurean @mmeloni Anyway debug/restore tests should be updated since now DB contains timestamps inside records value. As timestamps are generated each time when record is created - we can't compare with static old backup file. I guess it's job more to developer of backup feature, than to me.

mmeloni:
> @padurean @mmeloni Anyway debug/restore tests should be updated since now DB contains timestamps inside records value. As timestamps are generated each time when record is created - we can't compare with static old backup file. I guess it's job more to developer of backup feature, than to me.

Thanks @nredko . In order to test restore functionality client should have a dependency on a timestamp generator service, completely decoupled by client. In this way you can fake timestamp inside tests.

PR #49, opened 2020-04-04
Closes #34 - Add native Go SDK

### LINKED ISSUE #34: Golang SDK immudb

padurean:
Also add safeget, safeset, safereference, safezadd, backup and restore commands to CLI


pkg/client/client.go
```diff
@@ -267,12 +447,245 @@ func (c *ImmuClient) ZAdd(setReader io.Reader, score float64, keyReader io.Reade
...
+
+	bkpClient, err := c.serviceClient.Backup(ctx, &empty.Empty{})
+	if err != nil {
+		return counter, err
+	}
+
+	var offset int64
+	var errs []string
+	for {
+		kvList, err := bkpClient.Recv()
```
mmeloni, line 570: You need to close stream.
Take a look at https://grpc.io/docs/tutorials/basic/go/
stream.SendAndClose

pkg/client/client_test.go
```diff
@@ -0,0 +1,241 @@
+package client
```
mmeloni, line 1: add License

pkg/client/client.go
```diff
@@ -167,98 +310,135 @@ func (c *ImmuClient) SetBatch(request *BatchRequest) (*schema.Index, error) {
...
 		Key:       &schema.Key{Key: key},
 	})
 	c.Logger.Debugf("reference finished in %s", time.Since(start))
 	return result, err
 }
 
-func (c *ImmuClient) ZAdd(setReader io.Reader, score float64, keyReader io.Reader) (*schema.Index, error) {
+// SafeReference ...
+func (c *ImmuClient) SafeReference(ctx context.Context, reference []byte, key []byte) (*VerifiedIndex, error) {
 	start := time.Now()
```
mmeloni, line 381: All safe methods need to be thread safe.
Use mutex to lock immuclient properly, and release lock when methods return

padurean:
@mmeloni I've implemented all the changes/fixes requested. Thanks and please let me know if you notice anything else.

padurean:
@mmeloni beside the failing check on linux (which doesn’t seem to be related to code), there’s one merge commit that needs to be cleared from the log before merging it into `master` branch.

PR #67, opened 2020-04-08
Chore/toml2ini

mmeloni:


pkg/gw/server.go
```diff
@@ -167,8 +167,7 @@ func (s *ImmuGwServer) Start() error {
 	}
 
 	s.installShutdownHandler()
-	s.Logger.Infof("Starting immugw at %s:%d", s.Options.Address, s.Options.Port)
-
+	s.Logger.Infof("starting immudgw: %v", s.Options)
```
padurean, line 170: @mmeloni i think there’s a typo here: shouldn’t it be “immugw” instead of “immuDgw”?

mmeloni: Ty

PR #72, opened 2020-04-09
feat: add token based authentication

padurean:


pkg/api/schema/schema.swagger.json
```diff
@@ -1,9 +1,8 @@
 {
   "swagger": "2.0",
-  "host": "127.0.0.1:8081",
```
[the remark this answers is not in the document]
padurean: I had to manually add it back together with version (generation doesn't seem to add those).

padurean:
@mmeloni I've implemented all the changes you've requested, except this part (which i intend do tomorrow):

> We have a big problem related to the fact that a secure tls communication is mandatory for security reason. So we can use login method and tokens only if --mtls flag is provided. If not anybody who can listen on our communication can know sensitive infos.
For debug purpose it's better to add a configuration option to enable token check.

cmd/immu/immu.go
```diff
@@ -750,33 +754,52 @@ secondRoot: %x at index: %d
...
+		return err
+	}
 	viper.SetDefault("default.port", gw.DefaultOptions().ImmudPort)
 	viper.SetDefault("default.address", gw.DefaultOptions().ImmudAddress)
+	viper.SetDefault("default.auth", client.DefaultOptions().Auth)
 	return nil
 }
 
-func options(cmd *cobra.Command) (*client.Options, error) {
+func options(cmd *cobra.Command, withToken bool) (*client.Options, error) {
```
mmeloni, line 781: Why here we need to add withToken? 
Simply if auth is enabled we could check for the token, without the boolean.

pkg/api/schema/schema.proto
```diff
@@ -204,7 +241,10 @@ service ImmuService {
 	rpc CurrentRoot(google.protobuf.Empty) returns (Root) {
 		option (google.api.http) = {
             get: "/v1/immurestproxy/root"
-        };
+				};
+		option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
```
mmeloni, line 245: If there no a specific reason to let public CurrentRoot please remove public access

pkg/auth/types.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+func (t TokenAuth) GetRequestMetadata(ctx context.Context, in ...string) (map[string]string, error) {
+	return map[string]string{
+		"authorization": "Bearer " + t.Token,
+	}, nil
+}
+
+func (TokenAuth) RequireTransportSecurity() bool {
+	// TODO OGG: set this to true, unless config says is ok to return false (i.e. for testing)
```
mmeloni, line 32: You forgot to switch this?

pkg/gw/options.go
```diff
@@ -41,9 +42,10 @@ func DefaultOptions() Options {
 		ImmudAddress: "127.0.0.1",
 		ImmudPort:    3322,
 		MTLs:         false,
-		Config:       "configs/immu.ini",
-		Pidfile:      "",
-		Logfile:      "",
+		// TODO OGG: check with Michele: shouldn' this be immugw.ini?
+		Config:  "configs/immu.ini",
```
mmeloni, line 46: You are right

cmd/immu/immu.go
padurean: @mmeloni The `withToken` value differs for each command as there are some commands which do not require authentication, like `login`, for example.

pkg/api/schema/schema.proto
padurean: @mmeloni The reason `CurrentRoot` is always without auth, is because it seems to be called during `Connected` => since even `Login` is wrapped in a `Connected` call, it would make it impossible to login.

pkg/auth/types.go
padurean: @mmeloni Not really forgot, but it is a bit tricky to set it to `true` since in the case when both client and server run on the same machine authentication should work even without MTLS.

cmd/immu/immu.go
mmeloni: But if login command send a token to the server token is simply ignored

padurean: @mmeloni Ok, i removed it.

PR #95, opened 2020-04-14
Chore/flush size cfg

mmeloni:


cmd/immudb/immudb.go
```diff
@@ -51,7 +51,9 @@ Environment variables:
   IMMUDB_AUTH=false
   IMMUDB_PKEY=./tools/mtls/3_application/private/localhost.key.pem
   IMMUDB_CERTIFICATE=./tools/mtls/3_application/certs/localhost.cert.pem
-  IMMUDB_CLIENTCAS=./tools/mtls/2_intermediate/certs/ca-chain.cert.pem`,
+  IMMUDB_CLIENTCAS=./tools/mtls/2_intermediate/certs/ca-chain.cert.pem
+  IMMUDB_EVENTUAL_CONSISTENCY=true
+  IMMUDB_CACHE_SIZE=true`,
```
padurean, line 56: @mmeloni shouldn't this be an integer number instead of bool?

cmd/immudb/immudb.go
```diff
@@ -140,6 +146,8 @@ func setupFlags(cmd *cobra.Command, options server.Options, mtlsOptions server.M
 	cmd.Flags().StringP("address", "a", options.Address, "bind address")
 	cmd.Flags().StringP("dbname", "n", options.DbName, "db name")
 	cmd.Flags().StringVar(&o.CfgFn, "config", "", "config file (default path are configs or $HOME. Default filename is immudb.ini)")
+	cmd.Flags().Bool("eventual-consistency", options.EventualConsistency, "enable/disable eventual consistency")
+	cmd.Flags().Uint64("cache-size", options.CacheSize, "enable/disable eventual consistency")
```
padurean, line 150: @mmeloni description here looks the same as for the enable/disable config item, i suppose it should be something like _maxcache size in MB_

configs/immudb.ini.dist
```diff
@@ -1,12 +1,14 @@
-dir = .
+dir = ./immudata
 network = tcp
 address = 127.0.0.1
 port = 3322
-dbname = immudata
+dbname = immudb
```
padurean, line 5: @mmeloni is this ok? i thought it should also be `immudata`

padurean: @mmeloni i just saw you've added an intermediary directoy named `immudata`, so i suppose it's ok.

mmeloni: Yes the point is that actually dbname is  the name of the db istance. The folder in which data is written is dir. Internally it uses dir+dbname to build the complete path of data folder. @andrei-lihu it's ok?

PR #114, opened 2020-04-16
Insertion order index

mmeloni:


pkg/store/store.go
```diff
@@ -296,30 +300,59 @@ func (t *Store) Count(prefix schema.KeyPrefix) (count *schema.ItemsCount, err er
...
-
-	stream.Send = func(list *pb.KVList) error {
-		for _, kv := range list.Kv {
-			key = kv.Key
-			value = kv.Value
-			version = kv.Version
-			found = true
+	var refkey []byte
+	// cache reference lookup
+	if key := t.tree.rcaches[0].Get(readTs - 1); key != nil {
```
leogr, line 311: `t.tree` should be read-locked here, shouldn't?

pkg/store/store.go
```diff
@@ -296,30 +300,59 @@ func (t *Store) Count(prefix schema.KeyPrefix) (count *schema.ItemsCount, err er
 	return
 }
 
-// fixme(leogr): need to be optmized
-func (t *Store) itemAt(readTs uint64) (version uint64, key, value []byte, err error) {
-	stream := t.db.NewStreamAt(readTs)
-	stream.ChooseKey = func(item *badger.Item) bool {
-		return item.UserMeta() != bitTreeEntry && item.Version() == readTs
+func (t *Store) itemAt(readTs uint64) (index uint64, key, value []byte, err error) {
+	if readTs > t.tree.w {
```
leogr, line 304: This seems inconsistent to me. 
`t.tree.w` represents the width of the computed Merkel tree, not the last valid item index. For a short time frame, newly inserted items can be already stored in the DB although their node in the Merkel tree have not been computed yet.

Anyhow, if you used `t.tree.w` I believe `t.tree` should be read-locked at least.

pkg/store/store.go
mmeloni: Tree is now locked

pkg/store/store.go
mmeloni: I remove the guard.
Method return all items currently stored. When the iterator miss the first element a not found error is returned.

mmeloni:
closed with #119

PR #116, opened 2020-04-17
WiP feat: detect db file changes done by processes other than immudb - Closes #79

### LINKED ISSUE #79: lock on the db files
It should not be possible to edit the DB files. We need a lock on them.

We also need a security mechanism to detect that data is not lost. Check with mmap. Also be aware of NFS.

padurean:


padurean:
This feature has been parked for the moment.
Work that has been done is kept in this branch: https://github.com/codenotary/immudb/tree/feat/db-files-tampering-detection

PR #127, opened 2020-04-22
Metrics generation and visualisation in immuadmin client - Closes #121

### LINKED ISSUE #121: immdb metrics generation
The following metrics should be available for immudb:
- performance KPIs: average query time, latency, duration
- status (health-check): listener up, badger up, and all statistics from underlying badger
- db size
- number of records  
- number of clients connected to the db
- last client connected
- data status: 
   - integrity check (this is the output produced by the integrity detector)
   - consistency check

padurean:


cmd/immuadmin/immuadmin.go
```diff
@@ -0,0 +1,220 @@
...
+			WithServername(servername).
+			WithCertificate(certificate).
+			WithPkey(pkey).
+			WithClientCAs(clientcas)
+	}
+	return options
+}
+
+func optionsWithAuth() *client.Options {
+	var token string
```
mmeloni, line 201: I believe this can be avoid thanks to the new constructor available in client: client.NewImmuClient(options())

pkg/server/server.go
```diff
@@ -123,6 +140,16 @@ func (s *ImmuServer) Start() error {
...
 	s.GrpcServer = grpc.NewServer(options...)
 	schema.RegisterImmuServiceServer(s.GrpcServer, s)
+	//===> !NOTE: See Histograms section here:
+	// https://github.com/grpc-ecosystem/go-grpc-prometheus
+	// TL;DR:
+	// Prometheus histograms are a great way to measure latency distributions of
+	// your RPCs. However, since it is bad practice to have metrics of high
+	// cardinality the latency monitoring metrics are disabled by default. To
+	// enable them the following has to be called during initialization code:
+	grpc_prometheus.EnableHandlingTimeHistogram()
```
mmeloni, line 150: IMHO this has to be enable by config flag

pkg/server/server.go
```diff
@@ -114,7 +126,12 @@ func (s *ImmuServer) Start() error {
 		return err
 	}
 
-	metricsServer := StartMetrics(s.Options.MetricsBind(), s.Logger)
+	metricsServer := StartMetrics(
+		s.Options.MetricsBind(),
+		s.Logger,
+		func() float64 { return float64(s.Store.CountAll()) },
+		func() float64 { return time.Since(startedAt).Hours() },
```
mmeloni, line 133: is possible to use an inline time.Now() and avoid to use a var?

cmd/immuadmin/statisticscmd/cmd.go
```diff
@@ -0,0 +1,160 @@
...
+	return runUI(newMetricsLoader(metricsURL(serverAddress)))
+}
+
+func NewCommand(optionsFunc func() *client.Options, immuClient *client.ImmuClient) *cobra.Command {
+	cmd := cobra.Command{
+		Use:     "statistics",
+		Short:   fmt.Sprintf("Show statistics"),
+		Aliases: []string{"s"},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			//--> workaround to achieve auth (this command does HTTP requests which do not go through ImmuClient)
```
mmeloni, line 120: metric server should not protected by auth. This maybe is useless

padurean:
@mmeloni Thanks for the review!

- Where do you think we should put a more comprehensive message regarding visual statistics option?
- I am not sure why/how `scanln`would prevent exposing sensitive data, can you please explain? Also, i do not know how to achieve that since the args are extracted by the `cobra` library. Also, the error you've shown seems unavoidable, unless the user uses single quotes around the password OR if we change the password generation to not include special characters anymore (i.e. so that password contains just alphanumeric characters).

mmeloni:
> @mmeloni Thanks for the review!
> 
> * Where do you think we should put a more comprehensive message regarding visual statistics option?
> * I am not sure why/how `scanln`would prevent exposing sensitive data, can you please explain? Also, i do not know how to achieve that since the args are extracted by the `cobra` library. Also, the error you've shown seems unavoidable, unless the user uses single quotes around the password OR if we change the password generation to not include special characters anymore (i.e. so that password contains just alphanumeric characters).

This are only comments, not change requests, but it can be helpful when we will implement new users auth. scanln read a line until it found \n char . Please see vcn in qhich password is prompted https://github.com/vchain-us/vcn/blob/4037b956ac216c4d8d85e962b1b6e79b9f093b9b/pkg/cmd/internal/cli/credentials.go

pkg/server/server.go
padurean: @mmeloni the `startedAt` value needs to be set only once: when server was started. If a call to `time.Now()` would be used there, the duration returned by `time.Since` will always be almost zero (from `time.Since` godoc: It is shorthand for `time.Now().Sub(t)`, so it will basically compare now with now).

padurean:
@mmeloni i've fixed all your requests. Regarding the graphical statistics view, i've made it the default mode and i've replaced the `-v` option with a `-t` one which will show the text version.

PR #149, opened 2020-05-02
refactor: remove immuclient dependency from root level command

mmeloni:


cmd/immuadmin/commands/stats.go
```diff
@@ -48,12 +49,23 @@ func (cl *commandline) stats(cmd *cobra.Command) {
 		Use:     "stats",
 		Short:   fmt.Sprintf("Show statistics as text or visually with the '-v' option. Run 'immuadmin stats -h' for details."),
 		Aliases: []string{"s"},
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) (err error) {
+			if cl.immuClient, err = client.NewImmuClient(options()); err != nil {
```
padurean, line 53: @mmeloni Since almost all sub-commands have the same implementation for pre- and post- run functions, i suppose their implementation could be extracted in a single place and referenced in all sub-commands(?)

mmeloni: Better now?

PR #150, opened 2020-05-02
Feat/immuadmin service installer

mmeloni:


cmd/immuadmin/immuadmin.go
```diff
@@ -45,8 +45,9 @@ Environment variables:
   IMMUADMIN_ADDRESS=127.0.0.1
   IMMUADMIN_PORT=3322
   IMMUADMIN_MTLS=true`,
+		SilenceUsage:      true,
```
padurean, line 48: @mmeloni I suppose this is silencing the usage for all commands, not just for `service`? If so, maybe we can set it dynamically, in the `service` sub-command only?

cmd/immuadmin/immuadmin.go
```diff
@@ -45,8 +45,9 @@ Environment variables:
...
+		SilenceErrors:     true,
```
padurean, line 49: @mmeloni Same remark as dor `SilenceUsage`: we should probably set it dynamically if it’s possible.

cmd/immuadmin/immuadmin.go
mmeloni: Removed option in service subcommand. The main one is persisted also in subocmmands

PR #157, opened 2020-05-06
fix immugw immud services in windows os

mmeloni:


cmd/immuadmin/commands/service.go
```diff
@@ -116,62 +97,70 @@ sudo ./immuadmin service immudb uninstall
...
+			if args[1] == "install" {
+				if execPath, err = service.CopyExecInOsDefault(localFile); err != nil {
 					return err
 				}
 			}
 
-			remove, _ := cmd.Flags().GetBool("remove-files")
-			if remove {
-				if err = os.Remove(localFile); err != nil {
+			// todo remove all involved files
```
padurean, line 125: @mmeloni Is this TODO still valid?

mmeloni: Yes, it make no sense this flag in my opinion, uninstall already do everything

mmeloni:
IMHO is ready now

PR #183, opened 2020-05-11
Store auth tokens in user home dir by default and other auth enhancements

padurean:


pkg/common/file.go
```diff
@@ -0,0 +1,70 @@
+package common
```
mmeloni, line 1: I would avoid to use common package name. Often the purpose of package with poor names become difficult to understand. I would move this file inside client package in a file called homedir.go.

PR #208, opened 2020-05-16
immuadmin - add "set auth|mtls" and "user list" subcommands

### LINKED ISSUE #200: immuadmin should be able to enable/disable auth on the immudb server
- applies for `immuclient` (`immuadmin` always has auth enabled, implicitly)
- for the moment it should have 2 options: `none` and `password` (in the future  3rd option will be implemented: `cryptosig`(?) - i.e. cryptographic signatures)

### LINKED ISSUE #201: >immuadmin user list
New immuadmin command that returns the list of immudb users and their permission (admin, read/write, read)

### LINKED ISSUE #202: > immuadmin set mtls true|false
New command to set mtls on/off from immuadmin

padurean:
closes #200 
closes #201 
closes #202


pkg/server/server.go
```diff
@@ -203,7 +206,7 @@ func (s *ImmuServer) Stop() error {
 }
 
 func (s *ImmuServer) Login(ctx context.Context, r *schema.LoginRequest) (*schema.LoginResponse, error) {
-	if !s.Options.Auth && !auth.IsAdminClient(ctx) {
```
mmeloni, line 206: old options.Auth is useless now?

padurean: @mmeloni No, it is not useless: `auth.AuthEnabled` is set with the value from `s.Options.Auth` during server start. The "dynamic" update (done from `immuadmin`, while server is running is updating both, so it doesn't really matter which one of them is checked here).

PR #210, opened 2020-05-17
Feat/trust checker

mmeloni:


mmeloni:
> @mmeloni Looks good. Can you please let me know if you think i should reuse code from the `trustchecker` in the `inmugw` auditor that i am writing? I suppose i could use the new server and store function `ByRawSafeIndex`?

Feel free to reuse code you need. ByRawSafeIndex is needed by trustchecker in order to retrieve elements that could be structured or not structured. If I correctly understand your task  you have to check only current root on a past root. Then  the consistency and currentRoot methods should be enough

pkg/tc/server.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	<-s.Done
+	s.Logger.Infof("Trust checker stopped")
+
+	return err
+}
+
+// Stop shutdown trust checker and status http server
+func (s *ImmuTcServer) Stop() error {
```
volknanebo, line 108: redunant return

pkg/tc/server.go
```diff
@@ -0,0 +1,150 @@
...
+		status["status"] = s.Checker.GetStatus(r.Context())
+		if err := json.NewEncoder(w).Encode(status); err != nil {
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+		return
+	})
+
+	return &http.Server{
+		Addr:         s.Options.Address + ":" + strconv.Itoa(s.Options.Port),
```
volknanebo, line 128: Can be rewritten with fmt.Sprintf instead of converting string and manually parsing .

pkg/tc/options.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// WithLogfile sets logfile
+func (o Options) WithLogfile(logfile string) Options {
+	o.Logfile = logfile
+	return o
+}
+
+func (o Options) Bind() string {
+	return o.Address + ":" + strconv.Itoa(o.Port)
```
volknanebo, line 116: Can be replaced with fmt.Sprintf instead of converting string and manually parsing the address.

pkg/store/safe.go
```diff
@@ -286,3 +286,43 @@ func (t *Store) SafeZAdd(options schema.SafeZAddOptions) (proof *schema.Proof, e
...
+
+func (t *Store) BySafeIndex(options schema.SafeIndexOptions) (safeitem *schema.SafeItem, err error) {
+
+	var item *schema.Item
+
+	idx, key, value, err := t.itemAt(options.Index + 1)
+	if err != nil {
+		return nil, err
+	}
+	if err == nil {
```
volknanebo, line 298: needless check.

pkg/tc/checker.go
```diff
@@ -0,0 +1,119 @@
...
+	s.Quit = true
+}
+
+// GetStatus return status of the trust checker. False means that a consistency checks was failed
+func (s *immuTc) GetStatus(ctx context.Context) bool {
+	return s.Trusted
+}
+
+func (s *immuTc) checkLevel0(ctx context.Context) (err error) {
+	for ok := true; ok; ok = !s.Quit {
```
volknanebo, line 69: you can make s.Quit channel and make a select loop it with time.Tick(10*time.Second). So in any quit case you can remove this 10 seconds barrier.

PR #234, opened 2020-05-19
immugw auditor

### LINKED ISSUE #197: immugw runs consistency check
immugw needs to be able to audit immudb for consistency check and report when data tampering has been identified.  

for immugw to be able to report the results of its trust audit it needs to provide a Prometheus exporter that can be scraped to run alerts 

Please align with Mert and Michele for the exporter to be consistent on all components

padurean:
~Still to do:~ **DONE**:

- [x] populate metrics

Closes #197


pkg/trustchecker/trustchecker.go
```diff
@@ -0,0 +1,302 @@
+/*
```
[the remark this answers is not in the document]
padurean: @mmeloni I don't understand what exactly i should do. Should i move `pkg/gw/trustchecker.go` go to the `pkg/tc` package?

development-lab:
I need code to build my own apps

pkg/auth/auth.go
```diff
@@ -300,6 +300,9 @@ func IsAdminClient(ctx context.Context) bool {
 var AdminUserExists func(ctx context.Context) bool
 var CreateAdminUser func(ctx context.Context) (string, string, error)
 
+var ErrServerAuthDisabled = status.Errorf(
```
volknanebo, line 303: formatting function at non formatting case , please use formatting string.

pkg/gw/options.go
```diff
@@ -122,6 +157,11 @@ func (o Options) Bind() string {
 	return o.Address + ":" + strconv.Itoa(o.Port)
 }
 
+// MetricsBind return metrics bind address
+func (o Options) MetricsBind() string {
+	return o.Address + ":" + strconv.Itoa(o.MetricsPort)
```
volknanebo, line 162: please do not format address manually , instead use Sprintf

pkg/auth/auth.go
padurean: @volknanebo i replaced `Errorf` with `Error` - is supposed that's what you meant?

volknanebo: +

pkg/gw/options.go
padurean: @volknanebo i blindly copied from `Bind()` :) I've changed it to `Sprintf` in both places.

pkg/trustchecker/trustchecker.go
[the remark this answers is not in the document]
padurean: @mmeloni i've created a dedicated package because i thought it should be used in more places, not just `immugw` (e.g. maybe also in `immuclient`)

leogr:
Just a thought.
I'm not sure the term "trust" reflects what is happening here.
IIRC, RFC 6962 never uses the term in this way. The security model is based on the fact that the logger (ie. the DB) is untrusted and must be audited... and I was confused when I read that a trust-checker is auditing an untrusted thing.

Just my 2 cents

vchaindz:
this is just one of many components that create a trust in combination with each other. There will be one on immudb, one on immugw and one on immuclient. Each checking the consistency until the level that makes sense (its not only about the merkle tree consistency, but also data consistency). Eventually we don't want to confuse with different names - therefore trust-checker is a concept not just one piece. Thanks for raising your concern though!

pkg/trustchecker/trustchecker.go
```diff
@@ -0,0 +1,302 @@
...
+		a.logger.Errorf("error reading roots dir %s: %v", rootsDir, err)
+		return noErr
+	}
+
+	var prevRoot *schema.Root
+	var root *schema.Root
+	verified := true
+	if len(roots) > 0 {
+		prevRootFilename := filepath.Join(rootsDir, roots[len(roots)-1].Name())
+		prevRootBytes, err := ioutil.ReadFile(prevRootFilename)
```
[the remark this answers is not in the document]
padurean: @mmeloni Please check the latest changes (i've extracted it and created an interface for it) and lmk if it's better.

Issue #260, opened 2020-05-26
pointer dereference if JSON fields are wrong /v1/immurestproxy/item/safe


Gjergj:
**Steps to reproduce:**

start immudb
./immudb

strart immugw
./immugwSteps to reproduce:

start immudb
./immudb

strart immugw
./immugw

issue
curl --location --request POST 'http://localhost:3323/v1/immurestproxy/item/safe?k1=v5' --header 'Content-Type: application/json' --header 'Authorization: Bearer {{token}}' --header 'Content-Type: text/plain' --data-raw '{
    "fsdf": {
         "key": "'$(echo -n client:Ms. Maci Schuppe | base64)'",
    "value": "'$(echo -n MasterCard 2232703813463070 12/19 | base64)'"
         }
}'

**Result**
runtime error: invalid memory address or nil pointer dereference.
immugw successfully recovers.

**Expect**
Reply with an error to the client.

PR #261, opened 2020-05-26
fix: #260

Gjergj:


Gjergj:
I am also checking the others.

Gjergj:
Other handlers also have this problem.
Will follow them in another issue and PR as this is specific to this one.

pkg/gw/safeset_handler_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+	immuclient "github.com/codenotary/immudb/pkg/client"
+	"github.com/codenotary/immudb/pkg/server"
+	"github.com/grpc-ecosystem/grpc-gateway/runtime"
+	"github.com/lithammer/shortuuid"
+)
+
+func TestSafeset(t *testing.T) {
+	op := server.DefaultOptions()
+	op.Dir = "db_" + shortuuid.New()
```
padurean, line 20: Is it worth using a 3rd party library for this? I am thinking maybe just using the current unix timestamp would be good enough.
Usually we try to avoid adding a new dependency, unless it is really worth it.

Gjergj: Refactored the test and removed the dependency

pkg/gw/safeset_handler_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+func TestSafeset(t *testing.T) {
+	key := base64.StdEncoding.EncodeToString([]byte("Pablo"))
+	value := base64.StdEncoding.EncodeToString([]byte("Picasso"))
+	tcpPort := generateRandomTCPPort()
+	op := immudb.DefaultOptions().WithPort(tcpPort).WithDir("db_" + strconv.FormatInt(int64(tcpPort), 10))
+	s := immudb.DefaultServer().WithOptions(op)
+	go s.Start()
+	time.Sleep(2 * time.Second)
+	defer s.Stop()
```
mmeloni, line 35: When I launch test on my machine It return following error:
```
 WARNING: LOG Compact FAILED with error: open db_57655/immudb/000002.sst: no such file or directory
While opening new table: 2
```
I belive that folder is deleted before badger stop:
try with something:
```
defer func() {
		s.Stop()
		os.RemoveAll(op.Dir)
	}()
```

pkg/gw/safeset_handler_test.go
```diff
@@ -0,0 +1,125 @@
...
+		payload        string
+		wantStatus     int
+		wantFieldName  string
+		wantFieldValue interface{}
+	}{
+		{ //Pablo = UGFibG8=
+			//Picasso = UGljYXNzbw==
+			"Send correct request",
+			`{"kv": {"key": "` + key + `", "value": "` + value + `"	} }`,
+			200,
```
mmeloni, line 57: please use a more meaningful constant like http.StatusOK

pkg/gw/safeset_handler_test.go
Gjergj: Yes, already fixed for #264 in 
https://github.com/codenotary/immudb/pull/274/commits/7e4d9789485a4c43755dd3a72450580e1de85f71
Will do the same here

pkg/gw/safeset_handler_test.go
Gjergj: Actually the status is not used in the tests yet as mentioned in the comment at the end of the function.
#263 is a blocker for that

PR #265, opened 2020-05-27
feat: Audit agent added to immuclient.

### LINKED ISSUE #193: immuclient as a consistency check agent
immuclient can be started as an agent that checks the current immudb state and if ok stores the current root every period (ENV variable defines the frequency, by default it is 1min); and also has the ability to check where it has been tampered, in case the last entry comes back tampered. 

immuclient startagent | stopagent 

The process starts in the background by default

volknanebo:
closes #193


cmd/immuclient/audit/auditor.go
```diff
@@ -0,0 +1,108 @@
...
+var ErrAgentNotActive = errors.New("agent not active")
+
+func (cAgent *auditAgent) InitAgent() (AuditAgent, error) {
+	srv, err := service.NewDaemon(name, description, name)
+	if err != nil {
+		return nil, err
+	}
+	ctx := context.Background()
+	cAgent.Daemon = srv
+	freqstr := os.Getenv("trust-check-interval")
```
padurean, line 50: @volknanebo is this env var name ok? i thought the trust checker name is obsolete and we use audit(or) instead.

cmd/immuclient/audit/auditagent.go
```diff
@@ -0,0 +1,199 @@
...
+
+type AuditAgent interface {
+	Manage(args []string) (string, error)
+}
+
+type auditAgent struct {
+	daemon.Daemon
+	cycleFrequency int
+	promot         promotheusExporter
+	ImmuTc         auditor.Auditor
```
padurean, line 41: @volknanebo Why the name recalling the old idea of trust checking?

cmd/immuclient/cli/flags.go
```diff
@@ -71,17 +70,21 @@ func parseflags() error {
...
 		return err
 	}
+	if err := viper.BindPFlag("config", pflag.Lookup("config")); err != nil {
+		return err
+	}
 	pflag.Parse()
 	o.InitConfig(o.CfgFn)
-	viper.SetDefault("immudb-port", gw.DefaultOptions().ImmudbPort)
-	viper.SetDefault("immudb-address", gw.DefaultOptions().ImmudbAddress)
-	viper.SetDefault("auth", client.DefaultOptions().Auth)
```
mmeloni, line 78: @padurean it's correct to remove auth flag?

padurean: @mmeloni I know this was @volknanebo s dream :) since the first time he saw it and if i remember correctly it was agreed we should remove it. In `immuadmin` it is removed already and it works like this: if the token file is present and can be read, the client will send the token (i.e. the token file contents). So no flag needed. `immuclient` can do the same, so if it is tested and it works in both cases (with and without auth disable in `immudb` server), then it's fine.

cmd/immuclient/audit/utils.go
```diff
@@ -0,0 +1,37 @@
...
+
+package audit
+
+import (
+	"fmt"
+	"os"
+)
+
+// QuitToStdErr prints an error on stderr and closes
+func QuitToStdErr(msg interface{}) {
```
mmeloni, line 25: We should use helper in https://github.com/codenotary/immudb/blob/master/cmd/helper/cmd.go#L102 to avoid duplicates. This change request is postponed after released

cmd/immuclient/audit/auditagent.go
```diff
@@ -0,0 +1,199 @@
...
+		options.MTLsOptions = client.DefaultMTLsOptions().
+			WithServername(servername).
+			WithCertificate(certificate).
+			WithPkey(pkey).
+			WithClientCAs(clientcas)
+	}
+	return options
+}
+
+func getCommands() string {
```
mmeloni, line 189: This is not used it seems

cmd/immuclient/cli/flags.go
volknanebo: @padurean :)
Yes , auth flag removed.

cmd/immuclient/service/constant.go
```diff
@@ -0,0 +1,33 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package service
+
+import "errors"
+
+var (
```
mmeloni, line 21: Duplicated part. Postponed

padurean:
@volknanebo Why couldn't the existing auditor (that was already implemented and used in `immugw`) be reused in your code? It was designed to be as pluggable as possible - i.e. all that's needed to reuse it is to initialize it (with  interval, server address and port, dial options, username, password, a cache implementation and a callback to update metrics related to audit result) and call `.Run` on it:

```
if s.Options.Audit {
	defaultAuditor, err := auditor.DefaultAuditor(
		s.Options.AuditInterval,
		fmt.Sprintf("%s:%d", s.Options.ImmudbAddress, s.Options.ImmudbPort),
		cliOpts.DialOptions,
		s.Options.AuditUsername,
		s.Options.AuditPassword,
		cache.NewHistoryFileCache(filepath.Join(cliOpts.Dir, "auditor")),
		Metrics.UpdateAuditResult,
	)
	if err != nil {
		s.Logger.Errorf("unable to create auditor: %s", err)
		return err
	}
	auditorDone := make(chan struct{})
	go defaultAuditor.Run(s.Options.AuditInterval, ctx.Done(), auditorDone)
	defer func() { <-auditorDone }()
}
```

padurean:
Not sure if this branch was rebased from master, but the at least the [user.go](https://github.com/codenotary/immudb/blob/feat/immuclient-audit-agent/cmd/immuadmin/command/user.go) file has old code (compared to the `master` branch) which is causing weird errors to Filippo when testing.

PR #274, opened 2020-05-27
Fix/264

Gjergj:


pkg/gw/safereference_handler_test.go
```diff
@@ -0,0 +1,187 @@
...
+	}{
+		{
+			"Send correct request",
+			`{
+				"ro": {
+					"reference":  "` + refKey + `",
+					 "key": "` + refKey + `"
+				}
+			}`,
+			200,
```
mmeloni, line 104: please use meaningful constants

pkg/gw/safezadd_handler_test.go
```diff
@@ -0,0 +1,159 @@
...
+	uknownKey := base64.StdEncoding.EncodeToString([]byte("Marias Callas"))
+	tcpPort := generateRandomTCPPort()
+	//MetricsServer must not be started as during tests because prometheus lib panis with: duplicate metrics collector registration attempted
+	op := immudb.DefaultOptions().WithPort(tcpPort).WithDir("db_" + strconv.FormatInt(int64(tcpPort), 10)).WithMetricsServer(false)
+	s := immudb.DefaultServer().WithOptions(op)
+	go s.Start()
+	time.Sleep(2 * time.Second)
+	defer func() {
+		s.Stop()
+		time.Sleep(2 * time.Second) //without the delay the db dir is deleted before all the data has been flushed to disk and results in crash.
```
mmeloni, line 30: In order to slower down tests execution with sleeps we could add a method on immudb like store.isClosed or similar and then loop a that bool. What's your opinion?

Gjergj: > we could add a method on immudb like store.isClosed or similar and then loop a that bool.

It's not worth it just for tests. At a later point I'd rather refactor these tests to run in parallel and two seconds for each in parallel won't add too much. I will open an issue for this so I won't forget it.
I'd say let's leave it as it is for the moment as there are not many tests.

mmeloni: Ok perfect

PR #281, opened 2020-05-28
fix: immuadmin dump hangs indefinitely if token is invalid

Gjergj:


development-lab:
Building everything with guitub I think we be ok

pkg/client/client.go
```diff
@@ -1066,18 +1066,18 @@ func (c *immuClient) Dump(ctx context.Context, writer io.WriteSeeker) (int64, er
 		}
 		if err != nil {
 			errs = append(errs, fmt.Sprintf("error receiving chunk: %v", err))
-			continue
+			break
```
padurean, line 1069: If we break the loop everywhere, instead of continuing it, then gathering errors does not make sense anymore.
In fact, if we break the loop on the first error, then it makes more sense to just return the error immediately - i.e. return, instead of break.
What do you think, @Gjergj ?

Gjergj: I agree. Still I'd like to keep the same flow to have the same structure and messages appear to the user.

padurean: Why?

Gjergj: As in the following commit

padurean: If the error happens in the inner `for`, the outer `for` will still run (and also any subsequent inner `for`s) => the error will be overwritten. Is this intended behaviour?

Gjergj: From the original code I assumed that the reason for the continue usage was that no matter what error still continue dumping data.
How should dump behave on error abort immediately or continue no matter what?

padurean:
@Gjergj In the initial version my idea was to dump as much as possible while gathering the errors for entries/chunks that could not be dumped.
From the changes in this PR i understand that was causing the bug that this very PR should solve (i.e. it was causing the dump to hang indefinitely when auth token is invalid).
So, regarding tour last question:
> How should dump behave on error abort immediately or continue no matter what?
the answer should come from you, i think.
From the code changes in this PR, which are basically about replacing the `continue` with `brwak`, i thought that’s the fix to the bug: aborting dump in case of error.
My point was that in this case we should just ‘return` right away (instead of `break` which is a bit tricky since there are inner loops involved.
Now if you as me whether it should continue to dump on error or not, i am asking you: if i yes, will this still fix the bug? Wouldn’t then it be the same as on the master version? Or is it something else that fixes the bug and this was just an extra- “cosmetic” touch?

Gjergj:
@padurean 
Since I joined the project last week I may have missed the reasoning behind some features.

> Or is it something else that fixes the bug and this was just an extra- “cosmetic” touch?
Was this affected by something else?

But I get your point, you are completely right.

padurean:
@Gjergj I've finally got to interactively test these changes.

1. With empty database
**Current behaviour**: Crash is fixed, but the file is still created (empty ) and the success message is printed with `0` for the lines number - i.e. `SUCCESS: 0 key-value entries were backed-up to file`.
**Suggestion**: In the function definition from `immuadmin/backup.go`, If there is no error, but result is <= 0, remove the file, print something like _Database is empty._. E.g.:
```
response, err := cl.immuClient.Dump(ctx, file)
if err != nil {
	color.Set(color.FgHiBlue, color.Bold)
	fmt.Println("Backup failed.")
	color.Unset()
	os.Remove(filename)
	c.QuitWithUserError(err)
} else if response <= 0 {
	os.Remove(filename)
	fmt.Println("Database is empty.")
	return nil
}
```

2. With invalid token:
**Current behaviour**: Same as with empty db (even if the db is not empty): bkp still gets created, it is not removed (as one would expect) and `SUCCESS: 0 key-value entries were backed-up to file` is printed.
This happens because the `break`s in the `Dump` function from `pkg/client/client.go` somehow fail to prevent the `err` from being `nil` until the function returns.
**Suggestion**: Just `return` on the first error encountered (instead of `break`). This way the behavior is as expected and the folowing error gets printed:
```
Backup failed.
error receiving chunk: rpc error: code = Unauthenticated desc = invalid token, re-login is required
```

padurean:
@Gjergj P.S.Am i wrong or you've created 2 PRs (different branches and also code differences, although with some overlapping)? Please keep just one as it is confusing right now. Thanks.

Gjergj:
Yes, #284 includes this one also.
So I will close this one and continue on the other.

PR #286, opened 2020-05-28
feat: add server-side logout

### LINKED ISSUE #231: Improve auth/token management

### LINKED ISSUE #262: add server logout to immudb
In the latest changes auth keys are stored per user and are dropped when user permissions are updated. This has the effect of invalidating any tokens that were signed with the dropped keys, hence user will be forced to re-login (and obtain a new token - user permissions are saved in the token payload, so they wouldn't have had an effect if the token wouldn't have been invalidated).
This is basically a simpler alternative to blacklisting tokens.
The same approach should be used to implement user logout on the server: the user keys should be dropped.
------
this is basically a duplicate of #231

padurean:
closes #262 
closes #231


Gjergj:
looks good

Issue #302, opened 2020-06-01
Cleanup config files from mixed data


vchain-us-mgmt:
when installing immudb service the /etc/immudb/immudb.toml file contains mixed immugw/immudb data
address = "127.0.0.1"
auth = false
certificate = "/etc/immudb/mtls/3_application/certs/localhost.cert.pem"
clientcas = "/etc/immudb/mtls/2_intermediate/certs/ca-chain.cert.pem"
dbname = "data"
detached = false
dir = "/var/lib/immudb"
immudb-address = "127.0.0.1"
immudb-port = 3322
logfile = "/var/log/immudb/immudb.log"
mtls = false
network = "tcp"
pidfile = "/var/lib/immudb/immudb.pid"
pkey = "/etc/immudb/mtls/3_application/private/localhost.key.pem"
port = 3322
servername = "localhost"
tokenfile = "token"

Gjergj:
@vchain-us-mgmt is this behavior also on windows?

PR #308, opened 2020-06-01
chore: remove immugw configs from immudb config file #302

Gjergj:


Gjergj:
Login doesn't work. It seems app and installation config are mixing.
Incoming with a fix.

padurean:
@mmeloni If you find the time please also add your review.
(I've tested it and it fixes the problem)

PR #320, opened 2020-06-04
feat: add auth support to immutest CLI

### LINKED ISSUE #319: add auth to immutest and print the 1st 100 values inserted

padurean:
closes #319


Gjergj:
Is tokenfile as a flag left out on purpose?

padurean:
@Gjergj Yes, but the only reason is that just an env var was requested.

PR #321, opened 2020-06-04
Fix/inclusionrebooting

mmeloni:


pkg/store/treestore.go
```diff
@@ -361,7 +361,9 @@ func (t *treeStore) Width() uint64 {
 
 func (t *treeStore) Set(layer uint8, index uint64, value [sha256.Size]byte) {
 	t.caches[layer].Set(index, &value)
-
+	if index < t.cPos[layer] {
+		t.cPos[layer] -= 1
+	}
```
leogr, line 366:
```suggestion
	if pos := index+1; pos < t.cPos[layer] {
		t.cPos[layer] = pos
	}
```
IIRC `index` is zero-based and `t.cPos` is equivalent to the length (ie. `index+1`),  thus we could calculate the exact position to invalidate the cache.

PR #330, opened 2020-06-09
chore: githubactions add windows and build step

Gjergj:
Actions:
   add windows as run step
   add build (make all) as run step


mmeloni:
@Gjergj win build is failing seems

Gjergj:
@mmeloni It's not this commit that is causing the fail. It simply made the fail evident.
This commit simply adds windows as a platform to run tests on.
Now we have to investigate why tests are failing on windows.

PR #343, opened 2020-06-12
use badger in memory for client tests

padurean:
Hi @mmeloni & @Gjergj 
I am not sure how good this idea is, so i am asking for your feedback through this PR:
**what do you think about using Badger in-memory (only) for tests which are not meant to test the store itself?**
As an example, i've modified the client tests to use this approach.

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -22,7 +22,6 @@ import (
 	"log"
 	"net"
 	"os"
-	"path/filepath"
 	"strconv"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


mmeloni:
If store is not involved should works and it should simplify things

Gjergj:
Thats a good idea and it makes sense for most tests, should make them faster also.
And the few ones that need to check the persistence functionality could use the old way.

padurean:
Ok, if you think it's a good idea, then @mmeloni you can merge the PR if you want.

PR #349, opened 2020-06-15
chore(cmd/helper): return a warning if a process is already present w…

mmeloni:
…hen launching in detached mode. Fixes 168


mmeloni:
> @mmeloni Is this requested just for linux?

Yes, but I will increase compatibility  also for windows.

PR #359, opened 2020-06-30
chore: enhance immudb server messages during start

padurean:


padurean:
image
image

pkg/store/logger.go
```diff
@@ -0,0 +1,44 @@
+/*
```
mmeloni, line 1: why we need the N logger? We have already simple and file logger in logger package.

PR #362, opened 2020-07-01
pkg/client tests

mmeloni:


cmd/immuclient/audit/executable.go
```diff
@@ -63,5 +63,5 @@ func (e *executable) Run() {
 			fmt.Println(err.Error())
 		}
 	}()
-	e.a.ImmuAudit.Run(time.Duration(e.a.cycleFrequency)*time.Second, e.stop, e.stop)
+	e.a.ImmuAudit.Run(time.Duration(e.a.cycleFrequency)*time.Second,false, e.stop, e.stop)
```
padurean, line 66: Perhaps configuring GoLand to format on save would be a good idea?
I am using VS Code and i have this enabled so that i don't have to manually trigger `go fmt`.
Not very important, but i thought i should mention it though. This is how it can be achieved in GoLand as far as i read:
> There's a file watcher section under Preferences → Tools → File Watchers, in there you can setup to run go fmt or goimports everytime you save a Go file.

pkg/client/auditor/auditor.go
```diff
@@ -100,16 +100,22 @@ func DefaultAuditor(
...
 	stopc <-chan struct{},
 	donec chan<- struct{},
-) {
+) (err error) {
 	defer func() { donec <- struct{}{} }()
 	a.logger.Infof("starting auditor with a %s interval ...", interval)
-	err := repeat(interval, stopc, a.audit)
-	if err != nil {
-		return
+	if singleRun {
```
padurean, line 109: I suppose this could've been easily achieved without any code change, by calling the `defaultAuditor.audit` method directly in the unit test (since unit tests test the implementation, anyway, not the interface). Not a big deal, though.

cmd/immuclient/audit/executable.go
mmeloni: I'm using precommit. https://github.com/codenotary/immudb/blob/master/.pre-commit-config.yaml  Probably I used wrongly -n (no verify) option.

pkg/client/auditor/auditor.go
mmeloni: Agree, but I liked to test Run method

PR #364, opened 2020-07-01
tests(pkg/server): consinstency checker, options, server

Gjergj:


pkg/server/corruption_checker_test.go
```diff
@@ -84,3 +87,7 @@ func (l *mockLogger) Warningf(f string, v ...interface{}) {}
 func (l *mockLogger) Infof(f string, v ...interface{}) {}
 
 func (l *mockLogger) Debugf(f string, v ...interface{}) {}
+
+func (l *mockLogger) CloneWithLevel(level logger.LogLevel) logger.Logger {
+	return logger.NewSimpleLogger("check loger", io.MultiWriter())
```
mmeloni, line 92: You could avoid the real implementation in a mock
```suggestion
	return &logger.FileLogger{}
```

PR #377, opened 2020-07-05
make DevMode default false and cleanup call to act…

padurean:
…ion message shwon right after immudb start.
Also fix bug: immudb must start on default port (not on port 0) if the specified port is not a number or not > 0.


cmd/immutest/command/init.go
```diff
@@ -72,7 +72,7 @@ func Init(cmd *cobra.Command, o *c.Options) {
 		}
 		defer cl.disconnect(cmd, nil)
 		serverAddress := cl.immuClient.GetOptions().Address
-		if serverAddress != "127.0.0.1" && serverAddress != "localhost" {
+		if serverAddress != "127.0.0.1" && serverAddress != "0.0.0.0" && serverAddress != "localhost" {
```
mmeloni, line 75: We need this? We can have local connection also with 0.0.0.0

padurean: I thought we should keep also the check fir `127.0.0.1` just in case the user explicitly configures it as such.

PR #380, opened 2020-07-06
test(pkg/gw): increase coverage in handlers test by abstracting away …

padurean:
…and mocking gRPC runtime, json and ImmuClient


padurean:
@mmeloni @Gjergj In order to increase the coverage in `immugw` handlers, i've added the following abstractions which allow injecting function definitions in tests and which may be useful also in other tests. So feel free to move them outside of the `pkg/gw` package and extend/enhance them if you find them useful (**click on them to view the code**).
- [`immuClientMock`](https://github.com/codenotary/immudb/blob/test/immugw-cli-codecov/pkg/gw/immuclient_mock_test.go)
    - it's a struct that implicitly implements the `ImmuClient` interface by embedding it
    - provides the possibility to override the definition of the needed function by setting the value of a `func` field
- [`json`](https://github.com/codenotary/immudb/blob/test/immugw-cli-codecov/pkg/gw/json.go)
    - follows the same approach as `immuClientMock`: a struct that implements a (new) interface
    - provides the possiblility to override the definition of the needed function by setting the value of a `func` field

For usages examples, please have a look in the immugw handlers tests from `pkg/gw`.

mmeloni:
> @mmeloni @Gjergj In order to increase the coverage in `immugw` handlers, i've added the following abstractions which allow injecting function definitions, hence they might be useful also in your/other tests. So feel free to move them outside of the `pkg/gw` package and extend/enhance them if you find them useful (**click on them to view the code**).
> 
> * [`immuClientMock`](https://github.com/codenotary/immudb/blob/test/immugw-cli-codecov/pkg/gw/immuclient_mock_test.go)
>   
>   * it's a struct that implicitly implements the `ImmuClient` interface by embedding it
>   * provides the possibility to override the definition of the needed function by setting the value of a `func` field
> * [`json`](https://github.com/codenotary/immudb/blob/test/immugw-cli-codecov/pkg/gw/json.go)
>   
>   * follows the same approach as `immuClientMock`: a struct that implements a (new) interface
>   * provides the possiblility to override the definition of the needed function by setting the value of a `func` field
> 
> For usages examples, please have a look in the immugw handlers tests from `pkg/gw`.

Thanks @padurean . In order to be usable everywhere I propose to follow golang httptest package. We could create, when needed, a *test package in which we put tools useful to mock or testing that specific package.

padurean:
@mmeloni 
- I've moved `ImmuClientMock` under `clienttest` package, as you suggest.
NOTE: only the `ImmuClient` methods that are used in `immugw` handlers test are added to the mock right now, but any other needed methods can be easily added.
- i've moved the JSON abstraction in it's own (new) package `pkg/json`

PR #382, opened 2020-07-07
test(pkg/server/servertest): add server test utils

mmeloni:


Gjergj:
GetStandardServerTestOptions might be a bit misleading and most uses of immuserver is with auth on

mmeloni:
> GetStandardServerTestOptions might be a bit misleading and most uses of immuserver is with auth 
Fix

PR #391, opened 2020-07-10
test/immuc

Gjergj:


cmd/immuclient/immuc/immuctest/currentstatus_test.go
```diff
@@ -14,12 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
```
mmeloni, line 14: To increase consistency of what you have done I suggest:
1. move all test files from immuctest in immuc.(one folder back)
2. in these file use immuc_test package Please see -> https://golang.org/pkg/cmd/go/internal/test/ 
> Cit: Test files that declare a package with the suffix "_test" will be compiled as a
> separate package, and then linked and run with the main test binary.

3. fix last imports

In this way we can have tests in the same folder of tested files (like in other places), and we can use test* folder as reported:

> The go tool will ignore a directory named "testdata", making it available
> to hold ancillary data needed by the tests.

In this way in my opinion is more consistent from what we have in the rest of the project.

cmd/immuclient/immuclienttest/helper.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (h *homedirServiceMock) DeleteFileFromUserHomeDir(pathToFile string) error {
+	return nil
+}
+
+func (h *homedirServiceMock) ReadFileFromUserHomeDir(pathToFile string) (string, error) {
+	return string(h.token), nil
+}
+
+func NewClient(pr helper.PasswordReader, dialer servertest.BuffDialer, hds client.HomedirService) immuc.Client {
```
mmeloni, line 56: This is confusing to me:
there is NewClient method that create new client from a PasswordReader, BuffDialer and HomedirService
then we have Login that do the same stuff of NewClient + login but it returns also HomedirService

We should have a struct with following exposed dependencies:
- client
- clientOptions
- tokenfile
- homeDirService(if you need it in a second time)

end on the receiver we could expose following methods:
- connect
- login

Feel free to improve or rearrange if you like

PR #396, opened 2020-07-14
immuadmin stats tests

mmeloni:


.travis.yml
```diff
@@ -89,9 +89,11 @@ jobs:
       script:
         - go get golang.org/x/tools/cmd/cover
         - go get github.com/mattn/goveralls
+        - go get -u github.com/ory/go-acc
```
Gjergj, line 92: On the github.com/ory/go-acc project it says that it is a cross platform replacement of a bash script. Since the code coverage takes into account only Linux tests can we simply use that bash script instead of go-acc project?

mmeloni: We could, but we are not adding the library in the repository, so I don't see the problem.

Gjergj: Fine

PR #398, opened 2020-07-14
Copy File Operations

pascaldekloe:
Minor fixes and tuning.

### DIFF
--- a/pkg/fs/copy.go
+++ b/pkg/fs/copy.go
@@ -29,16 +29,16 @@ import (
 // destination file exists, all it's contents will be replaced by the contents
 // of the source file. The file mode will be copied from the source and
 // the copied data is synced/flushed to stable storage.
-func CopyFile(src, dst string) (err error) {
+func CopyFile(src, dst string) error {
 	in, err := os.Open(src)
 	if err != nil {
-		return
+		return err
 	}
 	defer in.Close()
 
 	info, err := in.Stat()
 	if err != nil {
-		return
+		return err
 	}
 	if info.IsDir() {
 		return fmt.Errorf("%s is a directory", src)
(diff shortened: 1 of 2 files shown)


mmeloni:
Thanks @pascaldekloe, looking good to me. Could you please rewrite commit using conventional commits standard?

Ex:
fix(pkg/fs): don't overwrite copy error on Close malfunction. Sync seals the operation ; not Close.

Please also remove my unwanted merge commit (44257b94b034a90c4a8cfefe40c73ecf3d5d0724 ) and rebase with master to update branch

pascaldekloe:
Maybe those commit messages ain't that important? 😉 That pkg name is also meaningless. Rename to internal if you care about conventions. It explicitly declares the API as unstable.

padurean:
@pascaldekloe
Commit messages are important because release notes are generated based on them.
Regarding `pkg` name, it is present in Go package layout standards - see [here](https://github.com/golang-standards/project-layout). Code from this package is not exclusively internal (as it would be the case with an `internal` package), it is ok for it to be used also by external applications.

pascaldekloe:
Please don't be fooled by the name @padurean. These "standards" are just one man's ideas. Naming a package `pkg` is questionable to say the least. You can omit the level/name and not lose anything on descriptiveness nor structure.

If you really want people to use the content of `pkg` then you should also commit to compatibility. Meaning: any change should raise the major version of the entire project. Say you drop this `ring` package in favour of something that is already available: new immudb version.

https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface

The commit messages are very descriptive. If you care about the change-log, then package names have no place there.

padurean:
@pascaldekloe Nice article - thanks, i didn't know about it and i wasn't aware of Dave Cheney's opinion on this.
If we would follow the advice - renaming `pkg` to `internal` - we would probably have to change other things too.
For example we would definitely need to move `client.go` (which exposes a Go API external apps) from the `client` package underneath `pkg` to some other location.
I suppose it would be a good idea to do these changes, but right now there are many other important things that we are already working on, so perhaps this is not a good time to do it.

Coming back to the commits:
1. Although i don't really mind too much the `fs` instead of `pkg/fs` in the commit message, i also don't really see a reason why we would break consistency with both previous commits and the package folders right now. For sure it's not a big deal, but also not difficult to change, just for the sake of consistency.
2. Regarding this one: `chore(fs): tracable copy errors`, perhaps the right English word is `traceable`?

pascaldekloe:
Sorry for being a bit of pain here @padurean. This pull request was part of an investigation. I have an interview for the Go developer position scheduled on Friday. Cheers guys.

padurean:
No problem, @pascaldekloe. Good luck with the interview!

pascaldekloe:
Eh, no-go. Too bad. Please add the `os. O_TRUNC` flag to the `os.OpenFile` call. Last part of the test. Will leave you to it now.

padurean:
@pascaldekloe Ok - sorry and thanks!

PR #474, opened 2020-08-17
Fix immudb and immugw version and mangen commands errors

padurean:
Without this change, while _**immuclient**_ and _**immuadmin**_ still worked as expected, _**immudb**_ and _**immugw**_ `version` and `mangen` commands (and possibly other ones too) are throwing the following error:
```
./immugw version
Error: flag accessed but not defined: config
Usage:
  immugw version [flags]

Flags:
  -h, --help   help for version

flag accessed but not defined: config
```

### DIFF
--- a/cmd/helper/config.go
+++ b/cmd/helper/config.go
@@ -17,12 +17,13 @@ limitations under the License.
 package helper
 
 import (
+	"os"
+	"strings"
+
 	service "github.com/codenotary/immudb/cmd/immuclient/service/constants"
 	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	"os"
-	"strings"
 )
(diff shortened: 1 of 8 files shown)


cmd/helper/config.go
```diff
@@ -59,10 +60,9 @@ func (c *Config) Init(name string) error {
 	return nil
 }
 
+// LoadConfig loads the config file (if any) and initializes the config
 func (c *Config) LoadConfig(cmd *cobra.Command) (err error) {
-	if c.CfgFn, err = cmd.Flags().GetString("config"); err != nil {
-		return err
-	}
+	c.CfgFn, _ = cmd.Flags().GetString("config")
 	if err = c.Init(c.Name); err != nil {
```
jeroiraz, line 66: err will always be 'nil' at this point

PR #483, opened 2020-08-25
refactor: move immugw in a separate repository

mmeloni:


README.md
```diff
@@ -1,887 +1,745 @@
-<!--
```
jeroiraz, line 1: it seems the entire file got changed...

cmd/immuadmin/command/init.go
```diff
@@ -18,7 +18,7 @@ import (
 	"strings"
 
 	"github.com/codenotary/immudb/pkg/client"
-	"github.com/codenotary/immudb/pkg/gw"
+	"github.com/codenotary/immugw/pkg/gw"
```
jeroiraz, line 21: some additional refactoring might be needed around immuadmin as well, immugw should not be a dependency neither...

cmd/immutest/command/init.go
```diff
@@ -29,8 +29,8 @@ import (
 	"github.com/codenotary/immudb/pkg/api/schema"
 	"github.com/codenotary/immudb/pkg/auth"
 	"github.com/codenotary/immudb/pkg/client"
-	"github.com/codenotary/immudb/pkg/gw"
 	"github.com/codenotary/immudb/pkg/server"
+	"github.com/codenotary/immugw/pkg/gw"
```
jeroiraz, line 33: idem with immuadmin, it might be a signal of something that needs to be refactored

go.mod
```diff
@@ -3,41 +3,32 @@ module github.com/codenotary/immudb
 go 1.13
 
 require (
-	github.com/DataDog/zstd v1.4.5 // indirect
+	github.com/codenotary/immugw v0.0.0-20200825141010-b6326c81edf2
```
jeroiraz, line 6: I think we should avoid the need of having immugw as dependency

pkg/service/sservice.go
```diff
@@ -20,8 +20,8 @@ import (
 	"io"
 
 	immudb "github.com/codenotary/immudb/cmd/immudb/command"
-	immugw "github.com/codenotary/immudb/cmd/immugw/command"
 	"github.com/codenotary/immudb/pkg/immuos"
+	immugw "github.com/codenotary/immugw/cmd/immugw/command"
```
jeroiraz, line 24: also here

README.md
mmeloni: The reason is that I have also replaced CRLF with LF. I added a separated commit with only  CRLF to evidence changes

cmd/immuadmin/command/init.go
mmeloni: Immuadmin need immugw because it manages immugw service installing procedure. So it need to call some immugw api to, for example, install man pages.
A solution could be to move out also immuadmin, but this is out of scope for the moment.

cmd/immutest/command/init.go
mmeloni: Fixed

pkg/service/sservice.go
mmeloni: Same point as beforre

PR #484, opened 2020-08-26
feat(pkg/tbtree): initial implementation of tbtree data-structure

jeroiraz:


leogr:
Just out of curiosity: why is this needed?

PR #489, opened 2020-09-01
Crypto signature + root sign 

mmeloni:


pkg/api/schema/root.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/codenotary/immudb/pkg/signer"
+	"github.com/golang/protobuf/proto"
+)
+
+func NewRoot() *Root {
+	return &Root{Payload: &RootIndex{}, Signature: &Signature{}}
+}
+
+//CheckSignature
+func (r *Root) CheckSignature() (ok bool, err error) {
```
jeroiraz, line 30: returning error may be enough, nil when signature validates

mmeloni: Thanks @jeroiraz, in that case I should parse the error, because inside verify  'unmarshal' could fail. So if an error is found I need to check if there was an unmarshal error or the signature is not valid.

PR #515, opened 2020-10-01
Added a not on the file limit in the FAQ

SimoneLazzaris:
I've added some information about the file descriptor usage in the FAQ


SimoneLazzaris:
Added instructions for operating with many databases in the README.md file

PR #525, opened 2020-10-16
Fix ineffectual assignments

muesli:
This is a tougher PR to review, I'm sorry. I fully understand if you reject it
and approach this from a different angle. However I've stumbled upon a whole
range of ineffectual assignments in your code.

Most of them were assignments to an error variable, but missing the actual error
check. In the various test cases, I've tried to stick to whatever was the convention
in the particular test: require.NoError, require.Nil, direct err check.

I suggest you want to annotate a bunch of them. There were too many here for me
to do this in a reasonable fashion in this PR.

Should you intentionally skip these checks, I would recommend omitting the assignment
as well, as this clearly states an intention to future readers/reviewers.


pkg/store/scan_test.go
```diff
@@ -127,13 +127,18 @@ func TestIScan(t *testing.T) {
 	var zero []byte
 	zero = append(zero, byte(0))
 	idx, err := st.Set(schema.KeyValue{Key: zero, Value: []byte(`item0`)})
+	require.NoError(t, err)
```
muesli, line 130: Looks like this is supposed to trigger an error?

muesli:
Happy to rebase this should you consider it.

pkg/server/server_test.go
```diff
@@ -420,7 +420,7 @@ func testServerUseDatabase(ctx context.Context, s *ImmuServer, t *testing.T) {
 	}
 	m := make(map[string]string)
 	m["Authorization"] = "Bearer " + string(dbs.Token)
-	ctx = metadata.NewIncomingContext(ctx, metadata.New(m))
+	metadata.NewIncomingContext(ctx, metadata.New(m))
```
jeroiraz, line 423: actually, last three lines (421-423) may be removed as they are not doing any meaningful validation

pkg/store/scan_test.go
```diff
@@ -124,16 +124,16 @@ func TestIScan(t *testing.T) {
 	st, closer := makeStore()
 	defer closer()
 
+	var err error
 	var zero []byte
 	zero = append(zero, byte(0))
-	idx, err := st.Set(schema.KeyValue{Key: zero, Value: []byte(`item0`)})
```
jeroiraz, line 129: I think error validation is worth here

jeroiraz:
thanks a lot @muesli for all the improvements, I've placed a few comments. Please read them, then feel free to rebase with master and we'll be happy to merge your contributions

pkg/server/server_test.go
muesli: Removed.

pkg/store/scan_test.go
muesli: Please double-check this. If I understand it correctly, this first call is supposed to return an error, whereas the rest of the `st.Set` calls are supposed to succeed.

jeroiraz: Indeed, that's because first byte being 0 is currently reserved, it's a limitation that will be fixed. So for the scope and intent of this unit test, please feel free to remove that initial insertion. First statement (after store initialization) could be  idx, err := st.Set(schema.KeyValue{Key: []byte(`0`), Value: []byte(`itemZERO`)})

jeroiraz: last detail on this:
- lines 127-128 could be become useless, so can be removed as well.

muesli: You're right. Done.

PR #533, opened 2020-10-31
some formatting improvements

verygoodsoftwarenotvirus:
I had to do a bunch of debugging in this package recently, and to be frank: I hated it.

That said, I'm a big believer in being the change you wish to see in the world, so I went through the main file that bugged me and added visual breaks at what I felt were intuitive places to make it less of a headache to read. 

Some other things present in this changeset:
- I've renamed a UUID function that violated the [Golang rule on initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms). I felt comfortable doing this because it was a private function.
- I added a field to the `ImmuServer` struct called `ReadyChan` that's just a boolean channel. If it isn't nil, then it gets written to once the server is up and ready. This would make it so that things like [this](https://github.com/codenotary/immudb.io/blob/f0cd688c23e0eabc9ae816de38e92282912b2d7b/src/jumpstart.md#L208) didn't have to exist.

I'm completely willing to do this sort of work for this whole project, as well as to fix all the linter errors, but I figured I'd throw ya'll this PR first and see how receptive ya'll would be to that.

Thanks for all your hard work!

### DIFF
--- a/pkg/server/types.go
+++ b/pkg/server/types.go
@@ -20,6 +20,7 @@ import (
 	"net/http"
 	"os"
 	"sync"
+	"time"
 
 	"google.golang.org/grpc"
 
@@ -61,6 +62,8 @@ type ImmuServer struct {
 	metricsServer       *http.Server
 	mux                 sync.Mutex
 	RootSigner          RootSigner
+	startedAt           time.Time
+	ReadyChan           chan bool
 }
 
 // DefaultServer ...
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ GO ?= go
 DOCKER ?= docker
 PROTOC ?= protoc
 STRIP = strip
+GO_FORMAT := gofmt -s -w
 
 V_COMMIT := $(shell git rev-parse HEAD)
 #V_BUILT_BY := "$(shell echo "`git config user.name`<`git config user.email`>")"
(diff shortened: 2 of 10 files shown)


padurean:
Thank you @verygoodsoftwarenotvirus for this PR, we highly appreciate your contribution. We're looking into it and we'll come back with feedback soon.

pkg/server/types.go
```diff
@@ -61,6 +62,8 @@ type ImmuServer struct {
 	metricsServer       *http.Server
 	mux                 sync.Mutex
 	RootSigner          RootSigner
+	startedAt           time.Time
+	ReadyChan           chan bool
```
jeroiraz, line 66: I'm thinking ReadyChan could be implemented as a callback channel provided as an argument while constructing or starting the server instance, instead of having it as part of the struct. Also it could be of type `chan struct{}`so to maintain same type as quit channel

jeroiraz (changes requested):
Thanks a lot for the contribution! Please review my comment

devisions:
@verygoodsoftwarenotvirus Hi Jeffrey,

We really appreciate your interest and efforts on this.
Fortunately - and unfortunately for this PR case - we changed the core of the database considerably, shaping it for a better future.
Therefore, I will close this PR. But thanks again for the contribution! :pray: 

Your feedback has been received, and we strive to improve the linting state and namings.

PR #609, opened 2021-01-08
fix(cmd/sservice): fix services management and add permissions guard

mmeloni:


SimoneLazzaris (dismissed):
Function permissionGuard can be optimized moving out of the for cycle the os.Stat and the info.IsDir check.

But since permissionWhitelist has only one element, question is moot.

PR #638, opened 2021-02-02
Feat/streams

mmeloni:


jeroiraz (dismissed):

pkg/client/streams.go
```diff
@@ -0,0 +1,522 @@
...
+	//<--
+
+	s, err := c.streamVerifiableSet(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	kvss := c.StreamServiceFactory.NewKvStreamSender(s)
+
+	// 1st send the ProveSinceTx (build a "fake" KV with it):
```
mmeloni, line 201: It's a bit ugly to send the `ProveSinceTx` with a fake send.
You could use directly the message sender or create a NewVeridiedKvStreamerSender

PR #646, opened 2021-02-25
fix(pkg/server): use latest snapshot when listing users

jeroiraz:


dxps (approved):
LGTM & and `ListUsers` is working fine now.

PR #656, opened 2021-03-12
Non-blocking Index Compaction

jeroiraz:
**Non-blocking Index Compaction**

First, the btree is flushed to disk and the current snapshot is created.

Snapshot is dumped (nodes and commit logs) based on the latest index transaction: "nodes_TxID" and "commit_TxID" folders

Meanwhile, indexing and other operations behave as usual (non-blocking)

Once the dump is complete, nodes and commit folders may be replaced anytime. Note that indexed entries after the snapshot used to dump the index will be replayed.

It’s equivalent to restoring the index from a backup (which was compacted)

When index is replaced with a compacted one, the indexing goroutine is woken up.


SimoneLazzaris:
All stresstests are green now:insertion has gone steady for 72 hours, and (sampled) retrieving of data is working

PR #659, opened 2021-03-12
feat: add support for user, password and database flags in immuclient

padurean:
Signed-off-by: Valentin Padurean (Ogg) <912518+padurean@users.noreply.github.com>


dmacvicar (approved):
I don't like that we are looking into password error strings, but I understand we don't have the infrastructure to do something different yet.

mmeloni (dismissed):

jeroiraz (approved):

PR #676, opened 2021-04-01
restore some immuclient tests, add illegal state guard on verified methods

mmeloni:


mmeloni:
Reduced coverage will be fixed when state and signature verification interceptor will be added. At the moment make no sense add more tests on sdk if we agree on that interceptor creation.

jeroiraz (approved):

Issue #679, opened 2021-04-02
immuclient panic when using a wrongly formatted token


ellisab:
**What happened**

alex@MacBook-Air immudb % ./immuclient
panic: runtime error: makeslice: len out of range

goroutine 1 [running]:
github.com/codenotary/immudb/pkg/client.(*tokenService).parseContent(0x140000a8660, 0x140000a74b0, 0x140001ad928, 0x1, 0x0, 0x140000aa070, 0x140001ad901)
	/Users/alex/go/src/immudb/pkg/client/token_service.go:104 +0xb4
github.com/codenotary/immudb/pkg/client.(*tokenService).GetToken(0x140000a8660, 0x1037a1260, 0x0, 0x0, 0x1)
	/Users/alex/go/src/immudb/pkg/client/token_service.go:46 +0x28
github.com/codenotary/immudb/pkg/client.(*immuClient).SetupDialOptions(0x1400038a000, 0x1400008b800, 0x140000a8660)
	/Users/alex/go/src/immudb/pkg/client/client.go:308 +0x34c
github.com/codenotary/immudb/pkg/client.NewImmuClient(0x1400008b800, 0x140001adc01, 0x0, 0x0, 0x140001add08)
	/Users/alex/go/src/immudb/pkg/client/client.go:199 +0x124
github.com/codenotary/immudb/cmd/immuclient/immuc.(*immuc).Connect(0x140000be8c0, 0x1037a1260, 0x0, 0x0, 0x140001adcf0, 0x140001adcc8)
	/Users/alex/go/src/immudb/cmd/immuclient/immuc/init.go:91 +0x58
github.com/codenotary/immudb/cmd/immuclient/command.(*commandline).NewCmd.func1(0x140000e2000, 0x1037a1260, 0x0, 0x0, 0x0, 0x0)
	/Users/alex/go/src/immudb/cmd/immuclient/command/root.go:49 +0x50
github.com/spf13/cobra.(*Command).execute(0x140000e2000, 0x14000032210, 0x0, 0x0, 0x140000e2000, 0x14000032210)
	/Users/alex/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:842 +0x320
github.com/spf13/cobra.(*Command).ExecuteC(0x140000e2000, 0x18, 0x20, 0x140000ec601)
	/Users/alex/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950 +0x258
[...]

PR #680, opened 2021-04-02
fix for https://github.com/codenotary/immudb/issues/679

ellisab:


jeroiraz:
thanks @ellisab!, I think we can take the opportunity to improve the readability as well.

databasel stands for database name length, it's actually a buffer which is used to finally get the encoded length as a numeric value in variable databaselUint64. Probably both variables may be renamed.

A validation to ensure the length is greater than zero may be convenient as well as checking that remaining content has enough bytes as specified by that length.

What do you think?

pkg/client/token_service.go
```diff
@@ -93,14 +93,17 @@ func (ts *tokenService) parseContent() (string, string, error) {
...
-	databaselUint64 := binary.BigEndian.Uint64(databasel)
-	databasename := make([]byte, int(databaselUint64))
-	copy(databasename, content[8:8+int(databaselUint64)])
-
-	token := make([]byte, len(content)-8-int(databaselUint64))
-	copy(token, content[8+int(databaselUint64):])
+	dbnamelen := make([]byte, 8)
+	copy(dbnamelen, content[:8])
+	dbnamelenUint64 := binary.BigEndian.Uint64(dbnamelen)
+	if int(dbnamelenUint64) > len(content) {
```
jeroiraz, line 99: yes, actually should not be greater than remaining content, 8 bytes were already consumed when reading the encoded length

ellisab: @jeroiraz https://play.golang.org/p/WT-Nin6UVBb
checking that dbnamelenUint64. is not bigger than the whole token avoids the panic

I do not think that casting to int is needed  if uint64(len(content)) is used https://play.golang.org/p/cdH3lMq5TC0

jeroiraz: I mean something like
lenb := make([]byte, 8)
copy(lenb, content)

dbNameLen := int(binary.BigEndian.Uint64(lenb))

 `if dbNameLen > len(content)-8` ...  // note: here dbNameLen should be less than whole content less 8 bytes used to encode the len of the dbName itself

jeroiraz:
Thanks a lot @ellisab for your contribution :)

PR #685, opened 2021-04-07
Scans running over fully up-to-date snapshots as default behaviour

jeroiraz:
This PR changes default parameters when scanning keys by prefix and sorted-sets, default parameters will make use of a fully up-to-date snapshot.

If older snapshots can be used (or reused) `sinceTx=TxID` can be specified as a requirement for the index state


padurean (dismissed):
LGTM.

PR #687, opened 2021-04-07
Refactor README.md

### LINKED ISSUE #672: README revamp
The current README can be improved in many areas. Current problems that are visible:

* Mixes too many topics, from news, to hiring, deep explanations
* Duplicates documentation and goes into advanced topics like operating immudb in production (eg. services)
* Goes deep into exposing advanced options of commands, which are duplicated in the command help, and are not necessary for getting started

Proposing a new README mission and according structure that is consistent with itself (just an example. About, Getting Started, Contributing, Resources) would benefit the project.

Inspiration in other projects README's is encouraged, especially if they are in the same field.

- [ ] Propose a structure and content
- [ ] Implement the new structure and content

trwnh:
README.md is now much shorter and has a clear user story.

Yet to be done:

- [x] Verify quickstart instructions manually
- [ ] Figure out how to connect dockerized immuclient/immuadmin to immudb
- [ ] Link to playground

Fix #672

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,0 +1,41 @@
+# Build the binaries yourself
+
+To build the binaries yourself, simply clone this repo and run
+
+```
+make all
+```
+
+## Linux (by component)
+
+```bash
+GOOS=linux GOARCH=amd64 make immuclient-static immuadmin-static immudb-static
+```
+
+## MacOS (by component)
+
+```bash
+GOOS=darwin GOARCH=amd64 make immuclient-static immuadmin-static immudb-static
+```
+
+## Windows (by component)
+
+```bash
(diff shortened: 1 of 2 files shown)


dmacvicar (approved):
Thanks. I think it is a good step forward.

I think we can increase the impact of the first paragraph in a new iteration. Especially about highlighting the idea that the data is protected by the clients.

I was thinking something along the lines of:

>immudb is a key-value database with built in cryptographic proof and verification.  It can track changes in sensitive data and the integrity of the history will be protected by the clients, without the need to trust the server.
>
>Unlike blockchains, immudb can handle hundred thousands transactions per second, and can be used both as a lightweight service or embedded in your application as a library.
>
>Companies use immudb to protect credit card transactions and to secure processes by store digital certificates and checksums.

dmacvicar (approved):

PR #691, opened 2021-04-09
Further README improvements

dmacvicar:
* This simplifies the introduction text further
* Removes the "companies using immudb" section
* Adds a missing subtitle
* Move the quickstart section before the "Using immudb"


README.md
```diff
@@ -32,25 +32,13 @@ Don't forget to ⭐ this repo if you like immudb!
...
 
-immudb has 4 main benefits:
+Unlike blockchains, immudb can handle millions of transactions per second, and can be used both as a lightweight service or embedded in your application as a library.
 
-1. **immudb is immutable**. You can add records, but **never change or delete records**. This lets you store critical data without fear of it being changed silently.
-2. Data stored in immudb is **cryptographically coherent and verifiable**, like blockchains, just without all the complexity and at high speed.
-3. Anyone can get **started with immudb in minutes**. Whether you're using node.js, Java, Python, Go, .Net, or any other language. It's very easy to use and you can have your immutable database running in just a few minutes.
-4. Finally, immudb is **Open Source**. You can run it **on premise**, or in the **cloud**. It's completely free. immudb is governed by the Apache 2.0 License.
+Companies use immudb to protect credit card transactions and to secure processes by store digital certificates and checksums.
```
trwnh, line 39: s/store/storing

README.md
```diff
@@ -32,25 +32,13 @@ Don't forget to ⭐ this repo if you like immudb!
...
-immudb is a **lightweight, high-speed immutable database** with built-in cryptographic proof and verification.
-With immudb, you can track changes in sensitive data in your transactional databases and then record those changes permanently in a
-tamperproof immudb database.
-<img align="right" src="img/immudb-mascot-small.png" width="256px"/>
-
-Traditional DB transactions and logs are hard to scale and are mutable, so there is no way to know for sure if your data has been compromised.
-immudb provides **unparalleled insights** **retroactively** of changes to your sensitive data, even
-if your perimeter has been compromised. immudb guarantees immutability by using **cryptographic data structures** internally.
-This allows you to keep a persistent history of sensitive data, such as debit/credit card transactions.
+immudb is a key-value database with built-in cryptographic proof and verification. It can track changes in sensitive data and the integrity of the history will be protected by the clients, without the need to trust the server.
```
trwnh, line 35: "Traditional DB transactions and logs are hard to scale and are mutable, so there is no way to know for sure if your data has been compromised."

This line is useful to explain the value proposition of immudb -- I'd keep it.

README.md
```diff
@@ -32,25 +32,13 @@ Don't forget to ⭐ this repo if you like immudb!
...
 
-immudb has 4 main benefits:
+Unlike blockchains, immudb can handle millions of transactions per second, and can be used both as a lightweight service or embedded in your application as a library.
```
trwnh, line 37: I'd also keep benefits 1 and 2 at least -- it helps explain immutability and verification as well as why you would want them.

"**immudb is immutable**. You can add records, but **never change or delete records**. This lets you store critical data without fear of it being changed silently. Data stored in immudb is **cryptographically coherent and verifiable**, just like blockchains, but without all the complexity. Unlike blockchains, immudb can handle millions of transactions per second, and can be used both as a lightweight service or embedded in your application as a library."

README.md
```diff
@@ -160,6 +112,34 @@ To perform actions, simply run ```./immuclient``` on the same machine or ```./im
 
 To manage the database, simply run ```./immuadmin``` on the same machine. You can also find immuadmin in the immudb container at `/usr/local/bin/immuadmin` or run a Docker container to connect to the local immudb with `docker run -it --rm --name immuadmin codenotary/immuadmin:latest status`. For security reasons we recommend using immuadmin only on the same system as immudb. User management is restricted to localhost usage.
 
+## Using immudb
```
trwnh, line 115: Could go either way on the ordering here; I flipped it because of discussion around wanting potential users to understand how they'd use immudb before giving them quickstart instructions. But I'm not adamant about one over the other.

trwnh (changes requested):

README.md
dmacvicar: Thanks. I brought back this benefits with this paragraph, with a small edit:

>You can add records, but never change or delete records

I think this wording confuses people, like if you can't set the same key to a new value. So I changed it to:

>You can add new versions of existing records, but never change or delete records

README.md
dmacvicar: I think this "so there is no way to know for sure if your data has been compromised." is covered by the next part: "This lets you store critical data without fear of it being changed silently"

trwnh: I'd still keep it -- the latter is about immudb, the former is about traditional databases. Having both clauses sets up a compare-and-contrast for the reader.

README.md
```diff
@@ -32,25 +32,13 @@ Don't forget to ⭐ this repo if you like immudb!
...
-immudb provides **unparalleled insights** **retroactively** of changes to your sensitive data, even
-if your perimeter has been compromised. immudb guarantees immutability by using **cryptographic data structures** internally.
-This allows you to keep a persistent history of sensitive data, such as debit/credit card transactions.
+immudb is a key-value database with built-in cryptographic proof and verification. It can track changes in sensitive data and the integrity of the history will be protected by the clients, without the need to trust the server.
 
-immudb has 4 main benefits:
+immudb is immutable. You can add new versions of existing records, but never change or delete records. This lets you store critical data without fear of it being changed silently. Data stored in immudb is cryptographically coherent and verifiable, just like blockchains, but without all the complexity. Unlike blockchains, immudb can handle millions of transactions per second, and can be used both as a lightweight service or embedded in your application as a library.
```
trwnh, line 37: On second thought, this should be broken into two paragraphs, one for immutability and one for verifiability. Something like this:

"Traditional DB transactions and logs are hard to scale and are mutable, so there is no way to know for sure if your data has been compromised. immudb is immutable. You can add new versions of existing records, but never change or delete records. This lets you store critical data without fear of it being changed silently.

Data stored in immudb is cryptographically coherent and verifiable, just like blockchains, but without all the complexity. Unlike blockchains, immudb can handle millions of transactions per second, and can be used both as a lightweight service or embedded in your application as a library."

trwnh (changes requested):

trwnh (approved):
LGTM

PR #696, opened 2021-04-14
Call server logout even if local token file could not be deleted

### LINKED ISSUE #695: Logout call not done due to failure removing token
**What happened**

The logout call fails when trying to remove a token file.

```go
package main

import (
	"fmt"
	"context"
	"log"

	immuclient "github.com/codenotary/immudb/pkg/client"
)

func main() {
	client1, err := immuclient.NewImmuClient(immuclient.DefaultOptions())
	if err != nil {
		log.Fatal(err)
	}
	ctx1 := context.Background()
	// login with default username and password
	lr1, err := client1.Login(ctx1, []byte(`user1`), []byte(`TempPassword1.`))
	if err != nil {
		log.Fatal(err)
	}
	// immudb provides multidatabase capabilities.
	// token is used not only for authentication, but also to route calls to the correct database
	md1 := metadata.Pairs("authorization", lr1.Token)
	fmt.Println(lr1.Token)
	ctx1 = metadata.NewOutgoingContext(context.Background(), md1)

	_, err = client1.Set(ctx1, []byte(`hello`), []byte(`immutable world`))
	if err != nil {
		log.Fatal(err)
	}

	err = client1.Logout(ctx1)
	if err != nil {
		log.Fatal(err)
	}

	_, err = client1.Set(ctx1, []byte(`hello`), []byte(`immutable world`))
	if err != nil {
		log.Fatal(err)
	}

}
```

```
2021/04/14 10:39:07 remove /home/duncan/token: no such file or directory
exit status 1
```

This causes the problem that if the `Logout` error happens, the token is not removed from the metadata, and the next `Set` call succeeds.

Logout should do the logout call before cleaning any token.

Use the code attached above

**Environment**

Tried it on the Go SDK with 0.9.1 and 0.9.2.

padurean:
closes #695

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -23,6 +23,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
@@ -483,15 +484,23 @@ func (c *immuClient) Logout(ctx context.Context) error {
 		return ErrNotConnected
 	}
 
-	if err := c.Tkns.DeleteToken(); err != nil {
+	if _, err := c.ServiceClient.Logout(ctx, new(empty.Empty)); err != nil {
 		return err
 	}
 
-	_, err := c.ServiceClient.Logout(ctx, new(empty.Empty))
+	tokenFileExists, err := c.Tkns.IsTokenPresent()
+	if err != nil {
+		return fmt.Errorf("error checking if token file exists: %v", err)
+	}
+	if tokenFileExists {
+		if err := c.Tkns.DeleteToken(); err != nil {
+			return fmt.Errorf("error deleting token file during logout: %v", err)
+		}
+	}
 
 	c.Logger.Debugf("logout finished in %s", time.Since(start))
 
-	return err
+	return nil
 }
 
 // CurrentState returns current database state
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -17,13 +17,15 @@ package client
 
 import (
 	"context"
+	"errors"
 	"fmt"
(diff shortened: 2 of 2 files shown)


dmacvicar:
This fix still leaves questions open to me:

Why is the token deletion failing in a simple example? Is a token creation expected in this example?

If the creation is expected, then I'd ask why is not being created. The error should still be captured. In that case, I'd ask if we can revert the order of the calls, to execute the Logout on the server side anyways, and then delete the token locally.

If the creation is not expected, then I'd make the deletion a "delete if exists" like this fix.

padurean:
@dmacvicar In the example you used the token is set directly in the context metadata. The token service of the client is not actually used to save the token to a file by calling `SetToken`. In the immu CLIs code, the token is explicitly persisted locally:
```go
        response, err := i.ImmuClient.Login(ctx, user, pass)
	if err != nil {
		if strings.Contains(err.Error(), "authentication is disabled on server") {
			return "authentication is disabled on server", nil
		}
		return "", errors.New("username or password is not valid")
	}
	if err = i.ts.SetToken("", response.Token); err != nil {
		return "", err
	}
```

dmacvicar:
Ok. Then I'd not use the text `error deleting token file during logout:` as it is a normal situation. The token should be deleted only if it was persisted. No token means it was not persisted, but not an error.

padurean:
@dmacvicar Ok - i've changed the fix to first check if the token file is present before attempting to delete it.
Please review it again.

padurean:
@dmacvicar @mmeloni If you agree with the fix, please approve the PR so that i can merge it and get it out of the way.

dmacvicar (approved):
Looks good to me, but please @mmeloni look at it too.

mmeloni (approved):

PR #718, opened 2021-04-27
Feat/pgsqlwire sqlsupport

mmeloni:


jeroiraz (approved):

pkg/database/types.go
```diff
@@ -0,0 +1,61 @@
+package database
```
padurean, line 1: Seems like this new file is missing the CodeNotary copyright header?

pkg/pgsql/server/bmessages/data_row.go
```diff
@@ -0,0 +1,104 @@
...
+	"bytes"
+	"encoding/binary"
+	"github.com/codenotary/immudb/pkg/api/schema"
+)
+
+func DataRow(rows []*schema.Row, colNumb int, binaryFormat bool) []byte {
+	rowsB := make([]byte, 0)
+	for _, row := range rows {
+		rowB := make([]byte, 0)
+		////##-> dataRow
```
padurean, line 29: Just out of curiosity: this comment seems to follow some unclear convention - e.g. being a kind of start marker for a block of code? But then i don't see an end marker for the block (e.g. i was expecting something like `////##<-` somewhere lower) ...

pkg/pgsql/server/bmessages/ready_for_query.go
```diff
@@ -0,0 +1,30 @@
...
+*/
+
+package bmessages
+
+import (
+	"bytes"
+	"encoding/binary"
+)
+
+func ReadyForQuery() []byte {
```
padurean, line 24: Is it really mandatory or is there an advantage of having each of these messages in their own dedicated go files instead of all of them in a single file (as they seem relatively small and serving a similar purpose)?

pkg/pgsql/server/session_factory.go
```diff
@@ -17,34 +17,24 @@ limitations under the License.
 package server
```
padurean, line 17: _**pkg/server/databaselist.go**_ being moved and renamed to `pkg/pgsql/server/session_factory.go` looks a bit weird, doesn't it? Why would something that has a general purpose to something that is pgsql specific?
Or perhaps it's just git diff making it look weird?

padurean (commented):
Some general remarks:
- there are some TODOs in the code; not sure how important they are or how complex it is to get rid of them, but ideally we should have as few as possible to avoid them lingering in the code longer than needed
- in the _pkg/pgsql/server_ package, i noticed these warnings shown by VS Code when linting with _go-staticcheck_ is enabled:
image

pkg/pgsql/server.go
```diff
@@ -0,0 +1,33 @@
...
+package main
+
+import (
+	"github.com/codenotary/immudb/pkg/pgsql/server"
+	"log"
+)
+
+func main() {
+
+	sqlServer := server.New(server.Port(5439))
```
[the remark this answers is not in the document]
dmacvicar:
```
viper.SetDefault("pgsql-server-port", 5432)
```

I think this is a bug, as the port is configurable.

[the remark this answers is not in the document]
mmeloni: It's a main package, I will remove

pkg/pgsql/server/bmessages/ready_for_query.go
mmeloni: There are messages that are complex. So I prefer to keep them separated to eliminate at the beginning the risk of large files.

pkg/pgsql/server/session_factory.go
mmeloni: I think it's the second. Databaselist is something I moved from immudb server package to immudb database package

PR #725, opened 2021-04-30
Add http annotations to SQL rpc methods

dmacvicar:
This PR adds the missing annotations to expose the SQL methods that are used to generate the gateway

As master also changed the API prefix for those (when @mmeloni updated immugw), I changed those as well so that it will be - easier to merge later.

The idea is to add the REST gw and its options in master, but have ready the SQL end-points in here.

### DIFF
--- a/pkg/api/schema/schema.proto
+++ b/pkg/api/schema/schema.proto
@@ -665,14 +665,29 @@ service ImmuService {
 
 	// SQL
 	rpc SQLExec(SQLExecRequest) returns (SQLExecResult) {
+		option (google.api.http) = {
+			post: "/db/sqlexec"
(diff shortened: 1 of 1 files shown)


mmeloni:
@dmacvicar Immurestproxy prefix is already removed and tx is present on master. I suggest to rebase

dmacvicar:
If the prefix is not removed, this PR does not make much sense.
Rebasing sql_support on master is the best, but I don't know  how much effort that is. If that is possible, I'd rebase sql_support on master and then send this PR.

dmacvicar:
Rebased to master

jeroiraz (approved):

PR #738, opened 2021-05-08
feat: CREATE TABLE IF NOT EXISTS

dmacvicar:
This PR adds support for CREATE TABLE IF NOT EXISTS, which skips the table creation if a table with the same name is already present.


jeroiraz (dismissed):

dmacvicar:
@jeroiraz had to repush. Had committed/signed with the wrong email.

jeroiraz (approved):

PR #739, opened 2021-05-08
Feat SQL: `NOT NULL` constraint

jeroiraz:
This PR incorporates support for `NOT NULL` column constraint.

e.g. `CREATE TABLE myTable (id INTEGER, title VARCHAR, active BOOLEAN NOT NULL, PRIMARY KEY id)`

### DIFF
--- a/embedded/sql/catalog.go
+++ b/embedded/sql/catalog.go
@@ -42,6 +42,7 @@ type Column struct {
 	id      uint64
 	colName string
 	colType SQLValueType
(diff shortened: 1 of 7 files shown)


dmacvicar:
Needs rebase.

dmacvicar:
Is the reason that the validation occurs in the bytes() serializaton of the rowspec that it can be reused between upsert/upsert into?

dmacvicar (approved):

PR #745, opened 2021-05-11
Run full tests on Linux, and all client tests on Windows

dmacvicar:
This PR enables the full testsuite on Linux, and the client testsuite on Windows, by doing:

* Move all coverage stuff from `make test` out. There is a specific coveralls target, so there is no real need to calculate coverage in make test. If this is needed, a separate target could be done.
* Remove the exclusion of many modules from running the tests, which was done because of coverage
* Create a client-test target to tests the client package
* Adapt the Github workflow to run the full `test` target on Linux, and `test-client` on Windows.

There are failing tests.

* The Windows test failing is the bug on https://github.com/codenotary/immudb/tree/fix/file-lock-issue which would have been catch before affecting [vcn](https://github.com/codenotary/vcn/pull/147).
* More on the full testsuite, which should be fixed anyway.
* History cache:
  * The test was checking for failure if the directory does not exist, but the code `mkdirs` the directories it needs.
  * The test passed on Linux by accident: the non-existing directory used in the test was `/not-exists` and failed not because of non-existence but because of permissions.
  * There is a chance the test was correct, and actually testing just fishing coverage points for the rare case of permissions. If that is the case, the test did not work on Windows, because admin can anyway create the directory.
  * This PR adapts the test to create a non-existing directory with write-access (Tmpdir), and expecting no errors, as it should be auto-created.


dmacvicar:
@mmeloni I still have

```
immudb  2021/05/13 21:02:01 INFO: Flushing index 'data\systemdb\index' successfully completed
--- FAIL: TestUserManagement (0.21s)
    client_test.go:723: 
        	Error Trace:	client_test.go:723
        	Error:      	Received unexpected error:
        	            	rpc error: code = Unknown desc = please select a database first
        	Test:       	TestUserManagement
FAIL
FAIL	github.com/codenotary/immudb/pkg/client	7.492s
```

failing on Windows.

padurean (approved):
LGTM!

PR #756, opened 2021-05-14
Enhance metrics for DB size and number of entries to support multiple databases and add CORS to metrics endpoints

### LINKED ISSUE #730: immudb_db_size_bytes metrics (and others) does not support multiple-databases
The metric needs to be updated to support multiple databases (eg. give the database name as a label).

padurean:
closes #730 
Examples from the response of the `/metrics` endpoint:

---

- [x] DB sizes
```bash
# HELP immudb_db_size_bytes Database size in bytes.
# TYPE immudb_db_size_bytes gauge
immudb_db_size_bytes{db="cnlcdata"} 61044
immudb_db_size_bytes{db="cnlcinternalledger"} 20600
immudb_db_size_bytes{db="cnlcsyslogledger"} 2156
immudb_db_size_bytes{db="defaultdb"} 20612
immudb_db_size_bytes{db="lcledgeronekkgeln"} 107992
immudb_db_size_bytes{db="systemdb"} 11158
```
⚠️ The db sizes are **NOT** the actual size that is occupied on the disk. I tried to do it, initially, but it only worked on Linux (using `Fstat` from golang.org/x/sys package). I already spent to much time trying to find out how to do it in a cross-platform way, so i resorted to the same approach that is on `master` which uses `os.FileInfo.Size()`.

---

- [x] Number of entries
```bash
# HELP immudb_number_of_stored_entries Number of key-value entries currently stored by the database.
# TYPE immudb_number_of_stored_entries gauge
immudb_number_of_stored_entries{db="cnlcdata"} 34
immudb_number_of_stored_entries{db="cnlcinternalledger"} 0
immudb_number_of_stored_entries{db="cnlcsyslogledger"} 0
immudb_number_of_stored_entries{db="defaultdb"} 0
immudb_number_of_stored_entries{db="lcledgeronekkgeln"} 111
immudb_number_of_stored_entries{db="systemdb"} 11
```
⚠️ The number of entries are actually the latest transaction ID of each database, as i am not aware of any other (better) way of "counting" all entries.

---

- [x] Update `immuadmin stats` command to correctly handle the new metrics format above (both for the default graphical mode and the `--text` mode).
⚠️ Instead of the _defaultdb_, the `stats` command now shows data for the database which has the most entries (i.e. it's assuming that that one is the one which would be of most interest for the user). In the future we should enhance it to show stats for the specified db - e.g. with a new flag `--db`.

- [x] I've also added CORS to metrics endpoints in a separate commit c15b64042101db0eaa3996a14ba7b320eb405f51 (i closed the separate PR #757 that i created for that)

### DIFF
--- a/cmd/immuadmin/command/stats/controller.go
+++ b/cmd/immuadmin/command/stats/controller.go
@@ -119,17 +119,19 @@ func updatePlot(
 }
 
 func (p *statsController) Render(ms *metrics) {
-	uptime, _ := time.ParseDuration(fmt.Sprintf("%.4fh", ms.db.uptimeHours))
+	db := ms.dbWithMostEntries()
+
+	uptime, _ := time.ParseDuration(fmt.Sprintf("%.4fh", ms.uptimeHours))
 	p.SummaryTable.Rows = [][]string{
 		{"[ImmuDB stats](mod:bold)", fmt.Sprintf("[ at %s](mod:bold)", time.Now().Format("15:04:05"))},
-		{"Database", ms.db.name},
+		{"Database", db.name},
 		{"Uptime", uptime.String()},
-		{"Entries", fmt.Sprintf("%d", ms.db.nbEntries)},
+		{"Entries", fmt.Sprintf("%d", db.nbEntries)},
 		{"No. clients", fmt.Sprintf("%d", ms.nbClients)},
 		{"  active < 1h ago", fmt.Sprintf("%d", len(*ms.clientsActiveDuringLastHour()))},
 	}
 
-	totalSizeS, totalSize := byteCountBinary(ms.db.totalBytes)
+	totalSizeS, totalSize := byteCountBinary(db.totalBytes)
 	updatePlot(
 		p.SizePlot,
 		&p.SizePlotData,
--- a/cmd/immuadmin/command/stats/metrics.go
+++ b/cmd/immuadmin/command/stats/metrics.go
@@ -18,10 +18,13 @@ package stats
 
 import (
 	"fmt"
+	"math"
 	"time"
 
 	dto "github.com/prometheus/client_model/go"
 )
+
+const SystemdbName = "systemdb"
 
 var readers = map[string]bool{
 	"ByIndex":     true,
@@ -70,10 +73,9 @@ type rpcDuration struct {
 }
 
 type dbInfo struct {
-	name        string
-	totalBytes  uint64
-	nbEntries   uint64
-	uptimeHours float64
+	name       string
+	totalBytes uint64
+	nbEntries  uint64
 }
 
 type operations struct {
@@ -97,7 +99,8 @@ type metrics struct {
 	nbClients            int
 	nbRPCsPerClient      map[string]uint64
 	lastMsgAtPerClient   map[string]uint64
-	db                   dbInfo
+	uptimeHours          float64
+	dbs                  map[string]dbInfo
 	memstats             memstats
 }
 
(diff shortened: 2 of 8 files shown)


dmacvicar (changes requested):
As discussed in the team, we should follow the https://prometheus.io/docs/practices/naming/ recommendation:

>    ...should use base units (e.g. seconds, bytes, meters - not >milliseconds, megabytes, kilometers). See below for a list of base units.

dmacvicar (commented):
I think leaving the size and entries method the same is ok. It was not the scope of the PR.

padurean:
This PR is now ready to be reviewed. Unfortunately the latest pushes didn't seem to trigger the GitHub workflows checks ...

Hextar:
I think the Server is still missing to add the "Access-Control-Allow-Origin" in the "Access-Control-Allow-Headers" to avoid receiving a CORS error from the browser.

padurean:
Weird. I could not reproduce the problem - i never new that the `Access-Control-Allow-Origin` (and `Access-Control-Allow-Methods`?) needs to be explicitly added to the value of `Access-Control-Allow-Headers`.
Of course, i could just add them. But i am not sure that it is really needed: i can not reproduce your error with Firefox 88.0.1 on Ubuntu. This is the script that i use in an html page that i load on one machine, requesting metrics from another machine:

```html
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="UTF-8">
</head>
<body>
<script>
   fetch('http://192.168.1.247:9497/metrics')
     .then(data => { return data.text(); })
     .then(res => { console.log(res); })
     .catch(error => { throw(error); });
</script>
</body>
</html>
```

And in the browser console i can see the metrics - see screenshot - no CORS-related error.
image

If i then run it against the `master` branch of immudb (which does not have the CORS fix), it does throw a CORS error (second screenshot).
image

I am curious why i can't reproduce ...

Hextar:
@padurean because firefox is more permissive on CORS. Is better to directly test on Chrome.

padurean:
@Hextar I've just tried with Chrome 90.0.4430.212 on Ubuntu, with the same script and i still can't reproduce it:
image

And i was not able to find at least 1 CORS example which adds those `Access-Control-...` headers to the list of allowed headers ...

Does that HTML & JS that i pasted in my previous comment show the CORS error if you test with it?

padurean:
@Hextar I rewrote a bit the code so that only the needed headers and values are set on `OPTIONS` and `GET` methods.
I've still not added those 2 headers to the allowed headers header.
Lmk how it works (or not) for you with the latest changes.
If you still get those errors, i will just add the 2 headers to allowed headers.

Hextar:
@padurean using the .html I also have no problem fetching the resource, but even if it works with vanilla javascript, it could probably be something related to vue and/or axios (the dependecy used to interact with XHR).

I guess that I still need those 2 more headers if you could add them.

padurean:
@Hextar I've added them. Please retest.

Hextar:
@padurean I've tested your latest push and it was still saying that 'Access-Control-Allow-Credentials' was missing from the 'Access-Control-Allow-Headers'.

I've directly added it to the pkg/server/metrics.go file to avoid bothering you for just a string, and now it works!

Btw as far as I've understood those CORS errors happens only when running an node application

padurean:
@Hextar Alright. 👍

dmacvicar:
@Hextar I understand you added the missing headers in 8211525 and we can then merge this one?

dmacvicar (dismissed):

dmacvicar (approved):

dmacvicar (approved):

PR #759, opened 2021-05-16
Fix wrong History File Cache test

dmacvicar:
The test was checking for failure if the directory does not exist, but the code `mkdirs` the directories it needs.

The test passed on Linux by accident: the non-existing directory used in the test was `/not-exists` and failed not because of non-existence but because of permissions.

There is a chance the test was correct, and actually testing just fishing coverage points for the rare case of permissions. If that is the case, the test did not work on Windows, because admin can anyway create the directory.

This PR adapts the test to create a non-existing directory with write-access (Tmpdir), and expecting no errors, as it should be auto-created.

_This is a series of fixes to make the client tests pass on Windows._

### DIFF
--- a/pkg/client/cache/cache.go
+++ b/pkg/client/cache/cache.go
@@ -37,4 +37,3 @@ type HistoryCache interface {
 	Cache
 	Walk(serverUUID string, db string, f func(*schema.ImmutableState) interface{}) ([]interface{}, error)
 }
-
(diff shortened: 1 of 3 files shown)


dmacvicar:
I have added this commit to https://github.com/codenotary/immudb/pull/745 so I can evaluate how further it brings us in terms of Windows tests.

PR #760, opened 2021-05-16
feat(embedded/tools): initial SQL stress tool

dmacvicar:
This is an initial stress tool following the same pattern as `stress_tool.go` but using the embedded SQL engine.

Caveats:

* It fails if the database exists
* It generates serial IDs without persisting the last, therefore 2 consecutive runs would fail with duplicate IDs (perhaps a flag in the future indicating starting ID would help)

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -34,6 +34,6 @@ jobs:
           go get -u github.com/mattn/goveralls
           go get -u github.com/ory/go-acc
           set -o pipefail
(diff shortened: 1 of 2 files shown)


SimoneLazzaris:
Looks good for a first implementation. Possible improvements:
- add an option to use an existing database/table, so that the test can be run in series
- give the option to read the values instead of just doing insert, so that we can also test read speed
- git the option to have sequences, so that we can also check procedurally if insert are correctly read back

SimoneLazzaris (approved):

PR #761, opened 2021-05-17
fix: tlsConfig semantics changed when no TLS configured

dmacvicar:
Before pgsql branch merge, tlsConfig was nil if no TLS configured. This is not longer the case.

We don't make assumptions about the resulting tlsConfig object anymore. We *do* only start a TLS server if tlsConfig is non-nil and has certificates, but we only test the resulting server can accept the right type of connection.

### DIFF
--- a/pkg/server/webserver.go
+++ b/pkg/server/webserver.go
@@ -30,7 +30,7 @@ func StartWebServer(addr string, tlsConfig *tls.Config, s schema.ImmuServiceServ
 
 	go func() {
 		var err error
-		if tlsConfig != nil {
+		if tlsConfig != nil && len(tlsConfig.Certificates) > 0 {
 			l.Infof("Web API server enabled on %s/api (https)", addr)
 			err = httpServer.ListenAndServeTLS("", "")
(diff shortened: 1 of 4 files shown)


pkg/server/webserver.go
```diff
@@ -30,7 +30,7 @@ func StartWebServer(addr string, tlsConfig *tls.Config, s schema.ImmuServiceServ
 
 	go func() {
 		var err error
-		if tlsConfig != nil {
+		if len(tlsConfig.Certificates) > 0 {
```
mmeloni, line 33: Better to check also tlsconfig != Nil in order to avoid a nil pointer error in runtime

dmacvicar: I am not sure. It would be an example defensive programming. I'd rather prefer it to crash and make visible our semantics are not consistent. The tests would fail anyway if the semantics are changed.

mmeloni: It's not for tests, but it's for server embedding.

mmeloni (approved):

PR #769, opened 2021-05-18
Smoke testing / webconsole basic test

ufierro:
This PR includes a set of basic tests for Windows and Linux and a specific test to make sure the webconsole works, basic tests will run in windows and ubuntu, additionally all tests and a extra one (web console specific, simple health check) will run against a webconsole version compiled in ubuntu


dmacvicar (approved):

ufierro:
Sorry @dmacvicar re-requesting review since I had to sign the commits

dmacvicar (commented):
Perhaps we don't need to do the smoke testing without the webconsole as we don't have plans to produce builds without it, and save some build minutes.

The only use-case I can think is that we will add a default static page explaining why there is no webconsole (eg "it was built without it") instead of a 404, and therefore we may want to test that.

@jeroiraz @mmeloni opinions?

ufierro:
Unless @jeroiraz @mmeloni  feel like testing without the webconsole provides any value I'll just remove that test

dmacvicar:
Lets leave it. It will be helpful to test https://github.com/codenotary/immudb/pull/786 once I merge it.

We can remove it later if for some reason we change our minds.

mmeloni (approved):

dmacvicar (approved):

PR #776, opened 2021-05-19
Feat/pgsql wire version and nil values

mmeloni:
This is added in order to handle version statement and handle properly nil values


dmacvicar:
What client using version() has been tested with this workaround? Does it resolve the problem for them, or they get then stuck in another PostgreSQL specific query?

mmeloni:
@dmacvicar . Yes, version is not enough for most of the GUI. But it's something we should have on driver side, not immudb since is specific related to postgres.

dmacvicar (approved):

PR #786, opened 2021-05-21
feat(webconsole): default web console page

dmacvicar:
This PR changes the default 404 if the web console was not built-in with a default static page.
/ responds with a redirect to /missing so that tests can see the difference.

# How it works

By default we checkin a statik.go generated with `make webconsole/default` target which includes the content of `webconsole/default` directory. The default end-point answers with moved permanently and redirects to this page.

This means the full-webonsole `make WEBCONSOLE=1` only overrides the checked-in `statik.go` file. In both cases, files are protected by tags, to avoid the wrong files to be present in a build.

Screenshot from 2021-05-21 19-15-02


webconsole/webconsole_default.go
```diff
@@ -2,12 +2,23 @@
...
-func SetupWebconsole(_ *http.ServeMux, _ logger.Logger, _ string) error {
-	// no op
+func SetupWebconsole(mux *http.ServeMux, l logger.Logger, addr string) error {
+	statikFS, err := fs.New()
+	if err != nil {
+		return err
+	}
+	l.Infof("Webconsole not built-in")
+	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		http.Redirect(w, r, "/missing/", http.StatusMovedPermanently)
```
byo, line 20: Wouldn't it be better to use some temporary redirect here? Browsers could remember such permanent redirection (including rewriting bookmarks etc) that could be problematic if one wants to switch back to the built-in webconsole.

dmacvicar: You are right.

mmeloni (approved):

PR #789, opened 2021-05-24
Always create system db (even when auth is off)

padurean:


pkg/server/server.go
```diff
@@ -334,23 +335,21 @@ func (s *ImmuServer) loadSystemDatabase(dataDir string, adminPassword string) er
...
-			}
+		s.sysDb = db
+		//sys admin can have an empty array of databases as it has full access
+		adminUsername, _, err := s.insertNewUser([]byte(auth.SysAdminUsername), []byte(adminPassword), auth.PermissionSysAdmin, "*", false, "")
+		if err != nil {
+			return logErr(s.Logger, "%v", err)
+		}
 
-			time.Sleep(time.Duration(10) * time.Millisecond)
+		time.Sleep(time.Duration(10) * time.Millisecond)
```
mmeloni, line 350: I think is not needed anymore since by default wait for indexing is true

padurean: I did not touch that part, seems unrelated to the scope of the PR.
But of course, i can delete that line if we are sure it is useless and won't affect anything.

padurean: Done.

mmeloni (approved):

padurean:
@dmacvicar @jeroiraz If you agree with the changes, please approve so that we can merge it and get it out of the way. Thanks.

jeroiraz (approved):

PR #794, opened 2021-05-25
fix: go mod tidy/vendor with statik module

dmacvicar:
The way statik works is by generating a module. This module is not present until `go generate` is run.

This is not a problem as the files are guarded by build tags, but `go mod vendor` and `go mod tidy` run with all build tags enabled, which result in a module referenced but not present.

This is a workaround. We should eventually use `go embed` present in Go 1.16.

### DIFF
--- a/webconsole/statik/dummy.go
+++ b/webconsole/statik/dummy.go
@@ -1,0 +1,2 @@
+// this file is here so that go vendor and go mod tidy work
+package statik


jeroiraz:
Alternatively, it's also possible to instruct statik to use the same package as value holder:
`//go:generate go run github.com/rakyll/statik -f -src=./dist -p=webconsole -dest=../`

In such case, the following lines may be removed from webconsole.go file:
```
  // embedded static files
  _ "github.com/codenotary/immudb/webconsole/statik"
```

dmacvicar:
I think I like your solution more. I will do some tests locally.

dmacvicar:
Superseded by https://github.com/codenotary/immudb/pull/796

PR #795, opened 2021-05-26
fix(cmd/immuclient): improve immuclient tx error message

### LINKED ISSUE #711: improve immuclient tx error messages
**What happened**
immuclient supports to query by transaction id
`immuclient tx #`

But the error messages are not clear or when the transactions id is invalid:
`immuclient tx 0` returns `illegal arguments`, `immuclient tx -0` tries to map it with a flag `unknown shorthand flag`

better error message would be `invalid TXid - transaction id range start with 1`

mangalaman93:
Fixes #711 

I don't think we should do something about the case `-0`.
Reference: https://github.com/spf13/cobra/issues/124

### DIFF
--- a/cmd/immuclient/immuc/getcommands.go
+++ b/cmd/immuclient/immuc/getcommands.go
@@ -18,6 +18,7 @@ package immuc
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


jeroiraz:
thanks @mangalaman93

Could you add the validation in `VerifiedGetTxByID` as well?

cmd/immuclient/immuc/getcommands.go
```diff
@@ -31,6 +31,9 @@ func (i *immuc) GetTxByID(args []string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf(" \"%v\" is not a valid id number", args[0])
 	}
+	if id == 0 {
+		return "", fmt.Errorf(" invalid tx id (should be bigger than 0)")
```
jeroiraz, line 35: First white space may not be needed

mangalaman93: Updated the message. I think equality is fine, we have an uint64.

mangalaman93:
Done

jeroiraz (approved):

PR #801, opened 2021-05-28
immudb improve error handling

mmeloni:
This add immuError on SDK.
```
        _, err = client.StreamSet(ctx, kvs)
	require.Equal(t, stream.ErrMaxValueLenExceeded, err.Error())
	require.Equal(t, errors.CodDataException, err.(errors.ImmuError).Code())
```

immuError is structured as follow:
```
type ImmuError interface {
	Error() string
	Cause() string
	Stack() string
	Code() Code
	RetryDelay() int32
}
```

Error return the message.
Cause is the inner error cause.
Stack is present if immudb is running with LEVEL_INFO=debug
Code is the immudb error code
RetryDelay if present the error is retryable after N milliseconds

On server side errors need to be returned using fllowing interface present in immudb error package:
```
type Error interface {
	Error() string
	Message() string
	Cause() error
	Code() Code
	RetryDelay() int32
	Stack() string
}
```

An error can be created with:

errors.New
errors.Wrap

```
       if read == 0 && err == io.EOF {
		return errors.New(ErrReaderIsEmpty).WithCode(errors.CodInvalidParameterValue)
	}
```
or
```
u, err := s.getValidatedUser(r.User, r.Password)
	if err != nil {
		return nil, errors.Wrap(err, "invalid user name or password").WithCode(errors.CodSqlserverRejectedEstablishmentOfSqlconnection)
	}

```

Errors are converted in the transportation in gRPC status. When adding a new error code it's required to add an entry in gRPC error map. See map.go
```
func mapGRPcErrorCode(code Code) codes.Code 
```

In order to  avoid dependency between client and server package SDK contains proper error codes definitions, like server.

Both SDK ImmuError and server Error can be compared with errors.Is utility.


pkg/client/errors/meta.go
```diff
@@ -0,0 +1,30 @@
...
+limitations under the License.
+*/
+
+package errors
+
+type Code string
+
+const (
+	SuccessCompletion                             Code = "00000"
+	CodInternalError                              Code = "XX000"
```
byo, line 23: What the `Cod` prefix stands for ?

pkg/errors/error_wrapper.go
```diff
@@ -0,0 +1,45 @@
...
+
+package errors
+
+import (
+	"context"
+	"github.com/codenotary/immudb/pkg/client/errors"
+	"google.golang.org/grpc"
+)
+
+func ServerStreamInterceptor(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
```
byo, line 25: How are those interceptors attached to the communication stream?

pkg/client/errors/meta.go
mmeloni: Cod is only a prefix that could be useful with editor autocompletion, since everything is in the same package. I think here there is a leftover since some entries are not prefixed.

pkg/errors/error_wrapper.go
mmeloni: This is another leftover. I tested interceptors but they didn't fit. I forgot to remove this file

pkg/client/errors/errors.go
```diff
@@ -0,0 +1,131 @@
...
+*/
+
+package errors
+
+import (
+	"github.com/codenotary/immudb/pkg/api/schema"
+	"google.golang.org/grpc/status"
+)
+
+type ImmuError interface {
```
byo, line 24: Can we add some comment to this interface? That should make things clear for users of immudb SDK.
The PR description is pretty nice, I'd copy some parts of it over here.

mmeloni: Sure, I'm waiting your feedbacks then I will add comments when we agree on the implementation

pkg/client/streams.go
```diff
@@ -123,7 +109,7 @@ func (c *immuClient) StreamGet(ctx context.Context, k *schema.KeyRequest) (*sche
 	value, err := stream.ReadValue(vr, c.Options.StreamChunkSize)
 	if err != nil {
 		if err == io.EOF {
-			return nil, stream.ErrMissingExpectedData
+			return nil, errors.New(stream.ErrMissingExpectedData).WithCode(errors.CodInternalError)
```
jeroiraz, line 112: if error declaration already includes the code it will be easier to define and read the codes associated to the errors. Also it will prevent the assignment of multiple codes to the same error

pkg/errors/wrapped.go
```diff
@@ -0,0 +1,83 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package errors
+
+type wrappedError struct {
```
jeroiraz, line 19: what's the relationship with immuError under pkg/errors? I'd expect immuError to be enough but there may be something I'm not taking into account

pkg/client/streams.go
mmeloni: _errors.New(message string)_ is the canonical way to trigger errors.
If we modify this signature it will not possible anymore to replace error standard package with the new "github.com/codenotary/immudb/pkg/errors"
I can add a map in which we store as a key the message and as a value the code.
This could help?

pkg/errors/wrapped.go
mmeloni: wrappedError is required to implement wrapping capability and it contains the "cause" property.
immuError could be the cause of a wrappedError or can be returned directly.

pkg/client/streams.go
mmeloni: fixed with the internal map as discussed

jeroiraz (approved):

PR #802, opened 2021-05-28
move server UUID file to data dir

### LINKED ISSUE #799: Server UUID in data root path

Server UUID file is created under `defaultDB` path but it should be created in root data folder instead.

**Why is this needed**

Server UUID is cross-database information and thus it shouldn't be placed inside any database path.

**Additional context**

File to relocated is `immudb.identifier`

Modification should automatically work with existent databases in which such file was already located under defaultdb folder.

mangalaman93:
the file used to reside in default DB dir. Because the file contains cross-database information, it should not be in default db dir.

Fixes #799


jeroiraz (approved):

jeroiraz:
Thanks a lot @mangalaman93

PR #815, opened 2021-05-31
chore(pkg/stream): fix namings on stream api objects

mmeloni:
Fix naming on stream api object in order to clarify the purpose of message sender and receiver parameters

### DIFF
--- a/pkg/stream/receiver.go
+++ b/pkg/stream/receiver.go
@@ -30,7 +30,7 @@ func NewMsgReceiver(stream ImmuServiceReceiver_Stream) *msgReceiver {
 }
 
 type MsgReceiver interface {
-	Read(message []byte) (n int, err error)
+	Read(data []byte) (n int, err error)
 }
 
 type msgReceiver struct {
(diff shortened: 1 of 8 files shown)


pkg/stream/receiver.go
```diff
@@ -43,7 +43,7 @@ type msgReceiver struct {
 }
 
 // Read read fill message with received data and return the number of read bytes or error. If no message is present it returns 0 and io.EOF. If the message is complete it returns 0 and nil, in that case successive calls to Read will returns a new message.
```
jeroiraz, line 45: It might be expected to return the number of read bytes, even if the buffer is successfully read

jeroiraz (approved):

pkg/stream/receiver.go
mmeloni: When multiple messages are readed by receiver `read` method returns  every times the number of readed bytes. If the message is fully readed the next call to `read` will return 0 or io.EOF in case of no more messages are present on the wire.

mmeloni: Probably the comment need to be improved

PR #822, opened 2021-06-02
fix(cmd/immuclient): remove warnings on sql commands in interactive mode

### LINKED ISSUE #803: Running SQL interactively in immuclient always returns unnecessary warning
**What happened**

immuclient>exec SELECT id FROM people
INFO: Redunant argument(s) | [id FROM people] 
sql ok, Ctxs: 0 Dtxs: 0

immuclient understands the command correctly and I can't find a different way to run the same. So I expect no warnings.

**Environment**

1.0.0 (both client and server)

byo:
This PR changes the `exec` and `query` commands to variable ones.
That way those commands can accept arbitrary number of arguments without issuing a warning in the interactive mode.

Closes  #803

### DIFF
--- a/cmd/immuclient/cli/register.go
+++ b/cmd/immuclient/cli/register.go
@@ -59,8 +59,8 @@ func (cli *cli) initCommands() {
 	cli.Register(&command{"version", "Print version", cli.version, nil, false})
 
 	// SQL
-	cli.Register(&command{"exec", "Executes sql statement", cli.sqlExec, []string{"statement"}, false})
-	cli.Register(&command{"query", "Query sql statement", cli.sqlQuery, []string{"statement"}, false})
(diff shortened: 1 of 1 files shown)


mmeloni (approved):

dmacvicar (commented):
How does this play with the ability to use double quotes for the full sql statement? related, unrelated?

jeroiraz (approved):
I'd expect to require sql statements to be quoted.
This change may be done after this fix as well

byo:
> How does this play with the ability to use double quotes for the full sql statement? related, unrelated?

This change does not yet introduce quoted arguments in the interactive mode, added a card for that specific feature: https://github.com/orgs/codenotary/projects/5#card-62319658

jeroiraz (approved):

mmeloni (approved):

Issue #823, opened 2021-06-02
Epic: Self healing on startup


dmacvicar:
If for some reason (unexpected shutdown and/or not running in sync mode) the index and/or the hash tree need to be regenerated, immudb should detect this situation at startup and correct it, by regenerating everything before starting up.

Output is important, so this recovery mode should tell this is happening.

Other cards which can be split out from this:

* While optimizations are important, the first step is to recover, so finding out where to truncate the index and/or trees to start over is a nice to have.
* For monitoring purposes, it may be interesting to expose if the server is running on recovery or not, as a metric. ([example from other software](https://docs.rudderstack.com/user-guides/administrators-guide/monitoring-and-metrics#recovery-mode))

PR #825, opened 2021-06-02
Apply synced settings to indexing data

jeroiraz:
This PR applies global `synced` configuration to indexing data structure in order to prevent inconsistencies during unexpected/forced terminations.

While some performance degradation may be expected, it shouldn't be much noticeable to final applications given indexing data is eventually flushed to disk. Furthermore, depending on the use case, eventual indexing feature can be used per transaction basis by setting attribute `NoWait=true`.

This approach may be useful until a more elaborated solution is implemented i.e. #823

### DIFF
--- a/embedded/store/immustore.go
+++ b/embedded/store/immustore.go
@@ -408,7 +408,7 @@ func OpenWith(path string, vLogs []appendable.Appendable, txLog, cLog appendable
 		WithFileMode(opts.FileMode).
 		WithLog(opts.log).
 		WithFileSize(fileSize).
-		WithSynced(opts.Synced). // built from derived data, but temporarily to reduce chances of data inconsistencies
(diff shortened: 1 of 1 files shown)


embedded/store/immustore.go
```diff
@@ -408,7 +408,7 @@ func OpenWith(path string, vLogs []appendable.Appendable, txLog, cLog appendable
 		WithFileMode(opts.FileMode).
 		WithLog(opts.log).
 		WithFileSize(fileSize).
-		WithSynced(false). // index is built from derived data
+		WithSynced(opts.Synced). // built from derived data, but temporarily to reduce chances of data inconsistencies
```
byo, line 411: I don't fully understand the comment, shouldn't be something like:

```suggestion
		WithSynced(opts.Synced). // built from derived data, but temporarily modified to reduce chances of data inconsistencies until a better solution is implemented
```

PR #828, opened 2021-06-03
Add /initz, /readyz, /livez and /version endpoints for immudb and immuclient auditor

padurean:
[This issue](https://github.com/vchain-us/ledger-compliance/issues/735) contains more details.

ℹ️ `/initz` is intended to be used for startup probes, `/readyz` for readiness probes and `/livez` for liveness probes.

~There is an open point though:~
- ~❓ **what checks should be done** in the code which is executed for `GET` `initz`, `/readyz` and `livez` (of immudb) in order **to determine if immudb is healthy or not?**~

**UPDATE**: right now these endpoints will always return `200 OK`.

ℹ️ For auditor's `GET` `/initz`, `/readyz` and `/livez` (which are implemented identically right now), a `Health` command is issued to immudb and if there's no error and response contains `Status` `true`, then _200 OK_ is responded, otherwise _503 Service Unavailable_.


mmeloni:
Hi @padurean .
From what I read [healtz endpoint is deprecated on kube](https://kubernetes.io/docs/reference/using-api/health-checks/) . So I think we should implement livez and readyz.

In order to determine if immudb is ready to accept queries I think we need to check if there is no index and hashtree regeneration ongoing.

padurean:
@mmeloni Thanks for your feedback.
@SimoneLazzaris @dmacvicar Let's clarify this. Please have a look at Michele's and my own comments above.
Thanks.

padurean:
@mmeloni Regarding the index and hashtree regeneration, i am not sure if it would be a good check that they are not running: i assume the container orchestrator might destroy the container if it reports "unhealthy" beacause of that.
This needs to be also clarified, @SimoneLazzaris @dmacvicar

SimoneLazzaris:
Regarding the meaning of the endpoints, this is what kubernetes needs to fully probe a service:
 

-     one endpoint that returns 200 once the initialization phase is ended
-     one endpoint that triggers a container reboot when a non-200 code is returned (or if it fails)
-     one endpoint that blocks incoming traffic if a non-200 code is returned.


The first (startup probe) one is used to coordinate the startup process: if process B depends on process A, kubernetes won't start process B until process A has completed the startup phase.

Second one (liveness probe) is used to detect a dead/stuck container. It should trigger some tests that guarantee that the process is really alive, even if it's not responding to "normal" requests due to (for example) high load. We don't have something for that, I believe.

Third one (readiness) is used to allow/block incoming traffic until the process is ready to process it.

I think that for immudb we can use a single endpoint that function both as a startup and a ready indicator, or (better) two different endpoints that have for now the same semantic.

If/when some sort of parallelism is implemented  (let's say that we have n immudb instances, in a multi master configuration) we can use a dedicated readiness endpoint to temporarily divert incoming traffic away from an overloaded instance.

I'd like to have the two endpoints reserved, and route them (at the moment) to a single functions, allowing us to specialize the behavior in a second moment.

Let's say something like
- /livez
- /readyz (used also for startup)

or
- /livez
- /readyz 
- /initz (used for startup)

padurean:
@mmeloni @jeroiraz 
Regarding this comment from Michele:
> In order to determine if immudb is ready to accept queries I think we need to check if there is no index and hashtree regeneration ongoing.

1. **For which type of probes should this be used?**
   - [ ] startup probes
   - [ ] readiness probes
   - [ ] liveness proves

2. **Which is the way to determine this** (i.e. if an index and hashtree regeneration is ongoing)**?**

SimoneLazzaris (approved):

padurean:
ℹ️ **UPDATE** regarding [my previous comment](#issuecomment-857619102) and questions:
for the moment the `GET` `/initz`, `/readyz` and `/livez` of immudb **will always return `200 OK`** (i.e. the index regeneration does not influence the response code in anyway for now).

padurean:
@jeroiraz @SimoneLazzaris @mmeloni Is it ok if i merge this now? Thanks.

SimoneLazzaris:
I'm ok with merging this _as is_, so we can start rewrite kubernetes deployment to use the new endpoint.

I'm still thinking that in order to be effective, we need to return a non-200 code when some operation are refused.
More precisely, if reindexing is needed upon startup, then the /readyz endpoint should return a non-200 code until ready: it's best to delay the startup of other components in that condition. We can also program a long startup timeout to handle that.

For compaction/reindexing _once started_ I think that we can stick to the 200 code _for now_

SimoneLazzaris (approved):

mmeloni (dismissed):
Agree with @SimoneLazzaris . I think you have already  complete the infrastructural part and now we should expose an api on the storage level in order to understand if something is reindexing/compacting, and bind it.  I think we need @jeroiraz feedback

jeroiraz (approved):

PR #832, opened 2021-06-05
remove additional blank line from immuclient output

### LINKED ISSUE #829: Remove additional blank line from immuclient output
**What happened**

Output generated from immuclient commands includes an additional blank line

e.g.
```
> ./immuclient set k0 v0
tx:		1 
key:		k0 
value:		v0 
hash:		6930a890518dad8804ffcee19d0c0759c4a31040612db8ea12beb7f7231047a0 

>
```

```
>./immuclient set k0 v0
tx:		1 
key:		k0 
value:		v0 
hash:		6930a890518dad8804ffcee19d0c0759c4a31040612db8ea12beb7f7231047a0
>
```

mangalaman93:
Fixes #829

### DIFF
--- a/cmd/immuclient/command/cmd.go
+++ b/cmd/immuclient/command/cmd.go
@@ -18,6 +18,7 @@ package immuclient
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"strings"
 
@@ -75,4 +76,13 @@ func commandNames(cms []*cobra.Command) []string {
 		args = append(args, arg)
 	}
 	return args
+}
+
+// fprintln is equivalent to fmt.Fprintln but appends newline only if one doesn't exist.
+func fprintln(w io.Writer, msg string) {
+	if strings.HasSuffix(msg, "\n") {
+		_, _ = fmt.Fprint(w, msg)
+	} else {
+		_, _ = fmt.Fprintln(w, msg)
+	}
 }
--- a/cmd/immuclient/command/currentstatus.go
+++ b/cmd/immuclient/command/currentstatus.go
@@ -17,8 +17,6 @@ limitations under the License.
 package immuclient
 
 import (
-	"fmt"
-
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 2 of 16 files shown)


cmd/immuclient/command/cmd.go
```diff
@@ -76,3 +77,14 @@ func commandNames(cms []*cobra.Command) []string {
 	}
 	return args
 }
+
+// fprintln is equivalent to fmt.Fprintln but appends new line only if one doesn't exist.
+func fprintln(w io.Writer, msg string) {
```
jeroiraz, line 82: what do you think about printing original input if it ends it a line break and use `fmt.Fprintln` with original input otherwise?

mangalaman93: yeah, I think I can figure out a better solution for double newline case. Let me look.

mangalaman93: I have made the returned value consistent everywhere with a newline now. This way, we don't have to do a check and figure out whether to append a newline. Please have a look.

cmd/immuclient/command/currentstatus.go
```diff
@@ -34,7 +34,7 @@ func (cl *commandline) currentState(cmd *cobra.Command) {
 			if err != nil {
 				cl.quit(err)
 			}
-			fmt.Fprintf(cmd.OutOrStdout(), resp+"\n")
```
jeroiraz, line 37: I think your initial approach of having a print function which adds a line break if not already included is good. But I'd do it as simple as possible i.e. adding if the last char is not a line break.
Note: with latest changes a unit testing is failing, probably because the error message differs from what's expected.

mangalaman93: Fixed the unit test. I was able to remove the `\n\n` branch but in the process I realized that if we have already done the work of putting right newline characters in place, we don't need to go through another step to check and append a newline character. I don't think it is a big deal in any case, though, I think we don't need to do it anymore. Let me know what you think.

jeroiraz (approved):

PR #837, opened 2021-06-14
Feat replication

jeroiraz:
This PR provides the infrastructure needed to replicate databases.

When a database is created (and after its creation) it can be set to be a replica and only apply changes using replicated commits.

`Maintenance mode (read-only + recovery)`:
    read-only + recovery operations over any database  (./immudb --maintenance)
    if authentication is disabled, possibility to restore systemdb and defaultdb (./immudb --maintenance --auth=false)

    Please note, when systemdb is being restored, no user created databases will be loaded

`Permission changes`:
    Possibility to assign read-only permissions to systemdb (so to be able to replicate)

    Note: replicate operation is only possible when systemdb and defaultdb are initiated in recovery mode.

*Note* User management operations will not be allowed in maintenance mode


pkg/database/sql.go
```diff
@@ -325,7 +369,14 @@ func (d *db) SQLQueryPrepared(stmt *sql.SelectStmt, namedParams []*schema.NamedP
 	cols := make([]*schema.Column, len(colDescriptors))
 
 	for i, c := range colDescriptors {
-		cols[i] = &schema.Column{Name: c.Selector, Type: c.Type}
+		des := &sql.ColDescriptor{
```
byo, line 372: Why it is now needed to rewrite those selectors?

pkg/auth/permissions.go
```diff
@@ -56,6 +56,8 @@ var methodsPermissions = map[string][]uint32{
 	"StreamHistory":          {PermissionSysAdmin, PermissionAdmin, PermissionRW, PermissionR},
 	"TxByID":                 {PermissionSysAdmin, PermissionAdmin, PermissionRW, PermissionR},
 	"TxScan":                 {PermissionSysAdmin, PermissionAdmin, PermissionRW, PermissionR},
+	"ExportTx":               {PermissionSysAdmin, PermissionAdmin, PermissionRW, PermissionR},
```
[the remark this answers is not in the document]
jeroiraz: it's not needed but enough. `ExportTx` has the same permissions spec as `Get` operation

pkg/database/database.go
```diff
@@ -177,32 +219,30 @@ func NewDb(op *DbOptions, catalogDB DB, log logger.Logger) (DB, error) {
...
 	if err != nil {
 		return nil, logErr(dbi.Logger, "Unable to open store: %s", err)
 	}
 
-	_, _, err = dbi.sqlEngine.ExecPreparedStmts([]sql.SQLStmt{&sql.CreateDatabaseStmt{DB: dbi.options.dbName}}, nil, true)
-	if err != nil {
-		return nil, logErr(dbi.Logger, "Unable to open store: %s", err)
-	}
+	if !op.replicationOpts.Replica {
+		_, _, err = dbi.sqlEngine.ExecPreparedStmts([]sql.SQLStmt{&sql.CreateDatabaseStmt{DB: dbInstanceName}}, nil, true)
```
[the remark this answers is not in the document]
jeroiraz: In order to have an exact match with the source database, this initial transaction needs to be replicated as any other transaction

pkg/database/sql.go
jeroiraz: good point, the sql catalog and data are stored in the same store so to simplify replication and state validation.  So in order to support replicas with a different name than the source database, there must be an isolation from database name and internal database name, the solution included in this PR is the following: databases receive a fixed name `dbInstance` as there is a single database created in the embedded storage, but in order to return result sets with a different database name, selectors needs to provide a structured data type for its manipulation without incurring in string interpretation

byo: OK, was not aware that the replication can be done to a db with a different name than the original (or did I understand this incorrectly and there's some special use-case where the replication needs to rename DBs?).

If only the db name needs to change, why not:

```suggestion
		des := *c
		des.Database = d.options.dbName
```
?

That way any future modifications to the `ColDescriptor` struct will be much easier.

PR #840, opened 2021-06-18
Feat/pgsql wire ext query

mmeloni:


pkg/pgsql/server/fmessages/bind.go
```diff
@@ -0,0 +1,150 @@
...
+	if len(parameterFormatCodes) == 0 {
+		forceTXT = true
+	}
+	if len(parameterFormatCodes) == 1 {
+		if f, ok := parameterFormatCodes[0]; ok {
+			switch f {
+			case 0:
+				forceTXT = true
+			case 1:
+				forceBIN = true
```
byo, line 96: What about other values of `f`? Shouldn't it return an error on `default` case?

pkg/pgsql/server/fmessages/bind.go
```diff
@@ -0,0 +1,150 @@
...
+	preparedStatement, err := getNextString(r)
+	if err != nil {
+		return BindMsg{}, err
+	}
+
+	parameterFormatCodeNumber, err := getNextInt16(r)
+	if err != nil {
+		return BindMsg{}, err
+	}
+	parameterFormatCodes := make(map[int]int16, 0)
```
byo, line 70: Is a map needed here? We're storing those ints sequentially  so a slice should be enough

pkg/pgsql/server/fmessages/bind.go
```diff
@@ -0,0 +1,150 @@
...
+		if forceBIN {
+			params = append(params, pVal)
+			continue
+		}
+		if f, ok := parameterFormatCodes[i]; ok {
+			switch f {
+			case 0:
+				params = append(params, string(pVal))
+			case 1:
+				params = append(params, pVal)
```
byo, line 125: Shouldn't we handle the `default` case here as well ?

pkg/pgsql/server/fmessages/bind.go
```diff
@@ -0,0 +1,150 @@
...
+	if len(parameterFormatCodes) == 1 {
+		if f, ok := parameterFormatCodes[0]; ok {
+			switch f {
+			case 0:
+				forceTXT = true
+			case 1:
+				forceBIN = true
+			}
+		}
+	}
```
byo, line 99: Maybe it would be good to check if `len(parameterFormatCodes) == pCount` if it's not 0 or 1? I guess it would indicate a malformed packet.

pkg/pgsql/server/fmessages/bind.go
```diff
@@ -0,0 +1,150 @@
...
+
+	resultColumnFormatCodesNumber, err := getNextInt16(r)
+	if err != nil {
+		return BindMsg{}, err
+	}
+
+	resultColumnFormatCodes := make([]int16, 0)
```
byo, line 135:
```suggestion
	resultColumnFormatCodes := make([]int16, 0, resultColumnFormatCodesNumber)
```

pkg/pgsql/server/fmessages/describe.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+func ParseDescribeMsg(msg []byte) (DescribeMsg, error) {
+	descType := msg[0]
+	name := make([]byte, len(msg)-2)
+	copy(name, msg[1:len(msg)-1])
+	return DescribeMsg{
+		DescType: string(descType),
+		Name:     string(name),
+	}, nil
```
byo, line 33: Why an extra copy is needed? Wouldn't this be enough?

```suggestion
	return DescribeMsg{
		DescType: string(descType),
		Name:     string(msg[1:len(msg)-1]),
	}, nil
```

pkg/pgsql/server/stmts_handler.go
```diff
@@ -0,0 +1,36 @@
+package server
+
+import (
+	"regexp"
+)
+
+func (s *session) isInBlackList(statement string) bool {
+	var set = regexp.MustCompile(`(?i)set\s+.+`)
```
byo, line 8: Compiling regex may be a bit heavy operation if done on each statement -  can we move this to global var?

pkg/pgsql/server/stmts_handler.go
```diff
@@ -0,0 +1,36 @@
...
+		return true
+	}
+	if statement == ";" {
+		return true
+	}
+	return false
+}
+
+func (s *session) isEmulableInternally(statement string) interface{} {
+	if regexp.MustCompile(`(?i)select\s+version\(\s*\)`).MatchString(statement) {
```
byo, line 19: can we move the result of `regexp.MustCompile` to a global var?

pkg/pgsql/server/bmessages/parameter_description.go
```diff
@@ -0,0 +1,59 @@
...
+//Int32
+//Specifies the object ID of the parameter data type.
+// ParameterDescription send a parameter description message. Cols need to be lexicographically ordered by selector
+func ParameterDescription(cols []*schema.Column) []byte {
+	// Identifies the message as a run-time parameter status report.
+	messageType := []byte(`t`)
+	selfMessageLength := make([]byte, 4)
+
+	paramsNumberB := make([]byte, 2)
+	binary.BigEndian.PutUint16(paramsNumberB, uint16(len(cols)))
```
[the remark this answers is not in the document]
mmeloni: Ok I added a guard for this here, 08c76a8c87226c10fc79b0f04fd62ee0e4c5b350

embedded/sqlutils/params.go
```diff
@@ -0,0 +1,66 @@
+package sqlutils
```
jeroiraz, line 1: I'd move under api/schema package, otherwise it would introduce a dependency to the embedded one

pkg/pgsql/server/fmessages/string_reader.go
```diff
@@ -0,0 +1,48 @@
...
+
+package fmessages
+
+import (
+	"bufio"
+	"encoding/binary"
+)
+
+func getNextString(r *bufio.Reader) (string, error) {
+	s, err := r.ReadBytes(0)
```
jeroiraz, line 25: is the intention to fully read the data as a string?

jeroiraz (changes requested):

pkg/pgsql/server/fmessages/string_reader.go
mmeloni: yes, it's used to parse client messages

pkg/pgsql/server/fmessages/bind.go
```diff
@@ -0,0 +1,154 @@
...
+		return BindMsg{}, errors.New("malformed bind message. Parameters format codes didn't match parameters count")
+	}
+
+	params := make([]interface{}, 0)
+	for i := 0; i < int(pCount); i++ {
+		pLen, err := getNextInt32(r)
+		if err != nil {
+			return BindMsg{}, err
+		}
+		pVal := make([]byte, pLen)
```
byo, line 109: `pLen` can be of any size including negative values (that would panic here) or very high ones (allocation of huge memory buffer), please sanitize that value

pkg/pgsql/server/fmessages/bind.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	// The number of parameter format codes that follow (denoted C below).
+	// This can be zero to indicate that there are no parameters or that the parameters all use the default format (text);
+	// or one, in which case the specified format code is applied to all parameters; or it can equal the actual number
+	// of parameters.
+	parameterFormatCodeNumber, err := getNextInt16(r)
+	if err != nil {
+		return BindMsg{}, err
+	}
+	parameterFormatCodes := make([]int16, parameterFormatCodeNumber)
```
byo, line 68: `parameterFormatCodeNumber` can be negative here causing a panic. Please sanitize it.

pkg/pgsql/server/fmessages/bind.go
```diff
@@ -0,0 +1,154 @@
...
+	// The number of result-column format codes that follow (denoted R below).
+	// This can be zero to indicate that there are no result columns or that the result columns should all use the
+	// default format (text); or one, in which case the specified format code is applied to all result columns (if any);
+	// or it can equal the actual number of result columns of the query.
+	resultColumnFormatCodesNumber, err := getNextInt16(r)
+	if err != nil {
+		return BindMsg{}, err
+	}
+
+	resultColumnFormatCodes := make([]int16, 0, resultColumnFormatCodesNumber)
```
byo, line 139: `resultColumnFormatCodesNumber` can be negative here causing a panic. Please sanitize it.

byo (changes requested):

pkg/pgsql/server/query_machine.go
```diff
@@ -307,30 +308,33 @@ func (s *session) inferParamAndResultCols(statement string) ([]*schema.Column, [
...
-	cols, err := rr.Columns()
-	if err != nil {
-		return nil, nil, err
-	}
 	resCols := make([]*schema.Column, 0)
-	for _, c := range cols {
-		resCols = append(resCols, &schema.Column{Name: c.Selector, Type: c.Type})
+
+	sel, ok := stmt.(*sql.SelectStmt)
+	if ok {
```
byo, line 326: Just double-checking - this change does enable extended query support for non-query statements, right ?

pkg/pgsql/server/fmessages/bind.go
```diff
@@ -0,0 +1,168 @@
...
+			params = append(params, pVal)
+			continue
+		}
+
+		switch parameterFormatCodes[i] {
+		case 0:
+			params = append(params, string(pVal))
+		case 1:
+			params = append(params, pVal)
+		}
```
byo, line 139: This got lost somewhere along other comments, a check similar to the one on  [line 98](https://github.com/codenotary/immudb/pull/840/files#diff-7841e1c8241c390dd964fe42839fca3f6b0023e74dee5052b57cf34fcb5d1d80R98) is needed here too:

```suggestion
		default:
			return BindMsg{}, pgserrors.ErrMalformedMessage
		}
```

pkg/pgsql/server/query_machine.go
byo: (it's about this commit: https://github.com/codenotary/immudb/pull/840/commits/9cf09489719b6e9d10c207ec73da88b256884d56#diff-3fbf4583347647005ea75130017f5707f4957202e094cdf7849478d81eeeb0d5R326)

mmeloni: exactly

byo (dismissed):
Good job !

PR #843, opened 2021-06-24
typed parameters inference in sql stmts

jeroiraz:
This PR implements parameters and type checking in SQL statements.
While main reason for this work was due to the requirement to implement extended query support of psql protocol, it may be useful to statically detect inconsistencies in SQL statements.

New operations added to database interface under `pkg/database` package:
- `InferParameters(sql string) (map[string]sql.SQLValueType, error)`
- `InferParametersPrepared(stmt sql.SQLStmt) (map[string]sql.SQLValueType, error)`

At embedded SQL Engine level the following operations has been added:
- `InferParameters(sql string) (map[string]SQLValueType, error)`
- `InferParametersPreparedStmt(stmt SQLStmt) (map[string]SQLValueType, error)`

e.g.
```
params, err = engine.InferParameters("SELECT * FROM mytable WHERE active AND @active")
require.NoError(t, err)
require.Len(t, params, 1)
require.Equal(t, BooleanType, params["active"])
```


embedded/sql/stmt.go
```diff
@@ -845,6 +1092,36 @@ func (stmt *SelectStmt) Limit() uint64 {
...
+	return params, nil
+}
+
+func (stmt *SelectStmt) inferParameters(e *Engine, implicitDB *Database, params map[string]SQLValueType) error {
+	_, _, _, err := stmt.CompileUsing(e, implicitDB, nil)
+	if err != nil {
+		return err
+	}
+
+	snapshot, err := e.Snapshot()
```
byo, line 1112: Is it guaranteed that this snapshot will be the same when executing the query? What if the schema changes in the meantime?

embedded/sql/stmt.go
```diff
@@ -93,6 +94,8 @@ const (
 type SQLStmt interface {
 	isDDL() bool
 	CompileUsing(e *Engine, implicitDB *Database, params map[string]interface{}) (ces, des []*store.KV, db *Database, err error)
+	InferParameters(e *Engine) (params map[string]SQLValueType, err error)
```
mmeloni, line 97: I think we should provide InferParameter from database package.
Using engine is not pratical since server is using database package to expose api.

mmeloni (changes requested):
We should use new error package, isnt?

embedded/sql/stmt.go
jeroiraz: this is the internal implementation, but the upper layers will expose it as you describe

embedded/sql/stmt.go
jeroiraz: good point. Currently there is no catalog changes that could affect inferred types but even in such case, the query resolution may raise errors due to types mismatch. Annotated this topic so to be accordingly analyzed.

mmeloni (dismissed):

PR #846, opened 2021-06-28
fix(pkg/server): lock userdata map read

mmeloni:


pkg/server/user.go
```diff
@@ -559,6 +559,8 @@ func (s *ImmuServer) addUserToLoginList(u *auth.User) {
 }
 
 func (s *ImmuServer) getLoggedInUserdataFromCtx(ctx context.Context) (int64, *auth.User, error) {
+	s.userdata.Lock()
```
jeroiraz, line 562: sync may be done in `getLoggedInUserDataFromUsername` function

mmeloni: done

jeroiraz (approved):

PR #859, opened 2021-07-19
Test/coverage increase

byo:


embedded/store/indexer.go
```diff
@@ -312,6 +312,7 @@ func (idx *indexer) doIndexing(cancellation <-chan struct{}) {
 		idx.stateCond.L.Lock()
 		for {
 			if idx.state == stopped {
+				idx.stateCond.L.Unlock()
```
byo, line 315: @jeroiraz I believe this unlock is missing here, can you confirm?

jeroiraz: yes, for sake of completeness but it shouldn't affect any functionality

jeroiraz (approved):

PR #861, opened 2021-07-20
Chore/coverage

SimoneLazzaris:
I've added some test function to cover corner cases

### DIFF
--- a/cmd/helper/error.go
+++ b/cmd/helper/error.go
@@ -25,10 +25,12 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+var osexit = os.Exit
+
 // QuitToStdErr prints an error on stderr and closes
 func QuitToStdErr(msg interface{}) {
 	_, _ = fmt.Fprintln(os.Stderr, msg)
-	os.Exit(1)
+	osexit(1)
 }
 
 // QuitWithUserError ...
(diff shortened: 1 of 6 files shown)


cmd/helper/error.go
```diff
@@ -25,10 +25,12 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+var osexit = os.Exit
```
byo, line 28: This looks like a quite intrusive change since - IMO we should first try to increase the code coverage without altering the immudb code and only if that won't bring us above certain % threshold then use more intrusive methods.

Can you please extract tests that require this `osexit` into a separate PR? That way we'll still be able to merge the majority of this PR.

jeroiraz: However, it's on the command line components, if there is no easy way for dealing with it, I think it'd be ok

SimoneLazzaris: Unfortunately, there is a relatively big chunk of code that fails with "os.Exit", which makes it impossible to account for with coverage. Hence my modification, which is really very localized, but I can see your point.

Of course I can extract those tests into a different PR.

byo: Ok, as Jeronimo said, I guess it's acceptable - there's likely no other way to test this code.

byo (approved):

byo (approved):
Good job! We're green on coverage now!

PR #862, opened 2021-07-20
initialize db settings if not present

jeroiraz:


pkg/server/server_test.go
```diff
@@ -401,6 +402,21 @@ func TestServerLoaduserDatabase(t *testing.T) {
 	}
 }
 
+func TestServerLoadUserDatabases(t *testing.T) {
+	copier := fs.NewStandardCopier()
+	require.NoError(t, copier.CopyDir("../../test/data_v1.0.1", "data_v1.0.1"))
```
byo, line 407: Wouldn't it be better to have a separate dataset just for this failure case?
Simething like `test/data_v1.0.1_dbsettings_test`

jeroiraz: it could be, but actually its the migration needed for any db up to v1.0.1.

byo (approved):

PR #864, opened 2021-07-21
fix(sql/engine): Harden DecodeValue

byo:

### DIFF
--- a/embedded/sql/engine.go
+++ b/embedded/sql/engine.go
@@ -957,7 +957,7 @@ func DecodeValue(b []byte, colType SQLValueType) (TypedValue, int, error) {
 	vlen := int(binary.BigEndian.Uint32(b[:]))
 	voff := EncLenLen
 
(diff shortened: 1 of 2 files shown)


embedded/sql/engine.go
```diff
@@ -957,7 +957,7 @@ func DecodeValue(b []byte, colType SQLValueType) (TypedValue, int, error) {
 	vlen := int(binary.BigEndian.Uint32(b[:]))
 	voff := EncLenLen
 
-	if len(b) < vlen {
+	if vlen < 0 || len(b) < voff+vlen {
```
mmeloni, line 960: vlen should be always positive, isnt?

byo: On 32-bit architectures (IIRC raspberry pi is one), the decoded uint32 may actually convert to negative integer when casted to int.

mmeloni (approved):

PR #872, opened 2021-07-24
fix(pkg/client/auditor): fix and enhance state signature verification

padurean:
Signed-off-by: Valentin Padurean (Ogg) <912518+padurean@users.noreply.github.com>

### DIFF
--- a/pkg/client/auditor/auditor.go
+++ b/pkg/client/auditor/auditor.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/ecdsa"
+	"crypto/elliptic"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
@@ -269,14 +270,10 @@ func (a *defaultAuditor) audit() error {
 		return noErr
 	}
 
-	if a.serverSigningPubKey != nil {
-		if okSig, err := state.CheckSignature(a.serverSigningPubKey); err != nil || !okSig {
-			a.logger.Errorf(
-				"audit #%d aborted: could not verify signature on server state at %s @ %s",
-				a.index, serverID, a.serverAddress)
-			withError = true
-			return noErr
-		}
+	if err := a.verifyStateSignature(serverID, state); err != nil {
+		a.logger.Errorf("audit #%d aborted: %v", a.index, err)
+		withError = true
+		return noErr
 	}
 
 	isEmptyDB := state.TxId == 0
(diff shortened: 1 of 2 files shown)


pkg/client/auditor/auditor.go
```diff
@@ -299,6 +296,12 @@ func (a *defaultAuditor) audit() error {
 			return noErr
 		}
 
+		if err := a.verifyStateSignature(serverID, prevState); err != nil {
```
jeroiraz, line 299: I think this additional signature verification is not needed. Verifying signature before accepting state evolution should be enough.
Verification of local state will double the amount of signature verifications and will make it harder to deal with changing server signing keys. If verification only happens when a new state is validated, it's easier to deal with a new key by just changing the public key at the auditor.

pkg/client/auditor/auditor.go
```diff
@@ -380,6 +383,37 @@ func (a *defaultAuditor) audit() error {
...
+		return fmt.Errorf(
+			"a server signing public key has been specified for the auditor, "+
+				"but the state %s at TX %d received from server %s @ %s is not signed",
+			serverState.GetTxHash(), serverState.GetTxId(), serverID, a.serverAddress)
+	}
+
+	if serverState.GetSignature() != nil {
+		pk := a.serverSigningPubKey
+		if pk == nil {
+			x, y := elliptic.Unmarshal(elliptic.P256(), serverState.GetSignature().GetPublicKey())
```
byo, line 401: Can we issue some kind of warning here? Checking the signature here without any form of public key verification does not protect from man-in-the-middle attacks and may give a false sense of security.

pkg/client/auditor/auditor.go
padurean: Fixed: i removed the signature verification for the previous state.

pkg/client/auditor/auditor.go
```diff
@@ -380,6 +377,37 @@ func (a *defaultAuditor) audit() error {
...
```
mmeloni, line 395: The code here is binded with elliptic p256 curve type. 
We should use an abstraction.
Inside Signer package there is an ecdsa implentation, we should modify Signer interface and expose UnmarshalKey method and then use it there.

mmeloni (changes requested):

mmeloni (approved):

pkg/client/auditor/auditor.go
byo: @padurean We'll be merging the PR as is, please prepare another PR that fixes this and @mmeloni 's comments

PR #878, opened 2021-07-29
Fix/corrupted commit log

byo:


embedded/store/immustore.go
```diff
@@ -256,35 +257,36 @@ func OpenWith(path string, vLogs []appendable.Appendable, txLog, cLog appendable
 
 	fileSize, ok := metadata.GetInt(metaFileSize)
 	if !ok {
-		return nil, ErrCorruptedCLog
+		return nil, fmt.Errorf("corrupted commit log metadata (filesize): %w", ErrCorruptedCLog)
```
jeroiraz, line 260: the downside of this changes is that error comparison by declared error is lost. Probably logging but returning the declared error until a more elaborated error struct is implemented for the embedded package would be better?

jeroiraz (approved):

PR #879, opened 2021-07-29
non-blocking sql initialization

jeroiraz:
SQL Functionality requires the catalog to be loaded. Currently, this implies indexing be up to date.

In cases were a complete index regeneration was required, initialization was blocking. Depending on the amount of data, not being able to periodically call compaction may require an excessive amount of disk space.

This PR places SQL initialization into the background, thus operations which don't require the indexing to be up to date can be normally executed.

### DIFF
--- a/embedded/sql/engine.go
+++ b/embedded/sql/engine.go
@@ -113,11 +113,6 @@ func NewEngine(catalogStore, dataStore *store.ImmuStore, prefix []byte) (*Engine
 
 	copy(e.prefix, prefix)
 
-	err := e.LoadCatalog()
-	if err != nil {
-		return nil, err
-	}
-
 	return e, nil
 }
 
@@ -130,10 +125,17 @@ func (e *Engine) EnsureCatalogReady() error {
 		return nil
 	}
 
-	return e.LoadCatalog()
+	return e.loadCatalog()
 }
 
-func (e *Engine) LoadCatalog() error {
+func (e *Engine) ReloadCatalog() error {
+	e.catalogRWMux.Lock()
+	defer e.catalogRWMux.Unlock()
+
+	return e.loadCatalog()
+}
+
+func (e *Engine) loadCatalog() error {
 	lastTxID, _ := e.catalogStore.Alh()
 	err := e.catalogStore.WaitForIndexingUpto(lastTxID, nil)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


embedded/sql/engine.go
```diff
@@ -113,29 +113,31 @@ func NewEngine(catalogStore, dataStore *store.ImmuStore, prefix []byte) (*Engine
...
-	if err != nil {
-		return nil, err
-	}
-
 	return e, nil
 }
 
 // TODO (jeroiraz); this operation won't be needed with a transactional in-memory catalog
-func (e *Engine) EnsureCatalogReady() error {
+func (e *Engine) EnsureCatalogReady(cancellation <-chan struct{}) error {
```
byo, line 120: Not necessarily for this PR but I think we should start passing [`context.Context`](https://pkg.go.dev/context#Context) around - it seems to be the native way to send cancellation requests in golang.

embedded/tbtree/tbtree.go
```diff
@@ -1193,10 +1193,6 @@ func (t *TBtree) SnapshotSince(ts uint64) (*Snapshot, error) {
 		return nil, ErrorToManyActiveSnapshots
 	}
 
-	if t.compacting {
```
byo, line 1196: Why it had to be removed?

embedded/sql/engine.go
jeroiraz: sure

pkg/database/database.go
```diff
@@ -137,53 +139,90 @@ func OpenDb(op *DbOptions, systemDB DB, log logger.Logger) (DB, error) {
...
+
+	go func() {
+		defer dbi.sqlInit.Done()
+
+		dbi.Logger.Infof("Loading SQL Engine for database '%s' (replica = %v)...", op.dbName, op.replicationOpts.Replica)
+
+		err := dbi.initSQLEngine(systemDB)
+		if err != nil {
+			dbi.Logger.Errorf("Unable to load SQL Engine for database '%s' (replica = %v). %v", op.dbName, op.replicationOpts.Replica, err)
+			return
```
byo, line 153: What happens here? Is the database in an always-broken state until restart?

embedded/tbtree/tbtree.go
jeroiraz: to be able to load the catalog while compaction is in progress

pkg/database/database.go
jeroiraz: currently, if there is an error while initiating sql engine yes, there is no re-attempt

byo (approved):

Issue #873, opened 2021-07-26
Correctly close all components on immustore initialization failure


byo:
**What happened**

If there's an error initializing new immustore, not all paths have a correct shutdown code.

**What you expected to happen**

We should ensure that all appendables are corrrectly closed and that all assisting goroutines are finished.

**How to reproduce it (as minimally and precisely as possible)**

The `TestImmudbStoreEdgeCases` unit test explores various edge cases including immustore initialization failures. In most cases, mocked appendables don't specify the `CloseFn` implementation. However if appendables were correctly closed on failures, some of those `Close` functions should be called.

**Environment**
```shell
$ git rev-parse origin/master
f6ca022a003737253d6e97b290f2b23349dca5bd
```

PR #881, opened 2021-08-02
fix(embedded/store): use reserved concurrency slot for indexing

jeroiraz:
Indexing go-routine was't using the reserved concurrency slot allocated for that purpose. As consequence of this, it was competing for resources as a normal transaction. Slowing down indexing tasks and potentially blocking if max concurrency limit was reached with synchronous commits.

### DIFF
--- a/embedded/store/indexer.go
+++ b/embedded/store/indexer.go
@@ -29,6 +29,7 @@ type indexer struct {
 	path string
 
 	store *ImmuStore
+	tx    *Tx
 
 	index *tbtree.TBtree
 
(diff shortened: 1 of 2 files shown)


embedded/store/indexer.go
```diff
@@ -163,6 +170,7 @@ func (idx *indexer) Close() error {
 
 	idx.stop()
 	idx.wHub.Close()
+	idx.store.releaseAllocTx(idx.tx)
```
byo, line 173: Since releasing of the tx slot has now moved to the Close method, there will be a tx leak if the Close method is not called correctly. There's one case where I'm aware of missing closed slot: https://github.com/codenotary/immudb/issues/873 - is it going to be any issue ?

jeroiraz: I don't think it will present an issue in this point, but cited issue needs to be solved.
The indexer is closed only when the store is closed.

byo (approved):

PR #882, opened 2021-08-02
Test/coverage increase

byo:


embedded/ahtree/ahtree.go
```diff
@@ -74,6 +74,12 @@ type AHtree struct {
 	_digests [256 * sha256.Size]byte // pre-allocated array for writing digests
 }
 
+func bugOn(cond bool) {
```
byo, line 77: Influenced by linux kernel's [`BUG_ON`](https://kernelnewbies.org/FAQ/BUG)

jeroiraz: I think it's much better to return the error and delegate to the caller to decide how to handle it

byo: The idea was to only use this if it's easy to prove that a given condition would not happen. E.g. in the case of the cache object all checked errors can not happen because either there is some check before that rules the error out (like initializing with negative size) or an error can not happen because of the way the code is used (e.g. integer can not be automatically casted to a nil interface when used as a function argument).

Removed the change with bugOn from the PR since it increases the coverage a lot in other places.

jeroiraz: Understand, but panicking it's a quite strong decision making the component less re-usable, because this entire component could be optional functionality in some cases.

Other approaches like creating the component by providing already built caches would provide both benefits

jeroiraz (dismissed):

PR #883, opened 2021-08-02
Improve signature verification during audit

padurean:
This PR addresses the changes requested by @byo & @mmeloni in the previous PR which was already merged - see these comments for details:
- [x] https://github.com/codenotary/immudb/pull/872#discussion_r676320188
- [x] https://github.com/codenotary/immudb/pull/872#discussion_r676347176

@mmeloni Please not that i actually did something a bit different than what you suggested (to avoid exposing `ecdsa.PublicKey` in generic `Signer` interface): i just called in the auditor code the existing `UnmarshalKey` function from signer package (which i enhanced a bit by returning an error if the key bytes are not a valid key).

### DIFF
--- a/pkg/client/auditor/auditor.go
+++ b/pkg/client/auditor/auditor.go
@@ -20,7 +20,6 @@ import (
 	"bytes"
 	"context"
 	"crypto/ecdsa"
-	"crypto/elliptic"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
@@ -38,6 +37,7 @@ import (
 	"github.com/codenotary/immudb/pkg/client/state"
 	"github.com/codenotary/immudb/pkg/client/timestamp"
 	"github.com/codenotary/immudb/pkg/logger"
+	"github.com/codenotary/immudb/pkg/signer"
 	"github.com/golang/protobuf/ptypes/empty"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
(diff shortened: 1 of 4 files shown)


pkg/signer/ecdsa_test.go
```diff
@@ -114,7 +119,16 @@ func TestSignature_VerifyFalse(t *testing.T) {
...
+	ok, err := Verify(rawMessage[:], m, ecdsaPK)
 	assert.False(t, ok)
 	assert.NoError(t, err)
 }
+
+func TestUnmarshalKey_Error(t *testing.T) {
+	ecdsaPK, err := UnmarshalKey([]byte(`wrongkey`))
+	require.Nil(t, ecdsaPK)
+	require.Error(t, err)
+	require.Equal(t, "invalid public key", err.Error())
```
byo, line 133: Once converted to an variable, please check if the error is returned with:

```suggestion
	require.ErrorIs(t, err, ErrInvalidPublicKey)
```

byo (changes requested):

pkg/client/auditor/auditor.go
```diff
@@ -392,8 +392,17 @@ func (a *defaultAuditor) verifyStateSignature(
...
-			pk = &ecdsa.PublicKey{Curve: elliptic.P256(), X: x, Y: y}
+			a.logger.Warningf(
+				"server signature will be verified using untrusted public key (embedded in the server state payload) " +
+					"- for better security please configure a public key for the auditor process")
+			var err error
+			pk, err = signer.UnmarshalKey(serverState.GetSignature().GetPublicKey())
+			if err != nil {
+				return fmt.Errorf(
+					"failed to verify signature for state %s at TX %d received from server %s @ %s: "+
+						"error unmarshaling the public key embedded in the server state payload: %v",
```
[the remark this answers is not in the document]
padurean: @byo Indeed: i use `%w` when it makes sense for the caller to check the error type with `errors.Is`, otherwise - e.g. if the caller only logs the error, like in this case - i see no point in wrapping it. Unless it is considered a good practice to wrap each and every error returned. In this case we should replace each and every occurrence of `%v` for errors with `%w` throughout the whole codebase.

pkg/signer/ecdsa_test.go
padurean: @byo This suggestion is a bit incorrect: the 2nd param is the expected error, the actual one is the 3rd. Otherwise i agree - will change right now.

byo: Yup, The `ErrorIs` is a bit inconsistent with the rest of `testify`, I guess it's because they want to be consistend with the `errors.Is` signature.

pkg/client/auditor/auditor.go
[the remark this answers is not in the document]
byo: Ah, true, forgot that this is within the auditor binary itself. Leaving the decision up to you then - maybe better to stay consistent with the rest of auditor's code?

pkg/signer/ecdsa_test.go
padurean: I've resolved all your suggestions.

byo (approved):
Good job :+1:

PR #885, opened 2021-08-05
some code refactoring to implement follower replication

jeroiraz:
Changes in this PR are required to avoid circular references when implementing replication in server package


pkg/database/dboptions.go
```diff
@@ -18,19 +18,19 @@ package database
...
-type DbOptions struct {
+//Options database instance options
+type Options struct {
 	dbName     string
 	dbRootPath string
 	storeOpts  *store.Options
 
 	corruptionChecker bool
 
+	replica         bool
```
byo, line 29: Why this has been moved from `ReplicationOptions` ?

pkg/integration/signature_verifier_interceptor_test.go
```diff
@@ -1,21 +1,38 @@
-package client
+/*
```
byo, line 1: :+1:

We could think of some gh action to check if such copyright header is present in all go files.

byo (approved):

pkg/database/dboptions.go
jeroiraz: replication code will be hold under it's own package i.e. pkg/replication.
Otherwise circular references between client and database package will arise. It's the server who coordinate replication for each database.
Database settings and users are currently stored in systemdb. Thus the server have the information required for it.
I think we may end up moving more things to the database and further removing the need of systemdb...

byo: Ahh, so the next step would be to move the `replicationOpts` out of the db options struct, now I get it.

So the only reason why the `replica` field is still in the `Options` struct is to avoid initializing any heavy machinery such as the sql engine if the database is not the primary one, right?

jeroiraz: yes. the database behaves differently if it's a replica or not e.g. read-only + tx replication (replica), vs read-write - tx replication (no replica). In addition to the initialization of the sql engine, where an initial statement creating a singleton database is executed (no replica) or wait to be replicated from master.

But the database is agnostic regarding how replication is done, if there is an external replicator or an embedded one. The embedded one managed by the server based on the db settings e.g. if master address was properly configured or not.

PR #890, opened 2021-08-12
fix(embedded/sql): enforce ordering by grouping column

jeroiraz:
This PR enforces all rows of the same group are consecutive so to get up most one entry per group.

As a consequence of this enforcement, grouping is restricted to the indexed column the results are ordered by.

```sql
CREATE TABLE t1(id INTEGER AUTO_INCREMENT, val1 INTEGER, PRIMARY KEY id)
CREATE INDEX ON t1(val1)
SELECT COUNT() as c FROM t1 GROUP BY val1 ORDER BY val1
```

### DIFF
--- a/embedded/sql/grouped_row_reader.go
+++ b/embedded/sql/grouped_row_reader.go
@@ -31,8 +31,13 @@ type groupedRowReader struct {
 }
 
 func (e *Engine) newGroupedRowReader(rowReader RowReader, selectors []Selector, groupBy []*ColSelector) (*groupedRowReader, error) {
-	if rowReader == nil || len(selectors) == 0 {
+	if rowReader == nil || len(selectors) == 0 || len(groupBy) > 1 {
 		return nil, ErrIllegalArguments
+	}
+
+	if len(groupBy) == 1 &&
+		rowReader.OrderBy().Selector() != EncodeSelector(groupBy[0].resolve(rowReader.ImplicitDB(), rowReader.ImplicitTable())) {
+		return nil, ErrLimitedGroupBy
 	}
 
 	return &groupedRowReader{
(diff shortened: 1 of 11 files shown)


embedded/sql/stmt.go
```diff
@@ -1119,6 +1119,10 @@ func (stmt *SelectStmt) compileUsing(e *Engine, implicitDB *Database, params map
 		return nil, ErrHavingClauseRequiresGroupClause
 	}
 
+	if len(stmt.groupBy) > 1 {
```
byo, line 1122: If this check was not here, such situation would still be caught by comparing groupBy columns with orderBy columns (https://github.com/codenotary/immudb/pull/890/files#diff-1096943eee67a63ccd645a914b0b5c90a70bfb49378dbe325ead03929b68329aR39), right?

embedded/sql/grouped_row_reader.go
```diff
@@ -31,10 +31,15 @@ type groupedRowReader struct {
 }
 
 func (e *Engine) newGroupedRowReader(rowReader RowReader, selectors []Selector, groupBy []*ColSelector) (*groupedRowReader, error) {
-	if rowReader == nil || len(selectors) == 0 {
+	if rowReader == nil || len(selectors) == 0 || len(groupBy) > 1 {
```
byo, line 34: How about `len(groupBy) == 0` ? Is this a valid case?

byo (approved):
Added two small questions but I don't see them as blockers for this PR

embedded/sql/stmt.go
jeroiraz: yes, it'll be caught there before. But preferred to have an earlier validation

embedded/sql/grouped_row_reader.go
jeroiraz: yes, it's the edge case for queries such as `select count() from mytable`

PR #892, opened 2021-08-13
multi error enhacements

jeroiraz:

### DIFF
--- a/embedded/ahtree/ahtree.go
+++ b/embedded/ahtree/ahtree.go
@@ -779,25 +779,19 @@ func (t *AHtree) Close() error {
 
 	t.closed = true
 
-	errors := make([]error, 0)
+	merrors := multierr.NewMultiErr()
 
-	pErr := t.pLog.Close()
-	if pErr != nil {
-		errors = append(errors, pErr)
-	}
+	err := t.pLog.Close()
+	merrors.Append(err)
 
-	dErr := t.dLog.Close()
-	if dErr != nil {
-		errors = append(errors, dErr)
-	}
+	err = t.dLog.Close()
+	merrors.Append(err)
 
-	cErr := t.cLog.Close()
-	if cErr != nil {
-		errors = append(errors, cErr)
-	}
+	err = t.cLog.Close()
+	merrors.Append(err)
 
-	if len(errors) > 0 {
-		return &multierr.MultiErr{Errors: errors}
+	if merrors.HasErrors() {
+		return merrors
 	}
 
 	return nil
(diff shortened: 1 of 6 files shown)


embedded/multierr/multierr.go
```diff
@@ -15,12 +15,45 @@ limitations under the License.
...
 
-import "fmt"
+import (
+	"errors"
+	"fmt"
+)
 
 type MultiErr struct {
-	Errors []error
+	errors []error
```
byo, line 24: I'm wondering if it would be good to add custom `As` and `Is` methods (https://blog.golang.org/go1.13-errors#TOC_4.) so that native `errors.Is` and `errors.As` work here.

Currently checking if the multierror was caused by a specific low-level error (e.g. IO write failure) would require casting to multierr and using the `Includes` error.

jeroiraz: I though about it too, but wasn't fully convinced. But I think it'll be better

embedded/ahtree/ahtree.go
```diff
@@ -779,25 +779,19 @@ func (t *AHtree) Close() error {
...
 
-	errors := make([]error, 0)
+	merrors := multierr.NewMultiErr()
 
-	pErr := t.pLog.Close()
-	if pErr != nil {
-		errors = append(errors, pErr)
-	}
+	err := t.pLog.Close()
+	merrors.Append(err)
```
byo, line 785: Is it really OK to continue if there was an error with the close?

If there's an error during flush of payload or digests then maybe it would be better to discard flushing commit log too?

embedded/store/immustore_test.go
```diff
@@ -524,10 +523,7 @@ func TestImmudbStoreEdgeCases(t *testing.T) {
 	err = store.aht.Close()
 	require.NoError(t, err)
 	err = store.Close()
-	require.IsType(t, &multierr.MultiErr{}, err)
-	mErr := err.(*multierr.MultiErr)
-	require.Len(t, mErr.Errors, 1)
-	require.ErrorIs(t, mErr.Errors[0], ahtree.ErrAlreadyClosed)
+	require.ErrorIs(t, err, ahtree.ErrAlreadyClosed)
```
byo, line 526: :+1: Much cleaner now, I like that

embedded/ahtree/ahtree.go
jeroiraz: consistency does not relies on Close. The intention is to properly close as many files as possible.

Not in the scope on this PR, but if we find a different way on handling close operation, it'll be ok of course

byo (approved):

Issue #889, opened 2021-08-11
Reproducible databases (option for no server timestamp in hash)


MikeBeller:
Please add the ability to run immudb, or an immudb database, in "reproducible" mode -- where the server commit timestamp is not included in the hash.

Use case: I am feeding my database from Kafka for example.  With databases other than Immudb, if I  need to reload, or if I want to run parallel database replicas, I can just load them independently from the Kafka queue and the databases will be identical from all important perspectives.  Not the case with Immudb, as the series of hashes will not be reproducible.  The timestamps will be different.

If Immudb had a "don't put your timestamps in this database" option, then this would solve my problem.  If that idea doesn't work because Immudb needs timestamps for some reason, then if I could supply the timestamp through the API, that would probably be ok too.  ( I would just make it part of the message persisted in the Kafka queue and would feed it into the API when I reload.)

jeroiraz:
Thanks for creating this enhancement request @MikeBeller

jeroiraz:
Hi @MikeBeller, we've a PR associated to this feature request. Would you like building immudb with the associated branch?

User created databases either using immuadmin or an sdk would be reproducible by setting `exclude-commit-time` flag or `ExcludeCommitTime` attribute.

MikeBeller:
Great!  I was able to build the reproducible_stores branch, start immudb, create two databases, and put the same k/v pairs in them in the same order -- got the same hashes.  Thanks for this!  Looking forward to seeing it merged.  

![Screen Shot 2021-09-06 at 3 03 50 PM](https://user-images.githubusercontent.com/1700778/132254675-f81b9b4e-415f-4fb1-973b-2c43e37c1b57.png)

victornor:
> Hi @MikeBeller, we've a PR associated to this feature request. Would you like building immudb with the associated branch?
> 
> User created databases either using immuadmin or a sdk would be reproducible by setting `exclude-commit-time` flag or `ExcludeCommitTime` attribute.

Where is this CommitTime accessible? I'm asking here because i see no mention about history timestamps or commit times anywhere else, so I'm wondering if this is something we can access?

I want to compare linux access times to key/value history change timestamps, so that I'm able to determine if a value change was made during an ssh connection to the server.

Is this possible?

jeroiraz:
> > Hi @MikeBeller, we've a PR associated to this feature request. Would you like building immudb with the associated branch?
[...]

PR #895, opened 2021-08-15
Reproducible stores

jeroiraz:
This PR aims to provide reproducible databases by setting a known fixed commit time.

This work is still under progress and should provide a solution to #889 

Database creation/update may be done using the SDK or with immuadmin CLI tool.

e.g. creating/updating a database using immuadmin
```bash
./immuadmin database create myreproducibledb --exclude-commit-time=true
```


cmd/immuadmin/command/database.go
```diff
@@ -81,19 +81,26 @@ func (cl *commandline) database(cmd *cobra.Command) {
...
 			}
 
 			fmt.Fprintf(cmd.OutOrStdout(), "database '%s' (replica = %v) successfully created\n", args[0], isReplica)
 			return nil
 		},
-		Args: cobra.ExactArgs(1),
+		Args: cobra.MaximumNArgs(2),
 	}
 	cc.Flags().BoolP("replica", "r", false, "set database as a replica")
+	cc.Flags().Bool("exclude-commit-time", false, "useful when reproducibility is a desired feature")
```
byo, line 103: I'd be a bit more verbose here, something like:

```suggestion
	cc.Flags().Bool("exclude-commit-time", false,
	    "do not include server-side timestamps in commit checksums, useful when reproducibility is a desired feature")
```

cmd/immuadmin/command/database.go
```diff
@@ -81,19 +81,26 @@ func (cl *commandline) database(cmd *cobra.Command) {
...
+				ExcludeCommitTime: excludeCommitTime,
 			}); err != nil {
 				return err
 			}
 
 			fmt.Fprintf(cmd.OutOrStdout(), "database '%s' (replica = %v) successfully created\n", args[0], isReplica)
 			return nil
 		},
-		Args: cobra.ExactArgs(1),
+		Args: cobra.MaximumNArgs(2),
```
byo, line 100: Not sure I understand why this is needed - something changed in the number of positional arguments?

cmd/immuadmin/command/database.go
```diff
@@ -111,19 +118,26 @@ func (cl *commandline) database(cmd *cobra.Command) {
...
 			}
 
 			fmt.Fprintf(cmd.OutOrStdout(), "database '%s' (replica = %v) successfully updated\n", args[0], isReplica)
 			return nil
 		},
-		Args: cobra.ExactArgs(1),
+		Args: cobra.MaximumNArgs(2),
 	}
 	cu.Flags().BoolP("replica", "r", false, "set database as a replica")
+	cu.Flags().Bool("exclude-commit-time", false, "useful when reproducibility is a desired feature")
```
byo, line 140: Same as for line 103

cmd/immuadmin/command/database.go
```diff
@@ -111,19 +118,26 @@ func (cl *commandline) database(cmd *cobra.Command) {
 				c.PrintfColorW(cmd.OutOrStdout(), c.Yellow, "Replication is a work-in-progress feature. Not ready for production use\n")
 			}
 
+			excludeCommitTime, err := cmd.Flags().GetBool("exclude-commit-time")
```
byo, line 121: I'm wondering if there will ever be a need to update this flag - either you want to have reproducible hashes and should start with no timestamps from the beginning or you start with timestamps and don't disable it later (because you won't have reproducibility anyway due to initial commits with timestamps). It could simplify the rest of the code I believe?

pkg/server/server.go
```diff
@@ -712,6 +719,12 @@ func (s *ImmuServer) CreateDatabaseWith(ctx context.Context, req *schema.Databas
 		AsReplica(settings.Replica).
 		WithReplicationOptions(replicationOpts)
 
+	if req.ExcludeCommitTime {
```
byo, line 722: This repeats in 3 places, a good candidate to extract as a separate function

byo (approved):
LGTM with some minor comments

pkg/server/server.go
jeroiraz: yes, this will change once another ongoing PR gets merged where a separate function is already there

cmd/immuadmin/command/database.go
jeroiraz: The reason to even naming the flag `exclude-commit-time` instead of `reproducible` it's because in order to fully ensure reproducibility there may be other things in consideration. Also making it explicit on the API to dynamically change it may help in understanding that reproducibility may be obtained if server commit time was always excluded. But in order to prevent changes, there must be logic to validate it never happen. Which may be out of the initial scope

PR #896, opened 2021-08-16
sql insert optimization

jeroiraz:
This PR includes a minor fix when validating constraints at kv store level and a sql insert optimization when using auto-incremental primary keys.

### DIFF
--- a/embedded/sql/stmt.go
+++ b/embedded/sql/stmt.go
@@ -591,9 +591,11 @@ func (stmt *UpsertIntoStmt) compileUsing(e *Engine, implicitDB *Database, params
 
 		constraint := store.NoConstraint
 
-		if stmt.isInsert {
+		if stmt.isInsert && !table.pk.autoIncrement {
 			constraint = store.MustNotExist
-		} else if table.pk.autoIncrement {
+		}
+
+		if !stmt.isInsert && table.pk.autoIncrement {
 			constraint = store.MustExist
 		}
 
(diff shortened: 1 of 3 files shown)


embedded/sql/stmt.go
```diff
@@ -591,9 +591,11 @@ func (stmt *UpsertIntoStmt) compileUsing(e *Engine, implicitDB *Database, params
 
 		constraint := store.NoConstraint
 
-		if stmt.isInsert {
+		if stmt.isInsert && !table.pk.autoIncrement {
```
byo, line 594: :+1: Nice catch! I didn't realize before that Upsert is also used for standard Insert.

Out of curiosity - why not the other way around so that the Upsert is a special case of Insert?

embedded/store/immustore.go
```diff
@@ -1012,7 +1012,7 @@ func (s *ImmuStore) commit(tx *Tx, offsets []int64, ts int64, blTxID uint64) err
 			if err == ErrKeyNotFound && txe.constraint == MustExist {
 				return ErrKeyNotFound
 			}
-			if err != ErrKeyNotFound {
+			if err != nil && err != ErrKeyNotFound {
```
byo, line 1015: Minor: maybe better to move right after `s.indexer.Get` ? That way error checking will be close to the actual method and will be a bit easier to read.

byo (approved):
Nice catch :+1:.

Only two minor comments which I don't consider blockers for this PR

embedded/store/immustore.go
jeroiraz: all the lines above after Get are error checking conditions as well ...

embedded/sql/stmt.go
jeroiraz: Upserts comes easier than Insert. Insert requires constraint checking while Upsert is able to work without any checking. That's why Insert functionality comes afterwards

PR #899, opened 2021-08-26
Follower replication

jeroiraz:
This PR implements embedded replication.

It's possible to create databases as read-only replicas from other databases.


pkg/replication/replicator.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	txr.mainContext, txr.cancelFunc = context.WithCancel(context.Background())
+
+	txr.nextTx = st.TxId + 1
+
+	txr.running = true
+
+	go func() {
+		for {
```
byo, line 99: Shouldn't we also check if the context was not cancelled?

```suggestion
		for txr.mainContext.Err() == nil {
```

I smell some subtle timing issues that could cause reopening of the client connection after the `Stop` was called

byo: A corner case is if there's some connect / fetchTx error and both the timeout and context cancellation fire at the same time. Then even though context is cancelled, the second path (timeout) can be selected and will cause one more reconnect.

A very rare case thus I don't think it should block this PR.

byo (approved):
Approved with some minor comment

byo (approved):
Some tests are failing but looks like the feature is working as expected

elimisteve:
Are there any docs showing us how to use this replication feature? Thanks!

jeroiraz:
> Are there any docs showing us how to use this replication feature? Thanks!

Hi @elimisteve, we described it here: https://docs.immudb.io/master/operations/replication.html

elimisteve:
Thanks @jeroiraz!

PR #909, opened 2021-09-14
chore(Makefile): More explicit webconsole version

byo:
Makefile now contains an information what webconsole version
is the best match for the current immudb code. That version is
selected when building with `make WEBSONCOLE=default ...`.
Alternatively one can select the latest released version of
webconsole with `make WEBCONSOLE=latest ...` or use a specific
version with `make WEBCONSOLE=<version> ...`.


Hextar (approved):
LGTM now!

PR #910, opened 2021-09-14
chore(docker): Docker image build improvements

byo:
The previously used `latest` tag should only be used when doing
official releases so that running docker images with that tag
use the most up-to-date stable version.


.github/workflows/push.yml
```diff
@@ -90,15 +90,15 @@ jobs:
           with:
             name: vcn
         - run: |
-            docker build --tag "${DOCKER_IMAGE_IMMUDB}:latest" -f Dockerfile .
-            docker build --tag "${DOCKER_IMAGE_IMMUADMIN}:latest" -f Dockerfile.immuadmin .
-            docker build --tag "${DOCKER_IMAGE_IMMUCLIENT}:latest" -f Dockerfile.immuclient .
+            docker build --tag "${DOCKER_IMAGE_IMMUDB}:devel" -f Dockerfile .
```
jeroiraz, line 93: why not `dev` ?

.dockerignore
```diff
@@ -2,14 +2,16 @@
...
+/data*
+
 # Vendor
 vendor
 
 # Unuseful stuff
 *.iml
 *.swp
 .idea/
-.git/
```
jeroiraz, line 12: just wondering why `.git` can be removed from the list

Dockerfile
```diff
@@ -36,7 +39,45 @@ RUN addgroup --system --gid $IMMU_GID immu && \
...
+
+FROM scratch as scratch
+MAINTAINER CodeNotary, Inc. <info@codenotary.io>
+
+ARG IMMU_UID="3322"
+ARG IMMU_GID="3322"
+
+ENV IMMUDB_HOME="/usr/share/immudb" \
+    IMMUDB_DIR="/var/lib/immudb" \
+    IMMUDB_DBNAME="immudb" \
```
jeroiraz, line 57: hmm, I think `IMMUDB_DBNAME` is not used....

Dockerfile
```diff
@@ -36,7 +39,45 @@ RUN addgroup --system --gid $IMMU_GID immu && \
...
+    IMMUADMIN_TOKENFILE="/var/lib/immudb/admin_token" \
+    USER=immu \
+    HOME="${IMMUDB_HOME}"
+
+COPY --from=build /src/immudb /usr/sbin/immudb
+COPY --from=build /src/immuadmin /usr/local/bin/immuadmin
+COPY --from=build --chown="$IMMU_UID:$IMMU_GID" /empty "$IMMUDB_HOME"
+COPY --from=build --chown="$IMMU_UID:$IMMU_GID" /empty "$IMMUDB_DIR"
+
+EXPOSE 3322
```
jeroiraz, line 77: aren't already exposed above?

SimoneLazzaris (dismissed):

.dockerignore
byo: We un-ignore the .git folder so that when the COPY is performed into the initial docker image, the `.git` folder is copied too.

Without this, `make` is not able to extract current git hash and thus the version info is incomplete.

Stefano just added displaying this hash in webconsole next to the version info (so that we know exactly what commit was compiled) to ease troubleshooting.

Alternative would be to pass the commit through docker build args but that would be much larger change.

.github/workflows/push.yml
```diff
@@ -90,24 +93,48 @@ jobs:
...
-            docker push "${DOCKER_IMAGE_IMMUDB}:latest"
-            docker push "${DOCKER_IMAGE_IMMUADMIN}:latest"
-            docker push "${DOCKER_IMAGE_IMMUCLIENT}:latest"
-           
+            docker push "${DOCKER_IMAGE_IMMUDB}:dev"
+            docker push "${DOCKER_IMAGE_IMMUDB}:dev-${DEBIAN_VERSION}"
+            docker push "${DOCKER_IMAGE_IMMUADMIN}:dev"
+            docker push "${DOCKER_IMAGE_IMMUCLIENT}:dev"
+
+            if [[ ! -z "$VERSION_TAG" ]]; then
```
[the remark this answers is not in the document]
byo: Bash will be needed in this script anyway, at the beginning we're extracting version using bash regexp.

I added an explicit bash requirement for this job.

Dockerfile
byo: Didn't find this option in immudb. Removed

.dockerignore
jeroiraz: probably after this PR:

it'd be better to build docker images by fetching already notarized immudb binaries

and we could expose the build info from immudb so the webconsole or any other app could consume it

byo: build info is (partially, only the version and git hash appended to it) available in the `HealthCheck` rpc call, webconsole will be calling this API to get the data.

SimoneLazzaris (approved):

PR #911, opened 2021-09-14
sql signed integers

jeroiraz:
This PR introduces support for signed integers. With this changes `INTEGER` type is interpreted as signed 64bits integer.

It's also possible to specify expressions when inserting rows:
```sql
insert into mytable(amount) values (1), (-3), (4+2-4);
insert into mytable2(active) values (1<2);
```

Or use negative values as part of a query condition:
```sql
select * from mytable where amount >= -3+3+1 order by amount
```


embedded/sql/engine.go
```diff
@@ -555,7 +555,7 @@ func (e *Engine) loadTables(db *Database, catalogSnap, dataSnap *store.Snapshot)
 				return ErrCorruptedData
 			}
 
-			table.maxPK = binary.BigEndian.Uint64(encMaxPK)
+			table.maxPK = int64(binary.BigEndian.Uint64(encMaxPK) + math.MaxInt64 + 1)
```
byo, line 558: I would go for just plain cast between signed and unsigned value here:

```suggestion
			table.maxPK = int64(binary.BigEndian.Uint64(encMaxPK))
```

It would be backwards-compatible and IIRC such casts are well-defined in golang, and that's how it's done in go's stdlib: https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/encoding/binary/binary.go;l=329

embedded/sql/engine.go
```diff
@@ -937,15 +937,16 @@ func EncodeValue(val interface{}, colType SQLValueType, maxLen int) ([]byte, err
...
 			if !ok {
 				return nil, ErrInvalidValue
 			}
 
+			// map to unsigned integer space
 			// len(v) + v
 			var encv [EncLenLen + 8]byte
 			binary.BigEndian.PutUint32(encv[:], uint32(8))
-			binary.BigEndian.PutUint64(encv[EncLenLen:], intVal)
+			binary.BigEndian.PutUint64(encv[EncLenLen:], uint64(intVal+math.MaxInt64+1))
```
byo, line 949: Same here, mapping from signed to unsigned space can be done by just a simple type conversion, no need to shift values:

```suggestion
			binary.BigEndian.PutUint64(encv[EncLenLen:], uint64(intVal))
```

embedded/sql/engine.go
```diff
@@ -1030,14 +1031,15 @@ func EncodeAsKey(val interface{}, colType SQLValueType, maxLen int) ([]byte, err
...
+			intVal, ok := val.(int64)
 			if !ok {
 				return nil, ErrInvalidValue
 			}
 
 			// v
+			// map to unsigned integer space
 			var encv [8]byte
-			binary.BigEndian.PutUint64(encv[:], intVal)
+			binary.BigEndian.PutUint64(encv[:], uint64(intVal+math.MaxInt64+1))
```
byo, line 1042:
```suggestion
			binary.BigEndian.PutUint64(encv[:], uint64(intVal))
```

embedded/sql/engine.go
```diff
@@ -1109,16 +1111,19 @@ func DecodeValue(b []byte, colType SQLValueType) (TypedValue, int, error) {
...
 		{
-			if vlen > 8 {
+			if vlen != 8 {
 				return nil, 0, ErrCorruptedData
 			}
 
-			buff := [8]byte{0}
+			buff := [8]byte{}
 			copy(buff[8-vlen:], b[voff:voff+vlen])
 			v := binary.BigEndian.Uint64(buff[:])
```
byo, line 1120: Since 8 bytes are now required, the extra buffer is no longer needed:

```suggestion
			v := binary.BigEndian.Uint64(b[voff:])
```

embedded/sql/engine.go
```diff
@@ -1109,16 +1111,19 @@ func DecodeValue(b []byte, colType SQLValueType) (TypedValue, int, error) {
...
 			voff += vlen
 
-			return &Number{val: v}, voff, nil
+			// map to signed integer space
+			v += math.MaxInt64 + 1
+
+			return &Number{val: int64(v)}, voff, nil
```
byo, line 1126: I would just use a simple cast:

```suggestion
			return &Number{val: int64(v)}, voff, nil
```

byo (changes requested):

embedded/sql/engine.go
jeroiraz: the shift when mapping from signed to unsigned space is needed in order to provide lexicographic order when encoding (big-endian)  e.g. -3 -1 0 1 3
Otherwise -1 would become the greatest one.

embedded/sql/engine.go
jeroiraz: replied above

embedded/sql/engine.go
jeroiraz: replied above

embedded/sql/engine.go
jeroiraz: replied above

embedded/sql/engine.go
byo: I agree on that for the index - we'd have to shift the range for correct sorting.
But I'm not so sure about the payload itself. Receiving proof for a row would not be so obvious anymore - it's not a trivial big-endian integer anymore.

jeroiraz: The proofs and indexing take place at the kv-level. At that layer we deal just with byte-arrays, so we need to serialize the data in the same way for storing, indexing and to build proofs

SimoneLazzaris (approved):
Tested, LGTM

SimoneLazzaris (approved):

PR #914, opened 2021-09-17
Test/std sql code coverage

mmeloni:
Connections are now released when database is closed

### DIFF
--- a/pkg/stdlib/sql.go
+++ b/pkg/stdlib/sql.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"database/sql"
 	"database/sql/driver"
+	"fmt"
 	"github.com/codenotary/immudb/pkg/api/schema"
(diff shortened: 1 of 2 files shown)


pkg/stdlib/sql.go
```diff
@@ -208,18 +224,6 @@ func (c *Conn) QueryContext(ctx context.Context, query string, argsV []driver.Na
 	return &Rows{conn: c, rows: queryResult.Rows}, nil
 }
 
-func (c *Conn) Ping(ctx context.Context) error {
```
byo, line 211: Why this was removed?

byo (approved):
Just one question regarding PIng, overall looks good but I didn't yet try to use this.

pkg/stdlib/sql.go
mmeloni: Implementing pinger interface Is not needed in the initial implementation.

PR #915, opened 2021-09-17
feat(Dockerimage): Use upx and strip to make immudb image smaller

byo:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,16 @@
 FROM golang:1.17 as build
+
+RUN apt-get update -qq && \
+    apt-get install -qq upx
+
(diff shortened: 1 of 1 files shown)


SimoneLazzaris:
Made some tests and I have some thoughts:

Pro:
- smaller image
- works like a charm

Con:
- it was already quite small
- compressing upx-processed images doesn't do any good: the resulting image is not smaller than the original, while compressing non-upx image gives even smaller images (basically, because you don't have to deal with the decompression code).

byo:
docker image layers are already transferred in a compressed form thus we don't win anything by compressing binaries inside.

PR #917, opened 2021-09-18
fix key validation and prevent external runtime changes in indexing data structure

jeroiraz:


embedded/tbtree/reader.go
```diff
@@ -262,3 +262,14 @@ func (r *Reader) Close() error {
 
 	return nil
 }
+
+func cp(s []byte) []byte {
```
byo, line 266: If a coy is needed , do we modify keys in-place then?

byo (approved):

embedded/tbtree/reader.go
jeroiraz: returning a copy is to prevent any external modification (conservative but safe approach).
e.g. sql layer was mapping keys and causing accidental changes, while it'd be possible to make a copy before the mapping, returning a copy will ensure this situation won't take place.

PR #926, opened 2021-09-21
chore: decoupled token service

mmeloni:
Token service interface now is implemented by:
**[inmemory](https://github.com/codenotary/immudb/blob/chore/decoupled-token-service/pkg/client/tokenservice/inmemory.go)** and **[file](https://github.com/codenotary/immudb/blob/chore/decoupled-token-service/pkg/client/tokenservice/file.go)** struct

WithHomedirService and WithTokenFilename was now removed from interface and are coupled only with file TokenService implementation.

[Login method](https://github.com/codenotary/immudb/blob/chore/decoupled-token-service/pkg/client/client.go#L467), useDatabase immuclient method now use token service to store token response before exit.
[Token interceptor](https://github.com/codenotary/immudb/blob/chore/decoupled-token-service/pkg/client/token_interceptor.go) if doesn't find a token in the context check if the token is present in the token service

Several tests are modified mainly for token service package movement, so no particular checks are needed on them IMO.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -482,10 +482,10 @@ func (c *immuClient) Login(ctx context.Context, user []byte, pass []byte) (*sche
 
 	err = c.Tkns.SetToken("defaultdb", result.Token)
 	if err != nil {
-		return nil, err
+		return nil, errors.FromError(err)
 	}
 
-	return result, errors.FromError(err)
+	return result, nil
 }
 
 // Logout ...
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -23,6 +23,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"github.com/codenotary/immudb/pkg/client/tokenservice"
 	"io"
 	"io/ioutil"
 	"os"
@@ -75,7 +76,7 @@ type ImmuClient interface {
 	WithStateService(rs state.StateService) *immuClient
 	WithClientConn(clientConn *grpc.ClientConn) *immuClient
 	WithServiceClient(serviceClient schema.ImmuServiceClient) *immuClient
-	WithTokenService(tokenService TokenService) *immuClient
+	WithTokenService(tokenService tokenservice.TokenService) *immuClient
 	WithServerSigningPubKey(serverSigningPubKey *ecdsa.PublicKey) *immuClient
 	WithStreamServiceFactory(ssf stream.ServiceFactory) *immuClient
 
@@ -166,7 +167,7 @@ type immuClient struct {
 	clientConn           *grpc.ClientConn
 	ServiceClient        schema.ImmuServiceClient
 	StateService         state.StateService
-	Tkns                 TokenService
+	Tkns                 tokenservice.TokenService
 	serverSigningPubKey  *ecdsa.PublicKey
 	StreamServiceFactory stream.ServiceFactory
 	sync.RWMutex
@@ -190,7 +191,7 @@ func NewImmuClient(options *Options) (c ImmuClient, err error) {
 	c.WithOptions(options)
 	l := logger.NewSimpleLogger("immuclient", os.Stderr)
 	c.WithLogger(l)
-	c.WithTokenService(options.Tkns.WithTokenFileName(options.TokenFileName))
+	c.WithTokenService(options.Tkns)
 
 	if options.ServerSigningPubKey != "" {
 		pk, err := signer.ParsePublicKeyFile(options.ServerSigningPubKey)
@@ -316,6 +317,8 @@ func (c *immuClient) SetupDialOptions(options *Options) *[]grpc.DialOption {
 			opts = append(opts, grpc.WithStreamInterceptor(auth.ClientStreamInterceptor(token)))
 		}
 	}
+	uic = append(uic, c.TokenInterceptor)
+
 	opts = append(opts, grpc.WithUnaryInterceptor(grpc_middleware.ChainUnaryClient(uic...)), grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(options.MaxRecvMsgSize)))
 
 	return &opts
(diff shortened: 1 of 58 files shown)


pkg/client/token_interceptor.go
```diff
@@ -0,0 +1,39 @@
...
+	"google.golang.org/grpc"
+)
+
+// TokenInterceptor inject token from tokenservice if present and if provided context contain no one
+func (c *immuClient) TokenInterceptor(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+	var ctxToken string
+	if md, ok := metadata.FromOutgoingContext(ctx); ok && len(md.Get("authorization")) > 0 {
+		ctxToken = md.Get("authorization")[0]
+	}
+	if token, err := c.Tkns.GetToken(); ctxToken == "" && err == nil {
```
byo, line 32: Even if `ctxToken` is alredy present, the `c.Tkns.GetToken()` will still be invoked here. This may cause some unexpected side effects.

mmeloni: But there is ctxToken == "" guard. This means that local token will not be injected into context if context token is provided. Isnt?

byo: That's correct, the issue is not that the invalid token is used but rather that the `c.Tkns.GetToken()` is always called, even if `ctxToken` is provided. Let's say this function is a complex logic querying some external secret management service, etc. There's no point in calling it if the token is already in the context.

pkg/client/client.go
```diff
@@ -472,9 +475,16 @@ func (c *immuClient) Login(ctx context.Context, user []byte, pass []byte) (*sche
...
 	})
-
+	if err != nil {
+		return nil, errors.FromError(err)
+	}
 	c.Logger.Debugf("login finished in %s", time.Since(start))
 
+	err = c.Tkns.SetToken("defaultdb", result.Token)
+	if err != nil {
+		return nil, err
```
byo, line 485: Shouldn't be:

```suggestion
		return nil, errors.FromError(err)
```

?

pkg/client/client.go
```diff
@@ -472,9 +475,16 @@ func (c *immuClient) Login(ctx context.Context, user []byte, pass []byte) (*sche
...
+	}
+
 	return result, errors.FromError(err)
```
byo, line 488: Here the `err == nil`, so we can simplify:

```suggestion
	return result, nil
```

pkg/client/token_interceptor.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+	"google.golang.org/grpc/metadata"
+
+	"google.golang.org/grpc"
+)
+
+// TokenInterceptor inject token from tokenservice if present and if provided context contain no one
+func (c *immuClient) TokenInterceptor(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+	if md, ok := metadata.FromOutgoingContext(ctx); !ok || len(md.Get("authorization")) == 0 {
+		if token, err := c.Tkns.GetToken(); err == nil && token != "" {
```
byo, line 29: The err is silently ignored here, I'm not sure if this is a good approach. Let's say that the file with token is not readable. In such situation the user will see authentication failure instead of IO error.

pkg/client/tokenservice/file.go
```diff
@@ -14,35 +14,29 @@ See the License for the specific language governing permissions and
...
-	WithTokenFileName(tfn string) TokenService
-	IsTokenPresent() (bool, error)
-	DeleteToken() error
-	GetToken() (string, error)
-	GetDatabase() (string, error)
-}
-
-type tokenService struct {
+type file struct {
 	tokenFileName string
```
byo, line 27: file token service is not thread-safe where the in-memory service is thread-safe. Shouldn't we always guarantee thread-safety?

pkg/client/tokenservice/inmemory_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestNewInmemoryTokenService(t *testing.T) {
+	ts := NewInmemoryTokenService()
+	err := ts.SetToken("db", "tk")
+	require.NoError(t, err)
+	_, err = ts.IsTokenPresent()
+	require.NoError(t, err)
```
byo, line 29: Please also check returned values, e.g.:

```suggestion
	present, err = ts.IsTokenPresent()
	require.NoError(t, err)
	require.True(t, present)
```

pkg/integration/client_test.go
```diff
@@ -525,6 +527,58 @@ func TestDatabasesSwitching(t *testing.T) {
...
+	bs := servertest.NewBufconnServer(options)
+
+	defer os.RemoveAll(options.Dir)
+	defer os.Remove(".state-")
+
+	bs.Start()
+	defer bs.Stop()
+
+	client, err := ic.NewImmuClient(ic.DefaultOptions().WithDialOptions(&[]grpc.DialOption{grpc.WithContextDialer(bs.Dialer), grpc.WithInsecure()}).WithTokenService(tokenservice.NewInmemoryTokenService()))
+	if err != nil {
```
byo, line 541: Minor: Why we're not using `require.NoError(t, err)` here? Is there some specific reason?

pkg/integration/client_test.go
```diff
@@ -525,6 +527,58 @@ func TestDatabasesSwitching(t *testing.T) {
...
+	options := server.DefaultOptions().WithAuth(true)
+	bs := servertest.NewBufconnServer(options)
+
+	defer os.RemoveAll(options.Dir)
+	defer os.Remove(".state-")
+
+	bs.Start()
+	defer bs.Stop()
+
+	client, err := ic.NewImmuClient(ic.DefaultOptions().WithDialOptions(&[]grpc.DialOption{grpc.WithContextDialer(bs.Dialer), grpc.WithInsecure()}).WithTokenService(tokenservice.NewInmemoryTokenService()))
```
byo, line 540: Minor: Pretty hard to read on github, can you split this into multiple lines?

byo (dismissed):

byo (dismissed):

jeroiraz (approved):

byo (approved):

PR #927, opened 2021-09-21
feat(embedded/sql): distinct row reader

jeroiraz:
This PR introduces support for `distinct` clause in sql queries
e.g.
```sql
SELECT DISTINCT error_type FROM mylogtable
```


embedded/sql/distinct_row_reader.go
```diff
@@ -0,0 +1,93 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		_, ok := dr.readRows[digest]
+		if ok {
+			continue
+		}
+
+		dr.readRows[digest] = struct{}{}
```
byo, line 85: This has no upper memory limit and can grow indefinitely, maybe we could add some artificial limit, something like 1M entries (in my quick test it took ~120MB of memory for such map) and fail with too many rows error?

jeroiraz: yes, but probably better as an engine option so it can be customized.

embedded/sql/row_reader.go
```diff
@@ -67,6 +68,31 @@ func (row *Row) Compatible(aRow *Row, selectors []*ColSelector, db, table string
...
 }
 
+func (row *Row) digest() (d [sha256.Size]byte, err error) {
+	h := sha256.New()
+
+	for _, v := range row.Values {
+		_, isNull := v.(*NullValue)
+		if isNull {
+			continue
+		}
```
byo, line 78: That way we skip null values in the tuples but it comes with some incorrect results, e.g.
`Row(1,null).digest == Row(null,1).digest`.

Can be easily confirmed with the following steps:
```bash
./immuclient exec "CREATE TABLE t1(id INTEGER AUTO_INCREMENT, a INTEGER, b INTEGER, PRIMARY KEY id)"
./immuclient exec "INSERT INTO t1(a,b) VALUES(1,null),(null,1)"
./immuclient query "SELECT DISTINCT a,b FROM t1"
# +------------------+------------------+
# | (DEFAULTDB T1 A) | (DEFAULTDB T1 B) |
# +------------------+------------------+
# |                1 | NULL             |
# +------------------+------------------+
```

For now I'd simply add some null indicator, e.g.:
```suggestion
		_, isNull := v.(*NullValue)
		if isNull {
			h.Write([]byte{0})
			continue
		}
		h.Write([]byte{1})
```

embedded/sql/row_reader.go
```diff
@@ -67,6 +68,31 @@ func (row *Row) Compatible(aRow *Row, selectors []*ColSelector, db, table string
 	return true, nil
 }
 
+func (row *Row) digest() (d [sha256.Size]byte, err error) {
+	h := sha256.New()
+
+	for _, v := range row.Values {
```
byo, line 74: The iteration order is non-deterministic for hash maps, this may have unexpected results.

I checked and even reserving map size changes the order:
https://play.golang.org/p/LZMrO6a5n-W

byo (changes requested):

embedded/sql/row_reader.go
jeroiraz: actually that may not be enough to prevent overlaps, each encoded value may need to include the offset as prefix

byo (approved):
:+1:

PR #932, opened 2021-09-27
UPSERT with secondary indexes

jeroiraz:
This PR aims to remove the current limitation of `UPSERT` ("upsert is only supported in tables without secondary indexes")

This implementation leverage new metadata record in the underlying store. Meaning this change preserves backward-compatibility.


byo:
Just a random thought:

how about maing a distinction between UPSERT and UPDATE?

UPSERT would only work on tables without secondary indexes and would require all columns - that would be the way you insert data if you want to have the maximum speed (guaranteed by the fact that we don't have to read back previous row and index states).

UPDATE on the other hand would be much more complex - would allow partial updates and updates if secondary indexes are present.

jeroiraz:
> Just a random thought:
> 
> how about maing a distinction between UPSERT and UPDATE?
> 
> UPSERT would only work on tables without secondary indexes and would require all columns - that would be the way you insert data if you want to have the maximum speed (guaranteed by the fact that we don't have to read back previous row and index states).
> 
> UPDATE on the other hand would be much more complex - would allow partial updates and updates if secondary indexes are present.

I think standard `UPDATE` may be implemented (partial updates, conditional row filtering). 

`UPSERT` won't take longer if there are no secondary indexes. Thus, if faster insertion/update is desired in a particular table, it's possible to do so just be having a primary key. So the optimization would still be possible if needed.

embedded/sql/stmt.go
```diff
@@ -695,6 +739,114 @@ func (stmt *UpsertIntoStmt) compileUsing(e *Engine, implicitDB *Database, params
...
+		// some rows might not indexed by every index
+		indexed := true
+
+		// existent index entry is deleted only if it differs from existent one
+		sameIndexKey := true
+
+		for i, col := range index.cols {
+			currVal, notNull := currValuesByColID[col.id]
+			if !notNull {
+				indexed = false
```
byo, line 811: Is this even possible? Currently we require all indexed columns to have non-null values.

embedded/tbtree/snapshot.go
```diff
@@ -35,6 +35,7 @@ type Snapshot struct {
 	t           *TBtree
 	id          uint64
 	root        node
+	unsafe      bool
```
byo, line 38: As discussed offline - this name is a bit misleading

pkg/client/sql.go
```diff
@@ -276,7 +276,7 @@ func verifyRowAgainst(row *schema.Row, decodedRow map[uint32]*schema.SQLValue, c
 }
 
 func decodeRow(encodedRow []byte, colTypes map[uint32]sql.SQLValueType) (map[uint32]*schema.SQLValue, error) {
-	off := 0
+	off := 1 // ignore DELETED flag during decoding
```
byo, line 279: Hmm, this introduces compatibility breakage in the client library. It will come with surprising results if the client is using 1.1 and the server is 1.2. If that's the case then we should definitely deprecate the old version so that people are warned about this.

byo:
Currently the deletion flag is realized on the SQL layer. I'm wondering if the kv layer won't need something similar too, if so then the SQL implementation could be based on the KV deletion rather than on its own.

Pros for deletion on the KV level:
* Can delete KV entries too (some people may need this)
* Can be used in the future to optimize count-based queries since btree could contain the information about which entries were removed and which were not
* Can be used to range scans in btree that would automatically (and quickly) skip deleted items.

Cons:
* KV layer becomes more complex
* Most likely it will be a breaking on-disk format on the KV level

jeroiraz:
> Currently the deletion flag is realized on the SQL layer. I'm wondering if the kv layer won't need something similar too, if so then the SQL implementation could be based on the KV deletion rather than on its own.
> 
> Pros for deletion on the KV level:
> 
> * Can delete KV entries too (some people may need this)
> * Can be used in the future to optimize count-based queries since btree could contain the information about which entries were removed and which were not
> * Can be used to range scans in btree that would automatically (and quickly) skip deleted items.
> 
> Cons:
> 
> * KV layer becomes more complex
> * Most likely it will be a breaking on-disk format on the KV level

Yes, initial path was that. But breaking change at KV level, not physical deletion on the btree so to support time travel queries, made me change the approach and implement it at SQL layer instead. But it may be something to be further discussed

embedded/sql/engine.go
```diff
@@ -1378,6 +1377,26 @@ func (e *Engine) ExecPreparedStmts(stmts []SQLStmt, params map[string]interface{
 		LastInsertedPKs: make(map[string]int64),
 	}
 
+	lastTxID, _ := e.dataStore.Alh()
+	err = e.dataStore.WaitForIndexingUpto(lastTxID, nil)
```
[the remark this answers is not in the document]
jeroiraz: delayed up to the specific moment it's actually needed i.e. when fetching current row state.

mmeloni (approved):

PR #933, opened 2021-09-28
Feat/types conv

mmeloni:
This increase supported types when converting to sql values.
Converting back to the original types is up to the application

### DIFF
--- a/pkg/api/schema/sql.go
+++ b/pkg/api/schema/sql.go
@@ -44,9 +44,24 @@ func asSQLValue(v interface{}) (*SQLValue, error) {
 	if v == nil {
 		return &SQLValue{Value: &SQLValue_Null{}}, nil
 	}
-
 	switch tv := v.(type) {
 	case uint:
+		{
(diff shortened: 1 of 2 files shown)


pkg/api/schema/sql.go
```diff
@@ -41,27 +42,50 @@ func EncodeParams(params map[string]interface{}) ([]*NamedParam, error) {
 }
 
 func asSQLValue(v interface{}) (*SQLValue, error) {
-	if v == nil {
+	if v == nil || (reflect.ValueOf(v).Kind() == reflect.Ptr && reflect.ValueOf(v).IsNil()) {
```
jeroiraz, line 45: just wondering if handling the nil pointer case is needed..., if not nil, it may return into an error, isn't it?

mmeloni: IMO a nil pointer can be considered nil...but the idea came from a third party application that Is sending nil pointer to SQL package ...

PR #942, opened 2021-10-01
auto-conversion from varchar to integer and boolean types

jeroiraz:
This PR provides automatic type conversion based on column specified type and provided varchar value (for integer and boolean types).

e.g. 
```sql
CREATE TABLE mytable(id INTEGER, name VARCHAR, active BOOLEAN, PRIMARY KEY id);

INSERT INTO mytable(id, name, active) VALUES ('1', 'name1', 'true');
```

### DIFF
--- a/embedded/sql/engine.go
+++ b/embedded/sql/engine.go
@@ -1098,40 +1098,38 @@ func EncodeAsKey(val interface{}, colType SQLValueType, maxLen int) ([]byte, err
 
 func intFrom(val interface{}) (int64, error) {
 	intVal, ok := val.(int64)
-	if !ok {
-		str, ok := val.(string)
-		if !ok {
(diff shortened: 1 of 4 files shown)


byo:
I'm not sure if implicit type conversions are a good idea. IMO it would be better to support explicit cast syntax.

Explicit type conversions caused a lot of issues for MySQL, examples:
* https://developpaper.com/analysis-of-mysql-implicit-conversion/
* https://adamj.eu/tech/2020/03/06/sql-implicit-type-conversion/

zaza81:
@byo  do you think this is not to merge?

jeroiraz:
implicit conversions are already in place. Given the source code changed over time, implementing conversion from varchar and boolean may be directly done over the newest code base

PR #954, opened 2021-10-06
Fix: env variable for immudb password must not be set 

SimoneLazzaris:
In order to random generate a password, the environmental variable `IMMUDB_ADMIN_PASSWORD` has to be unset, so I removed the default value in the (specific) Dockerfile


byo (approved):

SimoneLazzaris:
Fix: env variable for immudb password must not be set for random generation to occour

PR #966, opened 2021-10-08
sql: standardized datasource aliasing

jeroiraz:
This PR introduces grammar changes so to employ standard data source aliasing.

e.g.
```sql
SELECT * FROM mytable AS mt ORDER BY id

-- optional `AS`
SELECT * FROM mytable mt ORDER BY id

-- aliasing when data source is a subquery
SELECT * FROM (SELECT id, title FROM mytable) mt WHERE mt.id > 1
```


byo:
Must be merged after #965

PR #967, opened 2021-10-09
sql: create index if not exists

jeroiraz:
This PR extends index creation statement with conditional clause.

e.g.
```sql
CREATE INDEX IF NOT EXISTS ON mytable(title)

CREATE UNIQUE INDEX IF NOT EXISTS ON mytable(title)
```


byo:
Must be merged after #966

PR #969, opened 2021-10-16
Force default password change (or logout) for immuadmin command

### LINKED ISSUE #865: immuadmin is not forcing default password change
**What happened**

I was able to successfully login and operate using default credentials

After a successful login using default admin credentials immuadmin does request a password change but if confirmation password does not match the newly introduced one, I'd expect to be blocked but it was possible to operate.

Login with default credentials in inmuadmin
Enter mismatching passwords when prompted
User keeps logged in and can operate still with default password

jespino:
This ensure that the client, at least tries, to logout if your password change
fail. This only protects the CLI user interface, but still you can use the
default admin account to do all the operations through the underneath API.

Fixes #865

### DIFF
--- a/cmd/immuadmin/command/login.go
+++ b/cmd/immuadmin/command/login.go
@@ -19,6 +19,7 @@ package immuadmin
 import (
 	"context"
 	"fmt"
+	"os"
 
 	c "github.com/codenotary/immudb/cmd/helper"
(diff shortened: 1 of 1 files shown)


jespino:
@jeroiraz Cleaning up my pending PR list, feel free to reopen this PR if you think can get in, and let me know if it needs any changes.

PR #970, opened 2021-10-16
Keep order consistent on SELECT * FROM x INNER JOIN y

### LINKED ISSUE #839: Return consistent sorted column after a sql query if the * symbol has been used
A logic to retrieve SQL columns in a consistent way (always sorted in the same way) after a sql query that contains the * wildcard symbol, considering that now the order seems to be random.

eg: SELECT * FROM <table>

**Why is this needed**
That behavior directly affects the embedded web-console UI/UX.

**Additional context**

jespino:
This keeps the order of the columns consistent in the inner joins.

It can be done in more performant way if we calculate the number of columns
ahead of time somehow, and create the list with the right ammount of memory at
the beginning. If that is a requirement before we continue, I can take a look
if I can do it.

Fixes #839

### DIFF
--- a/embedded/sql/joint_row_reader_test.go
+++ b/embedded/sql/joint_row_reader_test.go
@@ -46,7 +46,7 @@ func TestJointRowReader(t *testing.T) {
 	db, err := engine.catalog.newDatabase(1, "db1")
 	require.NoError(t, err)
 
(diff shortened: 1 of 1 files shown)


jespino:
Sure, I'll do it tomorrow 🙂

PR #989, opened 2021-11-08
Use the column name as descriptor instead of the selector

### LINKED ISSUE #988: Return only column name to client
When querying ImmuDB, it return the full column path ( "_([database].[table].[column])_" ) instead of the column name ( "_[column]_" ). It would be useful, when returning response to transform column path, used for internal processing just the column name to allow end user to process the data correctly.

*Ex: query response in pgsql cli*
image

**Why is this needed**
When using some packages, in order to parse raw row data into clear objects, the driver need to get column names.
For example, when scanning a response using `database/sql` with `sqlx` in go. The package will look for the column name. But it wont be able to find anything because it can't parse the full column path. So sqlx returns an error saying it can't find the column name *"missing destination name (dbinstance.transactions.transaction_id) in \*model.Transaction"*

**Additional context**
We started discussing this issue on discord. So if you want te read beginning of that conversation, here is the link:
https://discord.com/channels/831257098368319569/831258571664130068/905107567313842246

0ctanium:
closes #988

### DIFF
--- a/pkg/pgsql/server/query_machine.go
+++ b/pkg/pgsql/server/query_machine.go
@@ -336,7 +336,7 @@ func (s *session) inferParamAndResultCols(statement string) ([]*schema.Column, [
 			return nil, nil, err
 		}
 		for _, c := range cols {
-			resCols = append(resCols, &schema.Column{Name: c.Selector(), Type: c.Type})
+			resCols = append(resCols, &schema.Column{Name: c.Column, Type: c.Type})
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/pgsql/server/query_machine.go
```diff
@@ -336,7 +336,7 @@ func (s *session) inferParamAndResultCols(statement string) ([]*schema.Column, [
 			return nil, nil, err
 		}
 		for _, c := range cols {
-			resCols = append(resCols, &schema.Column{Name: c.Selector(), Type: c.Type})
+			resCols = append(resCols, &schema.Column{Name: c.Column, Type: c.Type})
```
0ctanium, line 339: I know the database/sql descriptor is the descriptor that is exposed to the client.
But for this one I don't really know what its usage is

jeroiraz: if just the column name is used then queries like "select t1.id, t2.id ..." could lead to wrong interpretations, that's why full descriptor "db.table.col" is used instead... but the final layer may provide a way to just fetch the column name, or some way of dealing with such conflicts...

PR #992, opened 2021-11-10
immudb sessions

mmeloni:
STATUS on immudb

Transactions handling will require a connection status on immudb, in this way clients will be able to provide read and write operations in several calls but execute all of them in an atomic single execution or rollback.

IMMUDB sessions

On top of GRPC connections immudb will expose SESSIONS.

Lifecycle

SESSION is created at openSession action after validating credentials and is keept in memory until a closeSession occurs or maxSessionAge or MaxSessionIdle time is reached. At such time SESSION is closed.
When a SESSION is closed it’s possible to register,during immudb initialization, a function to be executed ( needed by transactions handling ).

To properly handling session life cycle immudb will expect that every client will send an heartbeat to prove his activity.
If a client will be not able to prove his activity ( for any kind of reasons) sending is heartbeat session will be dropped by server (after maxSessionIdle duration).

SESSION lifecycle parameters:
The following parameters will be exposed by flags or environment variables.

MaxSessionIdle 
is a duration for the amount of time after which an idle session would be closed by the server
MaxSessionAge
is a duration for the maximum amount of time a session may exist before it will be closed by the server
Timeout
the server waits for a duration of Timeout and if no activity is seen even after that the session is closed
SessionClosedCallbacks
It’s possible to register callbacks to be executed when a session is closed.

SESSION db properties

SessionID
User
SelectedDB
CreationTime
LastActivityTime
OngoingTransaction

Client side 
When a client requests a new session it will be its responsibility to send his heartbeat.
A new service will be injected into the client object in order to launch a dedicated goroutine for this purpose.

OpenSession method will returns:
the server UUID (needed to properly handling local status on clients)
the SESSION ID (ex token)

Secondaries tasks:
SDK and SERVER Login method deprecation -> OpenSession
SDK and SERVER  Logout method deprecation -> CloseSession
Sdk OpenSession will wrap grpc.connect and OpenSession call
Token renaming to SessionID
OpenSession parameters will be credentials + databasename. (in this way it will be possible to avoid 2 calls to select a database)


go.mod
```diff
@@ -50,3 +50,5 @@ require (
 )
 
 replace github.com/takama/daemon v0.12.0 => github.com/codenotary/daemon v0.0.0-20200507161650-3d4bcb5230f4
+
+replace github.com/spf13/afero => github.com/spf13/afero v1.5.1
```
byo, line 54: Why afero had to be bound here?

pkg/client/client.go
```diff
@@ -58,24 +57,36 @@ import (
 type ImmuClient interface {
 	Disconnect() error
 	IsConnected() bool
+	// Deprecated: grpc retry mechanism can be implemented with WithConnectParams dialOption
 	WaitForHealthCheck(ctx context.Context) (err error)
+	// Deprecated: grpc protocol is already handling
 	HealthCheck(ctx context.Context) error
```
byo, line 63: This could still be a valid operation - e.g. before some more complex operation I could want to ensure that the connection to the db is valid.

It wold also be useful for health checks (e.g. return unhealthy status of a whole service if it can't talk to the DB)

pkg/client/heartbeater/heartbeater.go
```diff
@@ -0,0 +1,72 @@
...
+	done          chan bool
+	t             *time.Ticker
+}
+
+func NewHeartBeater(sessionID string, sc schema.ImmuServiceClient) *heartBeater {
+	return &heartBeater{
+		sessionID:     sessionID,
+		logger:        logger.NewSimpleLogger("immuclient", stdos.Stdout),
+		serviceClient: sc,
+		done:          make(chan bool),
```
byo, line 43: I would replace this with a `context.Context` object so that it can easily be cancelled and also used inside the keepAliveRequest (another commend on that below).

pkg/client/heartbeater/heartbeater.go
```diff
@@ -0,0 +1,72 @@
...
+					return err
+				}
+			}
+		}
+	}()
+	return err
+}
+
+func (hb *heartBeater) keepAliveRequest() error {
+	_, err := hb.serviceClient.KeepAlive(context.TODO(), new(empty.Empty))
```
byo, line 67: This should use some kind of timeout to ensure we're not waiting indefinitely here.

What I'd suggest is to change theheartBeater.done channel into a context.Context object and here use a context.WithTimeout(...) - that way if the heartBeater is stopped and is still during some keepAliveRequest (possibly blocked like when the server is very slow to respond), it will be a more graceful shutdown.

pkg/client/heartbeater/heartbeater.go
```diff
@@ -0,0 +1,72 @@
...
+	go func() error {
+		for {
+			select {
+			case <-hb.done:
+				return nil
+			case t := <-hb.t.C:
+				hb.logger.Debugf("keep alive for %d at %s\n", hb.sessionID, t.String())
+				err = hb.keepAliveRequest()
+				if err != nil {
+					return err
```
byo, line 58: This is pretty dangerous - we're exiting the internal goroutine here, noone gets informed about this error but the heartbeat timer stops working.

It would be better to just log the error but continue with the next tick.

pkg/client/heartbeater/heartbeater.go
```diff
@@ -0,0 +1,72 @@
...
+				}
+			}
+		}
+	}()
+	return err
```
byo, line 63: This error is meaningless here, it will either be null (because goroutine did not yet start when this function exits) or even worse - the goroutine may start writing to it introducing race condition.

pkg/client/options.go
```diff
@@ -54,6 +55,7 @@ type Options struct {
 	LogFileName         string
 	ServerSigningPubKey string
 	StreamChunkSize     int
+	HeartBeatFrequency  time.Duration
```
byo, line 58: This field is not used to initialize the heartbeat ticker.

pkg/client/session.go
```diff
@@ -0,0 +1,79 @@
...
+		Password:     pass,
+		DatabaseName: database,
+	})
+	if err != nil {
+		return "", "", errors.FromError(err)
+	}
+
+	c.SessionID = resp.GetSessionID()
+
+	if err = heartbeater.NewHeartBeater(resp.GetSessionID(), c.ServiceClient).KeepAlive(); err != nil {
```
byo, line 54: How is heartbeater stopped? We don't even store pointer to it anywhere.

pkg/server/sessions/manager.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	sm.Running = false
+	sm.ticker.Stop()
+	sm.done <- true
+	sm.logger.Debugf("shutdown")
+	return nil
+}
+
+func (sm *manager) expireSessions() {
+	if sm.Running {
```
byo, line 152: Access to all fields should be protected by a mutex here, otherwise we'll have race conditions (even the sm.Running boolean).

for the best performance I'd 1) lock the mutex 2) check if we're still running 3) iterate over all sessions, update state accordingly and gather DEAD sessions 4) unlock mutex 5) cleanup dead sessions e.g. rollback pending transactions - I'd avoid doing this step with locked mutex

pkg/server/sessions/options.go
```diff
@@ -0,0 +1,38 @@
...
+	MaxSessionAge time.Duration
+	// Timeout the server waits for a duration of Timeout and if no activity is seen even after that the session is closed
+	Timeout time.Duration
+}
+
+func DefaultOptions() *Options {
+	return &Options{
+		SessionGuardCheckInterval: time.Second * 1,
+		MaxSessionIdle:            time.Second * 5,
+		MaxSessionAge:             time.Second * 300,
```
byo, line 35: I'm not sure if this is a good idea. A typical web application will open the connection once and keep this for hours or even days

pkg/client/heartbeater/heartbeater.go
mmeloni: Good point, an alternative is to log but check for an heartbeater error on a client interceptor. What do you think?

byo: I think simply logging the error here would be enough. From the client perspective the only reason it sends this keepalives in a loop is to prove to the server that its still alive, not to diagnose the connection. I wouldn't interfere with other operations, and if there's a critical error with the connection, any other GRPC call should fail too.

I'm not sure if there's a simple scenario where Keepalive fails but other GRPC calls succeed. Only if the server fails to process keepalive internally (so maybe new client with an old server that do not have keepalive?). I think that it's still better to just ignore keepalive errors in such situation.

pkg/server/sessions/options.go
mmeloni: It's possible to specify by options based on uses cases

go.mod
mmeloni: this fix go mod tidy

pkg/client/client.go
mmeloni: Yes, it's enough to deprecate waitForHealthCheck

pkg/client/heartbeater/heartbeater.go
mmeloni: Context is discouraged to be used inside a struct: https://pkg.go.dev/context

pkg/client/heartbeater/heartbeater.go
mmeloni: Due to previous comment probably here I can create a context with timeout instead an empty context

pkg/client/session.go
mmeloni: I will stop it on Close session

pkg/server/sessions/manager.go
mmeloni: I would like to modify this in transaction handling PR, since it's something could be adapted during implementation

PR #997, opened 2021-11-16
feat: helm chart for deploying immudb on kubernetes

SimoneLazzaris:
I've made an helm chart for deploying immudb on helm.
It creates a statefulset with 1 replica, which is accessible as a ClusterIP or (optionally) with a couple of Ingress. 
Tested on k3s with traefik ingress class, should also work with nginx if a tls certificate is configured.


helm/templates/NOTES.txt
```diff
@@ -0,0 +1,22 @@
+1. You can access immudb web console by running these commands:
+{{- if .Values.ingress.enabled }}
+  curl http{{ if $.Values.ingress.tls }}s{{ end }}://{{ .Values.ingress.hostname }}/
+  immuclient status -a {{ .Values.ingress.hostname }} -p 80
```
byo, line 4: Shouldn't the port depend on the tls setting?

helm/templates/NOTES.txt
```diff
@@ -0,0 +1,22 @@
...
+  immuclient status -a {{ .Values.ingress.hostname }} -p 80 
+{{- else if contains "NodePort" .Values.service.type }}
+  export NODE_PORT=$(kubectl get --namespace {{ .Release.Namespace }} -o jsonpath="{.spec.ports[0].nodePort}" services {{ include "immudb.fullname" . }})
+  export NODE_IP=$(kubectl get nodes --namespace {{ .Release.Namespace }} -o jsonpath="{.items[0].status.addresses[0].address}")
+  echo http://$NODE_IP:$NODE_PORT
+{{- else if contains "LoadBalancer" .Values.service.type }}
+     NOTE: It may take a few minutes for the LoadBalancer IP to be available.
+           You can watch the status of by running 'kubectl get --namespace {{ .Release.Namespace }} svc -w {{ include "immudb.fullname" . }}'
+  export SERVICE_IP=$(kubectl get svc --namespace {{ .Release.Namespace }} {{ include "immudb.fullname" . }} --template "{{"{{ range (index .status.loadBalancer.ingress 0) }}{{.}}{{ end }}"}}")
+  echo http://$SERVICE_IP:{{ .Values.service.port }}
```
byo, line 13: `immuclient` example is missing from that and the container port example - is that intentional?

helm/templates/ingress.yaml
```diff
@@ -0,0 +1,97 @@
...
+    {{- include "immudb.labels" . | nindent 4 }}
+  {{- with .Values.ingress.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- if and .Values.ingress.className (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion) }}
+  ingressClassName: {{ .Values.ingress.className }}
+  {{- end }}
+  {{- if .Values.ingress.tls }}
```
byo, line 24: I would still use additional field here such as:
```suggestion
  {{- if .Values.ingress.tls.enabled }}
```

so that you can easily disable tls (e.g. for testing) without disabling the whole tls section.

helm/templates/ingress.yaml
```diff
@@ -0,0 +1,97 @@
...
+    {{- range .Values.ingress.tls }}
+    - hosts:
+        {{- range .hosts }}
+        - {{ . | quote }}
+        {{- end }}
+      secretName: {{ .secretName }}
+    {{- end }}
+  {{- end }}
+  rules:
+    - host: {{ $.Values.ingress.hostname | quote }}
```
byo, line 35: It's only one hostname here but the tls section allows specifying multiple hosts - is there any particular reason for that?
Also I'm wondering if hosts in the TLS section couldn't be taken from this value - to make sure those two sections are in sync (at least that's what I've been doing so far in my charts).

helm/templates/ingress.yaml
```diff
@@ -0,0 +1,97 @@
...
+                name: {{ $fullName }}-http
+                port:
+                  number: {{ $.Values.service.ports.http }}
+              {{- else }}
+              serviceName: {{ $fullName }}
+              servicePort: {{ $.Values.service.ports.http }}
+              {{- end }}
+            {{- end }}
+---
+apiVersion: {{ include "immudb.chart.ingressapiversion" . }}
```
byo, line 53: I think both ingress objects could be enabled / disabled selectively - e.g. maybe I'd like to expose immudb web console externally but still hide the grpc port.

Thist could be a future improvement though.

helm/templates/ingress.yaml
```diff
@@ -0,0 +1,97 @@
...
+  {{- with .Values.ingress.annotations }}
+  annotations:
+    nginx.ingress.kubernetes.io/backend-protocol: "GRPC"
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- if and .Values.ingress.className (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion) }}
+  ingressClassName: {{ .Values.ingress.className }}
+  {{- end }}
+  {{- if .Values.ingress.tls }}
```
byo, line 68: Comments from the previous ingress object also apply here.

helm/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.serviceAccount.create -}}
```
byo, line 1: I"m not sure if we need explicit serviceannount here, if not provided, IIRC kubernetes creates a default service account if not explicitly created.

What would be the use case for that account?

helm/templates/statefulset.yaml
```diff
@@ -0,0 +1,74 @@
...
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "immudb.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```
byo, line 33: :+1: Good approach, it's also compatible with skaffold (to have a custom immudb docker image built for example)

helm/templates/statefulset.yaml
```diff
@@ -0,0 +1,74 @@
...
+            - name: metrics
+              containerPort: 9497
+              protocol: TCP
+            - name: grpc
+              containerPort: 3322
+              protocol: TCP
+          livenessProbe:
+            httpGet:
+              path: /readyz
+              port: metrics
```
byo, line 48: I would give it a higher failure threshold to avoid restarting immudb if there's some temporary failure, or have a totally different liveness probe.

This could be a future improvement though

helm/templates/statefulset.yaml
```diff
@@ -0,0 +1,74 @@
...
+          readinessProbe:
+            httpGet:
+              path: /readyz
+              port: metrics
+          {{- if $.Values.adminPassword }}
```
byo, line 53: It would be usefull to also specify existing secret that contains the password, like:
```suggestion
          {{- if $.Values.adminPasswordExistingSecret }}
          .....
          {{- else if $.Values.adminPassword }}
```

Another thing that I see more of a future update

byo (approved):
Mostly minor comments that I don't consider blockers for the first version of this chart

helm/templates/NOTES.txt
SimoneLazzaris: not really... immuclient afaik doesn't support https, so you have to use 80

helm/templates/NOTES.txt
SimoneLazzaris: I haven't (yet) tested immuclient with the loadbalancer and the container port configuration, I'll add it as soon as I have finished testing

helm/templates/ingress.yaml
SimoneLazzaris: Good idea

helm/templates/ingress.yaml
SimoneLazzaris: I've started from a template which allowed multiple hostname in the http section, so that's a leftover from that template. I'll fix that, it has more sense

helm/templates/ingress.yaml
SimoneLazzaris: I can add a flag for enabling/disabling the ingresses

helm/templates/serviceaccount.yaml
SimoneLazzaris: It's a leftover from the helm template, I'll remove it.

helm/templates/statefulset.yaml
SimoneLazzaris: Can't take full credit for that, it's from the helm template :)

helm/templates/statefulset.yaml
SimoneLazzaris: It can be useful, but I'll focus on the others point for the moment.

helm/templates/NOTES.txt
byo: Immuclient itself does support tls connection - even with mutual tls where the client authenticates to immudb using client cert. I'm not sure however how would it work with the ingress. Maybe just setting the port to 443 and adding the --certificate option with the CA generated by kubernetes (and I assume that this wouldn't be needed if there was letsencrypt cert in front for example) would work out-of-the-box?

helm/templates/NOTES.txt
byo: right, maybe we should then add some warning that this configuration was not fully tested then? (and that any feedback is welcome of course ;))

helm/values.yaml
```diff
@@ -0,0 +1,76 @@
...
+  # If not set and create is true, a name is generated using the fullname template
+  name: ""
+
+podAnnotations: {}
+
+podSecurityContext: {}
+  # fsGroup: 2000
+
+securityContext: {}
+  # capabilities:
```
byo, line 34: Do you think we could enable this by default? The goal is to make sure immudb works with the least privileged setup we could find. If someone needs some more permission for some reason (maybe some additional initialization script) he could then provide his own `securityContext` value.

helm/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: immudb
+description: The immutable database
+type: application
+version: 0.1.0
+appVersion: "1.1.0"
```
byo, line 6: Please update the `/build/RELEASING.md` file with info to also bump this version here if a new immudb release is made.

helm/templates/NOTES.txt
SimoneLazzaris: fixed in commit 11b315756d7fd4748bbc1653d89e41ac1a9f5718

helm/templates/ingress.yaml
SimoneLazzaris: fixed in commit e4c856577131fadfea45ea1cf822acb66c64ccfe

helm/templates/ingress.yaml
SimoneLazzaris: fixed in commit e4c856577131fadfea45ea1cf822acb66c64ccfe

helm/templates/serviceaccount.yaml
SimoneLazzaris: fixed in commit 11b315756d7fd4748bbc1653d89e41ac1a9f5718

helm/templates/NOTES.txt
SimoneLazzaris: Mutual tls is complicated, as we need to get the client certificate on immudb. Possible, but I'd leave it out this initial release.
I wasn't able to find a way to use immuclient in the "traditional" https way (i.e., not mutual tls).
In any case, port 80 seems to keep working even with tls enabled, so I'll let it that way for the moment.

helm/templates/statefulset.yaml
SimoneLazzaris: Done in commit 67126206

helm/values.yaml
SimoneLazzaris: We can enable it, but we must turn off `readOnlyRootFilesystem`, as we need to write to the disk. Done in commit 5828ab09

helm/Chart.yaml
SimoneLazzaris: fixed in commit 3cad64d8

helm/values.yaml
byo: Isn't is that only the data folder requires write permissions? If so then it should be mounted as a volume and be have read-write access even if the root fs is read-only

helm/templates/NOTES.txt
byo: I was thinking  that maybe we shouldn't expose immudb's protocol through ingress at all - it will rather be used internally by other pods (in which case there's always the service object), and we can expose the http interface for the web console. For native protocols such as the immudb grpc one, I think people should use rather nodeport or loadbalancer.

helm/templates/ingress.yaml
byo: I think we should initially disable the GRPC protocol (also take a look at my [other comment about this](https://github.com/codenotary/immudb/pull/997/files#r752186311))

helm/values.yaml
```diff
@@ -0,0 +1,66 @@
...
+    metrics: 9497
+    http: 8080
+
+ingress:
+  enabled: false
+  className: ""
+  annotations: {}
+    # kubernetes.io/ingress.class: nginx
+    # kubernetes.io/tls-acme: "true"
+  hostname: immudb-example.localhost
```
byo, line 45: The `.local` TLD suffix is reserved for this kind of domains, maybe better to use:

```suggestion
  hostname: immudb-example.local
```

webconsole/statik.go
```diff
@@ -10,7 +10,7 @@ import (
...
...xb0m^\xfc\xe0\xff\xdbG\xfe\xf0/t\xe7\xc6\xed\x1fZ\xcc\xce!\x941\xe3\xf0\xddX\xe7\xab\x8e\xf0\xd0\xee\x80\xe56\x16^\xb3 \x8f3\xf8\xe9Q,\xa2\x04\x07\xf3\xd6\xcc\xad\x8b\xef\xeb\x14\x0b\x1e\xbe\xe5\xa7\x07\xa9\xcf\xa3\x8e\xd3r\x8f\xdf\\\x96O7\x0b\xa3\xed\x85\xd0\xf1\x01c\xfd\xd1.\xce\xfa\xf0WO\xc1\xbb\x98\xc77\xd7\x1d\x95T\x81\xaf\xae\x1f\x1ek\xd4\xac\xb7\x99\xee\xd1\xc1\x17`\x87S\xb8Zq\n+\xf9y\xb2o\x12\xf1\xdc\x99&/O&;\xdf\xd4\x99&^\x1c\xb9qsH(\xc3\xbb\xa6\x86\xd4\xb2\x80{\xdc\x08\xfd2\x11b\xb2\x1f+\xa6\x9c\xf4\xaf\x8f\xe9\xa4\xa4N\xdf]\xcb\x9f\xca\x05\x17\xffy\xe4>~\xebI\x00\xfa\x9d\xca\xca\x1elr\x1df\xb7-!\xf9\xd2q\xc7\x12\x0b\x0b\x0bKCU\x07Q\xafl\x13\xf1?\x01\x00\x00\xff\xffPK\x07\x08{\x86\x02V\x89p\x00\x00\x99p\x00\x00PK\x01\x02\x14\x03\x14\x00\x08\x00\x08\x00Z]lS\n\x06\xc6F\x0f\x02\x00\x00\x14\x04\x00\x00\x12\x00	\x00\x00\x00\x00\x00\x00\x00\x00\x00\xa4\x81\x00\x00\x00\x00missing/index.htmlUT\x05\x00\x01<S\x8eaPK\x01\x02\x14\x03\x14\x00\x08\x00\x08\x00\xfcN\xc9R{\x86\x02V\x89p\x00\x00\x99p\x00\x00\x12\x00	\x00\x00\x00\x00\x00\x00\x00\x00\x00\xa4\x81X\x02\x00\x00missing/mascot.pngUT\x05\x00\x01-\x90\xc0`PK\x05\x06\x00\x00\x00\x00\x02\x00\x02\x00\x92\x00\x00\x00*s\x00\x00\x00\x00"
```
byo, line 13: Accidental change?

helm/values.yaml
SimoneLazzaris: Actually, RFC [2606](https://datatracker.ietf.org/doc/html/rfc2606) reserve domains .test, .example, .invalid and **.localhost** for demostrative use. In many resolver implementation (I use unbound) every domain with TLD .localhost resolve to 127.0.0.1. That's why I've chosen that.

**.local** is used by the bonjour protocol and only resolves to the local lan.

webconsole/statik.go
SimoneLazzaris: Yes, I'm going to reverse that

helm/templates/ingress.yaml
SimoneLazzaris: Done in commit 9e5664cf

helm/templates/ingress.yaml
SimoneLazzaris: Resolved in commit 9e5664cf

helm/values.yaml
SimoneLazzaris: Done in commit 725545aa

helm/templates/statefulset.yaml
```diff
@@ -0,0 +1,97 @@
...
+      {{- with .Values.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+---
+apiVersion: v1
```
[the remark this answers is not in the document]
SimoneLazzaris: Fixed in 672a2f1d

byo (approved):
LGTM!

PR #1000, opened 2021-11-19
Change ondisk format to support nullable columns

byo:


embedded/sql/stmt.go
```diff
@@ -33,9 +33,21 @@ const (
 	catalogTablePrefix    = "CTL.TABLE."    // (key=CTL.TABLE.{dbID}{tableID}, value={tableNAME})
 	catalogColumnPrefix   = "CTL.COLUMN."   // (key=CTL.COLUMN.{dbID}{tableID}{colID}{colTYPE}, value={(auto_incremental | nullable){maxLen}{colNAME}})
 	catalogIndexPrefix    = "CTL.INDEX."    // (key=CTL.INDEX.{dbID}{tableID}{indexID}, value={unique {colID1}(ASC|DESC)...{colIDN}(ASC|DESC)})
-	PIndexPrefix          = "P."            // (key=P.{dbID}{tableID}{0}({pkVal}{padding}{pkValLen})+, value={count (colID valLen val)+})
```
jeroiraz, line 36: original prefixes had the following meaning:
"P." -> Primary
"S." -> Secondary
"U." -> Unique

I understand we need to use different ones now due to the internal encoding but I’m just wondering about the choice of "R.", "E.", "N." as replacement

byo: It's made of second letters in original words: `pRimary`, `sEcondary` and `uNique` I didn't want to increase the length of prefixes not use something different than `.` for the separator

jeroiraz (approved):

PR #1005, opened 2021-11-23
1004 use auto increment in non primary columns

### LINKED ISSUE #1004: Use auto_increment in non primary columns
It would be useful to be able to auto increment a field but not setting it as a primary key.

**Why is this needed**
In the example of the european invoicing model, the invoice number increment need to follow the rest.

But in a data model where you use a unique id to refrenence the table in the overall model, you cannot use the uid column has primary key along with the invoice number with an auto increment

0ctanium:
Close #1004

### DIFF
--- a/embedded/sql/catalog.go
+++ b/embedded/sql/catalog.go
@@ -33,17 +33,19 @@ type Database struct {
 }
 
 type Table struct {
-	db              *Database
-	id              uint32
-	name            string
-	cols            []*Column
-	colsByID        map[uint32]*Column
-	colsByName      map[string]*Column
-	indexes         map[string]*Index
-	indexesByColID  map[uint32][]*Index
-	primaryIndex    *Index
-	autoIncrementPK bool
-	maxPK           int64
+	db                 *Database
+	id                 uint32
+	name               string
+	cols               []*Column
+	colsByID           map[uint32]*Column
+	colsByName         map[string]*Column
+	indexes            map[string]*Index
+	indexesByColID     map[uint32][]*Index
+	primaryIndex       *Index
+	autoIncrementIndex *Index
+	autoIncrementPK    bool
+	autoIncrementCol   *Column
+	maxPK              int64
 }
 
 type Index struct {
@@ -232,6 +234,10 @@ func (i *Index) IsPrimary() bool {
 	return i.id == PKIndexID
 }
 
+func (i *Index) IsAutoIncrement() bool {
+	return len(i.cols) == 1 && i.cols[0].autoIncrement
+}
+
 func (i *Index) IsUnique() bool {
 	return i.unique
 }
@@ -304,7 +310,7 @@ func (db *Database) newTable(name string, colsSpec []*ColSpec) (table *Table, er
 		}
 
 		if cs.autoIncrement && cs.colType != IntegerType {
-			return nil, ErrLimitedAutoIncrement
+			return nil, ErrAutoIncrementWrongType
 		}
 
 		if cs.colType == TimestampType {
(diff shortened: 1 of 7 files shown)


embedded/sql/catalog.go
```diff
@@ -232,6 +234,10 @@ func (i *Index) IsPrimary() bool {
 	return i.id == PKIndexID
 }
 
+func (i *Index) IsAutoIncrement() bool {
```
jeroiraz, line 237: I'm wondering if it's the autoincrement concept or feature belong to the index... It may be of the column but the index may be needed in order to speed up fetching the greatest value and ensuring uniqueness, but it should be technically possible to have an autoincremental column even without a real index (yet unpractical, but just so assign "responsibilities")

embedded/sql/catalog.go
```diff
@@ -383,13 +389,23 @@ func (t *Table) newIndex(unique bool, colIDs []uint32) (index *Index, err error)
 	// having a direct way to get the indexes by colID
 	for _, col := range index.cols {
 		t.indexesByColID[col.id] = append(t.indexesByColID[col.id], index)
+
+		if col.autoIncrement {
+			// cannot technically happen if the table creation, was properly handled. But it's better to handle the error
+			if t.autoIncrementPK {
```
jeroiraz, line 395: probably we can get rid of `autoIncrementPK` at table struct, and if needed it may just be a method, thus ensuring consistency by construction

embedded/sql/engine.go
```diff
@@ -40,7 +40,10 @@ var ErrTableDoesNotExist = errors.New("table does not exist")
 var ErrColumnDoesNotExist = errors.New("column does not exist")
 var ErrColumnNotIndexed = errors.New("column is not indexed")
 var ErrLimitedKeyType = errors.New("indexed key of invalid type. Supported types are: INTEGER, VARCHAR[256] OR BLOB[256]")
+var ErrAutoIncrementWrongType = errors.New("auto incremented column need to be INTEGER type")
+var ErrAutoIncrementMultiple = errors.New("several auto incremental column were found. Wrong schema")
```
jeroiraz, line 44: given this is a generalization of the auto_increment attribute, it might be worth supporting multiple auto_increment columns... WDYT?

embedded/sql/engine.go
```diff
@@ -40,7 +40,10 @@ var ErrTableDoesNotExist = errors.New("table does not exist")
...
 var ErrLimitedAutoIncrement = errors.New("only INTEGER single-column primary keys can be set as auto incremental")
+var ErrLimitedUpsert = errors.New("upsert is only supported in tables without secondary indexes")
```
jeroiraz, line 46: rare... this limitation was already overcomed in master branch

0ctanium: oh my bad I think I re-implemented that when merging with master branch locally

embedded/sql/engine.go
0ctanium: technically it won't change anything, because the columns will all use use the same index.
but it's a bit useless because it will be a duplication of effort because the columns will always have the same value

embedded/sql/catalog.go
0ctanium: can you explain in more detail?

embedded/sql/catalog.go
0ctanium: yes, just check if the col or the index exists

0ctanium: or maybe even better. We can get rid of both `autoIncrementPK ` and  `autoIncrementCol`. 

Then add a method to check if the the table has an auto incremental column by checking if the `autoIncrementIndex` is set.
And another method to get the autoIncrementCol by getting the index column(s?)

embedded/sql/engine.go
0ctanium: with that we also could remove the `ErrLimitedAutoIncrement` error because it has been detailed `ErrAutoIncrementWrongType` (and all other cases do not generate any errors now)

embedded/sql/engine.go
jeroiraz: all auto_incremental columns will have the same value until we include support for altering the schema e.g. switching from auto_increment on and off... or setting a different initial value or delta.. but I think it's ok so have limitations along the way

embedded/sql/catalog.go
jeroiraz: that I wouldn't assign to the index the notion of "autoincrement"... but on the logic implementing autoincrement search for the index that can be used for it... it's not a functional but implementation comment

agen007niaga: Snap theif been long time sorry...

agen007niaga (commented):
0ctanium:1004-use-auto-increment-in-non-primary-columns

embedded/sql/engine.go
0ctanium: I propose to resolve this conversation for now. Then, when the feature will be available, open a new issue to support multi column auto_increment

PR #1008, opened 2021-11-25
Feat Interactive Transactions

jeroiraz:
This PR introduces an enhanced transaction support.

At the embedded key-value store it's possible to create write-only-transactions, which provide exactly the same functionality as previous `Commit(entries)` method but being able to incrementally add entries.

Additionally, it's possible to create read-write-transactions with support for read-your-own-writes i.e. changes made in during the life of the transactions are seen by subsequent statements in the same transaction.

Transactions are fully isolated, ensuring full serializable semantics and while it's possible to have concurrent transactions, the first being committed will invalidate the other ones. Read keys comparison will be addressed in a different PR.

SQL grammar and engine were updated so to use new transaction support.

Changes affecting the grammar:
- begin transaction, commit became actual statements, thus ';' may be required as statement separator
- rollback statement was added
- count aggregation was standardized i.e. count(*)


embedded/store/immustore.go
```diff
@@ -924,36 +847,44 @@ func (s *ImmuStore) appendData(entries []*EntrySpec, donec chan<- appendableResu
...
 		return nil, ErrIllegalArguments
 	}
 
-	err := s.validateEntries(txSpec.Entries)
+	err := s.validateEntries(otx.entries)
 	if err != nil {
 		return nil, err
 	}
 
+	if !otx.IsWriteOnly() && otx.snap.Ts() <= s.committedTxID {
```
byo, line 875: Can we safely read `s.committedTxID` without a lock ?

embedded/store/key_reader.go
```diff
@@ -24,23 +24,27 @@ import (
...
-type FilterFn func(valRef *ValueRef) bool
+type valueRefInterceptor func(key []byte, valRef ValueRef) ValueRef
+
+type FilterFn func(valRef ValueRef) bool
 
 var (
-	IgnoreDeleted FilterFn = func(valRef *ValueRef) bool {
-		return valRef.kvmd == nil || !valRef.kvmd.deleted
+	IgnoreDeleted FilterFn = func(valRef ValueRef) bool {
+		return valRef.KVMetadata() == nil || !valRef.KVMetadata().deleted
```
byo, line 38: Shouldn't it be:
```suggestion
		return valRef.KVMetadata() != nil && !valRef.KVMetadata().deleted
```
?

What does it mean that `KVMetadata() == nil` ? I assumed that's for old-style entries.

pkg/database/database_test.go
```diff
@@ -185,15 +185,15 @@ func TestOpenDB(t *testing.T) {
 
 func TestOpenV1_0_1_DB(t *testing.T) {
```
byo, line 186: I think we should have test for opening both 1.0.1 and 1.1.0 databases

byo (approved):
Few minor comments (although only a brief review due to the size of the PR)

embedded/store/immustore.go
jeroiraz: no it's not

embedded/store/key_reader.go
jeroiraz: entries when filter function evaluates to `false` are currently filtered out, so I'll revert so to skip when filter evaluates to true. And in such case the condition will be as you describe.
KVMetadata may be nil even in new entries, because they are optional.

jeroiraz:
Changes covered by PR #1011

PR #1011, opened 2021-11-29
Tx Header versioning

jeroiraz:
This PR includes a version record into transaction headers. The record is incorporated in a backward compatible manner with respect to previous release (v.1.1.0). 

Metadata records both in transaction header and per key-value entry are incorporated since version = 1.

The hash calculation of key-metadata-value entries since version 1 differs from previous calculation. Thus, depending on the version included in the header is the method being used (but in server and client-side verifications).

It will be possible to directly open or operate with an existent database created or at least opened with the previous release.

The changes are compatible with replication when master instance is v.1.1.0 and follower instance is built from this branch. However, the other way around is not. It's better to upgrade to this new version once the release is published.

Some minor refactoring was made into Tx struct so to de-duplicate attributes.

Note: This PR is based on #1008 (interactive transactions).


embedded/store/tx.go
```diff
@@ -25,31 +25,26 @@ import (
 )
 
 type Tx struct {
+	header *TxHeader
```
byo, line 28: Is there a reason why this is a pointer and not an instance?

embedded/store/tx.go
```diff
@@ -67,37 +62,21 @@ func NewTxWithEntries(entries []*TxEntry) *Tx {
...
-		ID:       tx.ID,
-		PrevAlh:  tx.PrevAlh,
-		Ts:       tx.Ts,
-		Metadata: tx.Metadata,
-		NEntries: tx.nentries,
-		Eh:       tx.Eh(),
-		BlTxID:   tx.BlTxID,
-		BlRoot:   tx.BlRoot,
-	}
+	return tx.header
```
byo, line 72: We're returning a pointer to internal struct - that means it can be modified outside (previously it was not possible), is that expected?

embedded/store/tx.go
```diff
@@ -67,37 +62,21 @@ func NewTxWithEntries(entries []*TxEntry) *Tx {
...
 }
 
 func (hdr *TxHeader) Bytes() []byte {
-	// ID + PrevAlh + Ts + MDLen + MD + NEntries + Eh + BlTxID + BlRoot
-	var b [txIDSize + sha256.Size + tsSize + sszSize + maxTxMetadataLen + sszSize + sha256.Size + txIDSize + sha256.Size]byte
+	// ID + Ts + PrevAlh + Version + MDLen + MD + NEntries + Eh + BlTxID + BlRoot
+	var b [txIDSize + sha256.Size + tsSize + sszSize + (sszSize + maxTxMetadataLen) + sszSize + sha256.Size + txIDSize + sha256.Size]byte
```
byo, line 77: Low-prio (not a blocker): Isn't this code very simillar to ImmuStore.performCommit ?

embedded/store/tx.go
```diff
@@ -107,11 +86,22 @@ func (hdr *TxHeader) Bytes() []byte {
...
+		}
+
+		binary.BigEndian.PutUint16(b[i:], uint16(len(mdbs)))
+		i += sszSize
+
+		copy(b[i:], mdbs)
+		i += len(mdbs)
+	}
 
 	binary.BigEndian.PutUint16(b[i:], uint16(hdr.NEntries))
```
byo, line 106: As in case of ImmuStore.performCommit, to consider whether we shouldn't store nentries in 32-bit integer in case of a v1 encoding

embedded/store/tx.go
```diff
@@ -67,37 +62,21 @@ func NewTxWithEntries(entries []*TxEntry) *Tx {
...
-		BlTxID:   tx.BlTxID,
-		BlRoot:   tx.BlRoot,
-	}
+	return tx.header
 }
 
 func (hdr *TxHeader) Bytes() []byte {
-	// ID + PrevAlh + Ts + MDLen + MD + NEntries + Eh + BlTxID + BlRoot
-	var b [txIDSize + sha256.Size + tsSize + sszSize + maxTxMetadataLen + sszSize + sha256.Size + txIDSize + sha256.Size]byte
+	// ID + Ts + PrevAlh + Version + MDLen + MD + NEntries + Eh + BlTxID + BlRoot
```
byo, line 76:
```suggestion
	// ID + PrevAlh + Ts + Version + MDLen + MD + NEntries + Eh + BlTxID + BlRoot
```

embedded/store/tx.go
```diff
@@ -209,48 +222,67 @@ func (hdr *TxHeader) innerHash() [sha256.Size]byte {
...
+	if err != nil {
+		return err
+	}
 
-// Alh calculates the Accumulative Linear Hash up to this transaction
-// Alh is calculated as hash(txID + prevAlh + hash(ts + nentries + eH + blTxID + blRoot))
-// Inner hash is calculated so to reduce the length of linear proofs
-func (tx *Tx) CalcAlh() {
-	tx.Alh = tx.Header().Alh()
+	tx.header.Eh = root
```
byo, line 279: That's something new compared to the previous body of the function - this is about fixing the bug of incorrect htree root in transaction header we discussed before?

embedded/store/tx.go
```diff
@@ -272,68 +304,78 @@ func (tx *Tx) Proof(key []byte) (*htree.InclusionProof, error) {
...
 
-	mdLen, err := r.ReadUint16()
+	version, err := r.ReadUint16()
 	if err != nil {
 		return err
 	}
+	tx.header.Version = int(version)
 
-	var txmd *TxMetadata
+	if tx.header.Version > 0 {
```
byo, line 343: Shouldn't we explicitly check for the version value here? What if there's a header with version 2 (e.g. after some future update of immudb)? We should not continue in such case.

embedded/store/verification.go
```diff
@@ -131,3 +131,59 @@ func leafFor(d [sha256.Size]byte) [sha256.Size]byte {
...
+	var mdbs []byte
+
+	if kv.Metadata != nil {
+		mdbs = kv.Metadata.Bytes()
+	}
+
+	mdLen := len(mdbs)
+	kLen := len(kv.Key)
+
+	b := make([]byte, sszSize+mdLen+sszSize+kLen+sha256.Size)
```
byo, line 169: For future consideration - sha256 has a Hasher interface t hat is io.Writer, we could do this like:
```golang
h := sha256.New()
binary.Write(h, binary.BigEndian, uint16(mdLen))
h.Write(mdbs)
....
return h.Sum(nil)
```

embedded/store/verification_test.go
```diff
@@ -89,33 +89,28 @@ func TestVerifyDualProofEdgeCases(t *testing.T) {
...
 		dproof.SourceTxHeader.BlTxID++
-		verifies = VerifyDualProof(dproof, sourceTxID, targetTxID, sourceTx.Alh, targetTx.Alh)
+		verifies = VerifyDualProof(dproof, sourceTxID, targetTxID, sourceTx.header.Alh(), targetTx.header.Alh())
 		require.False(t, verifies)
 
 		// Restore proof
 		dproof.SourceTxHeader.BlTxID--
-
-		// Alter proof
-		dproof.TargetTxHeader.BlTxID++
```
byo, line 114: Removed temporarily?

pkg/api/schema/database_protoconv.go
```diff
@@ -61,14 +61,17 @@ func TxFromProto(stx *Tx) *store.Tx {
...
+	hdr.ID = stx.Header.Id
+	hdr.Ts = stx.Header.Ts
+	hdr.PrevAlh = DigestFromProto(stx.Header.PrevAlh)
+	hdr.BlTxID = stx.Header.BlTxId
+	hdr.BlRoot = DigestFromProto(stx.Header.BlRoot)
+	hdr.Version = int(stx.Header.Version)
+	hdr.Metadata = TxMetadataFromProto(stx.Header.Metadata)
 
 	tx.BuildHashTree()
-	tx.CalcAlh()
```
byo, line 71: Why this is removed ?

byo (approved):
Reviewed only the most recent commits regarding TX header versioning - overall good, mostly minor comments that should not be a blocker for this PR, the only thing to consider is to use 32-bit nentries field if txheader.version=1.

cmd/immuclient/immuc/print.go
```diff
@@ -43,7 +40,7 @@ func PrintKV(key []byte, md *schema.KVMetadata, value []byte, tx uint64, verifie
 		}
 
 		str.WriteString(fmt.Sprintf("value:		%s \n", value))
-		str.WriteString(fmt.Sprintf("hash:		%x \n", hash))
```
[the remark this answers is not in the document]
jeroiraz: I think from immuclient the hash may not be providing something useful. Displaying it will require immuclient to receive the header (or at least the version of the tx), which may imply changing the grpc endpoints in the server

embedded/store/immustore.go
```diff
@@ -1000,55 +1005,60 @@ func (s *ImmuStore) performCommit(tx *Tx, ts int64, blTxID uint64) error {
...
-	txSize += len(txmdbs)
+		binary.BigEndian.PutUint16(s._txbs[txSize:], uint16(len(txmdbs)))
+		txSize += sszSize
+
+		copy(s._txbs[txSize:], txmdbs)
+		txSize += len(txmdbs)
+	}
 
-	binary.BigEndian.PutUint16(s._txbs[txSize:], uint16(tx.nentries))
+	binary.BigEndian.PutUint16(s._txbs[txSize:], uint16(tx.header.NEntries))
```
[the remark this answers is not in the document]
jeroiraz: hmm, I think in schema migrations the atomic action may consist in the index creation, while a background process may respect the existent transactions (or not) but queries may be blocked or not allowed until the reindexing has taken place

embedded/store/tx.go
jeroiraz: I also though about it, but left the pointer because it's exposing a pointer to the TxMetadata, which may be optional. So it won't be a safe plain copy. Do you think it should be anyways?

embedded/store/tx.go
jeroiraz: discussion in the previous comment

embedded/store/tx.go
jeroiraz: in some way yes, here the entries are not serialized but the root of the hash tree of the entries is included

embedded/store/tx.go
jeroiraz: lets have a discussion about big transactions, use cases and scalability, related to the discussion in the comments above

embedded/store/tx.go
jeroiraz: it's the same behaviour but refactored when introducing the header struct into the tx. The hash of the tx is directly calculated from the header, the root hash of the entries is set when the hash tree is calculated.

embedded/store/verification.go
jeroiraz: yes, probably the gain in performance by allocating at once is not significative

embedded/store/verification_test.go
jeroiraz: unintentional

pkg/api/schema/database_protoconv.go
jeroiraz: because Alh is directly calculated on the header.
BuildHashTree is setting in the header the root hash of the entries

embedded/store/tx.go
jeroiraz: will take a second round on this, so to make the version more explicit the version

PR #1012, opened 2021-12-02
Feat/timestamp

byo:


embedded/sql/stmt.go
```diff
@@ -1172,6 +1172,9 @@ func (e *Engine) deleteIndexEntries(
 
 type ValueExp interface {
 	inferType(cols map[string]ColDescriptor, params map[string]SQLValueType, implicitDB, implicitTable string) (SQLValueType, error)
+
```
byo, line 1175: TODO: Cleanup

embedded/sql/engine.go
```diff
@@ -1004,15 +1005,54 @@ func EncodeValue(val interface{}, colType SQLValueType, maxLen int) ([]byte, err
...
+			binary.BigEndian.PutUint64(encv[EncLenLen:], uint64(intVal.UnixNano()))
+
+			return encv[:], nil
+		}
+	}
 
 	return nil, ErrInvalidValue
 }
 
+// func timestampFrom(val interface{}) (int64, error) {
```
byo, line 1029: TODO: Cleanup

pkg/client/sql.go
```diff
@@ -330,6 +330,10 @@ func typedValueToRowValue(tv sql.TypedValue) *schema.SQLValue {
 		{
 			return &schema.SQLValue{Value: &schema.SQLValue_Bs{Bs: tv.Value().([]byte)}}
 		}
+	case sql.TimestampType:
+		{
+			return &schema.SQLValue{Value: &schema.SQLValue_Ts{Ts: 1}}
```
byo, line 335:
```suggestion
			return &schema.SQLValue{Value: &schema.SQLValue_Ts{Ts: tv.Value().(time.Time).UnixNano()}}
```

embedded/sql/sql_grammar.y
```diff
@@ -769,6 +771,16 @@ opt_not:
         $$ = true
     }
 
+is_op:
```
jeroiraz, line 774: alternatively:
```
is_op:
        IS
        {
            $$ = EQ
        }
|
        IS NOT
        {
             $$ = NE
        }
```

or even

```
    exp IS NULL
    {
        $$ = &CmpBoolExp{left: $1, op: EQ, right: &NullValue{t: AnyType}}
    }
|
    exp IS NOT NULL
    {
        $$ = &CmpBoolExp{left: $1, op: NE, right: &NullValue{t: AnyType}}
    }
```

byo: I tried the second option yesterday and couldn't get it to work, must have been some typo, today it works without issues.

embedded/sql/stmt.go
```diff
@@ -1648,6 +1718,116 @@ func (v *SysFn) selectorRanges(table *Table, asTable string, params map[string]i
...
+					t, err := time.ParseInLocation(layout, str, time.UTC)
+					if err == nil {
+						return &Timestamp{val: t.UTC()}, nil
+					}
+				}
+
+				return nil, fmt.Errorf(
+					"%w: can not cast string '%s' as a TIMESTAMP",
+					ErrIllegalArguments,
+					str[:100],
```
jeroiraz, line 1752: 100?

embedded/sql/stmt.go
```diff
@@ -1648,6 +1718,116 @@ func (v *SysFn) selectorRanges(table *Table, asTable string, params map[string]i
...
 
+type Cast struct {
+	val ValueExp
+	t   SQLValueType
+}
+
+func (c *Cast) getConverter(src, dst SQLValueType) func(TypedValue) (TypedValue, error) {
+	if dst == TimestampType {
+
+		if src == IntegerType {
```
jeroiraz, line 1729: wondering is casting from typed null values e.g. a stmt involving a varchar column and some rows having null values...

embedded/sql/stmt.go
byo: To avoid too long strings ending up in the error message (should add some ... at the end but was to lazy ;))

byo: Changed the code a bit to be easier to understand

embedded/sql/stmt.go
byo: Added handling of null values (there was a panic if cast was done with a null str or null integer)

jeroiraz: it might be also better to return an error when there is no converter for the specified types?

byo: Updated the `getConverter` method - now errors are much more descriptive.

byo: Regarding conversion from NULL to NULL timestamp - I wanted to skip this in this version because that comes with a lot of weird corner cases because NULL can be of arbitrary type, let's start with a simple solution that works in most practical use-cases and extend it later.

PR #1014, opened 2021-12-03
Feat/transaction sdk rebased

mmeloni:


pkg/auth/auth_type.go
```diff
@@ -0,0 +1,30 @@
+package auth
+
+import (
+	"context"
+	"google.golang.org/grpc/metadata"
+)
+
+type AUTH_TYPE int
```
jeroiraz, line 8: camel case

pkg/auth/auth_type.go
```diff
@@ -0,0 +1,30 @@
...
+
+const (
+	TOKEN_AUTH AUTH_TYPE = iota
```
jeroiraz, line 11: camel case

pkg/client/client.go
```diff
@@ -159,6 +168,8 @@ type ImmuClient interface {
 	DescribeTable(ctx context.Context, tableName string) (*schema.SQLQueryResult, error)
 
 	VerifyRow(ctx context.Context, row *schema.Row, table string, pkVals []*schema.SQLValue) error
+
+	BeginTx(ctx context.Context, opts *TxOptions) (Tx, error)
```
jeroiraz, line 172: I still think it's rare this API, having BeginTx at this level and Commit or Close in the Tx.. @byo ?

pkg/client/errors.go
```diff
@@ -30,11 +30,13 @@ var (
 	ErrNotConnected       = errors.New("not connected")
 	ErrHealthCheckFailed  = errors.New("health check failed")
 	ErrServerStateIsOlder = errors.New("server state is older than the client one")
+	ErrSessionAlreadyOpen = errors.New("session already open")
```
jeroiraz, line 33: session already opened

pkg/client/session.go
```diff
@@ -0,0 +1,86 @@
...
+
+	c.WithStateService(stateService)
+
+	c.Options.CurrentDatabase = database
+
+	return nil
+}
+
+func (c *immuClient) CloseSession(ctx context.Context) error {
+	if !c.IsConnected() {
```
jeroiraz, line 70: shouldn't sessionID be cleared?

pkg/client/transaction.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (c *tx) Rollback(ctx context.Context) error {
+	_, err := c.ic.ServiceClient.Rollback(ctx, new(empty.Empty))
+	return errors.FromError(err)
+}
+
+func (c *immuClient) BeginTx(ctx context.Context, options *TxOptions) (Tx, error) {
+
+	if options.TxMode == schema.TxMode_WRITE_ONLY {
+		// only in key-value mode, in sql we read catalog and write to it
```
jeroiraz, line 56: just to be sure, this validation should also be in the server

pkg/server/server.go
```diff
@@ -1058,6 +1076,18 @@ func (s *ImmuServer) UseDatabase(ctx context.Context, req *schema.Database) (*sc
 		return nil, err
 	}
 
+	if auth.GetAuthTypeFromContext(ctx) == auth.SESSION_AUTH {
```
jeroiraz, line 1079: I personally prefer this `auth.SESSION_AUTH` over camel case constants, but I think camel case is the standard in Golang ...

pkg/server/sessions/internal/transactions/transactions.go
```diff
@@ -0,0 +1,104 @@
...
+
+import (
+	"github.com/codenotary/immudb/embedded/sql"
+	"github.com/codenotary/immudb/pkg/api/schema"
+	"github.com/codenotary/immudb/pkg/database"
+	"sync"
+)
+
+type transaction struct {
+	sync.Mutex
```
jeroiraz, line 27: lock should be a private attribute

pkg/server/sessions/internal/transactions/transactions.go
```diff
@@ -0,0 +1,104 @@
...
+	transactionID string
+	sqlTx         *sql.SQLTx
+	txMode        schema.TxMode
+	db            database.DB
+	sessionID     string
+}
+
+type Transaction interface {
+	GetID() string
```
jeroiraz, line 36: Database getter may be useful for the future as well

pkg/server/sessions/internal/transactions/transactions.go
```diff
@@ -0,0 +1,104 @@
...
+	GetMode() schema.TxMode
+	Rollback() error
+	Commit() ([]*sql.SQLTx, error)
+	GetSessionID() string
+	SQLExec(request *schema.SQLExecRequest) error
+	SQLQuery(request *schema.SQLQueryRequest) (*schema.SQLQueryResult, error)
+}
+
+func NewTransaction(sqlTx *sql.SQLTx, transactionID string, mode schema.TxMode, db database.DB, sessionID string) *transaction {
```
jeroiraz, line 45: not blocking the PR but I think the NewTransaction should be created from the session struct (assigning the unique id for the tx), but the sqlTx may be created in the constructor of the transaction when executing sqlExec on the db..., no need to start with a sqlTx, it may be just nil

pkg/server/sessions/internal/transactions/transactions.go
```diff
@@ -0,0 +1,104 @@
...
+	ntx, cTxs, err := tx.db.SQLExec(&schema.SQLExecRequest{Sql: "COMMIT;"}, tx.sqlTx)
+	if err != nil {
+		return nil, err
+	}
+	tx.sqlTx = ntx
+	return cTxs, nil
+}
+
+func (tx *transaction) GetSessionID() string {
+	tx.Lock()
```
jeroiraz, line 84: could be a readLock

jeroiraz (approved):
Please check the comments

Great @mmeloni

pkg/client/client.go
mmeloni: Done

pkg/client/transaction.go
mmeloni: already therehttps://github.com/codenotary/immudb/pull/1014/files#diff-6d3b6766c2d825705f2eb1e6fd1e3e84c44c8ab7be2fb28c7190d44247fd0851R73

pkg/server/sessions/internal/transactions/transactions.go
mmeloni: yes, at the moment is not used. When it will be needed I will add

pkg/server/sessions/internal/transactions/transactions.go
mmeloni: is exactly as you described here https://github.com/codenotary/immudb/pull/1014/files/82995df19d329da19a45aec9d3aafbd78a84c93b#diff-6d3b6766c2d825705f2eb1e6fd1e3e84c44c8ab7be2fb28c7190d44247fd0851R70

PR #1017, opened 2021-12-07
logical data expiration

jeroiraz:
This PR provides support for logical data expiration, making impossible to access actual values associated to expired entries.

Export and Replicate endpoints require Admin permissions on a database.

as a first iteration, a new method was added in the client: `ExpirableSet(key, value, expiresAt)`

@mmeloni , unit testing for replication with token-based auth eventually fails, could you check if there is some issue with sessions or validations?

@byo, could you check all possibilities to access data are covered?

Note: is work is currently covering key-value endpoints

Note: we may add clock changes so to detect abrupt changes in system time and act upon it...


embedded/store/immustore_test.go
```diff
@@ -1016,6 +1016,29 @@ func TestImmudbStoreRWTransactions(t *testing.T) {
...
+		require.False(t, valRef.KVMetadata().Expirable())
+	})
+
+	t.Run("expired keys should not be reachable", func(t *testing.T) {
+		now := time.Now()
+
+		tx, err := immuStore.NewTx()
+		require.NoError(t, err)
+
+		err = tx.Set([]byte("expirableKey"), NewKVMetadata().ExpiresAt(&now), []byte("expirableValue"))
```
byo, line 1029: This could also check another key with expiration date but that has not yet expired

embedded/store/kv_metadata.go
```diff
@@ -36,22 +40,60 @@ func (md *KVMetadata) Deleted() bool {
...
+	return b[:]
 }
 
 func (md *KVMetadata) ReadFrom(b []byte) error {
 	if len(b) == 0 {
 		return nil
 	}
 
-	md.deleted = b[0]&deletedFlag != 0
+	md.deleted = b[0] != 0
```
byo, line 85: I think the first byte should still be a bit field - one for deleted, other for expiration date.
Otherwise it will be hard to extend the metadata with new fields.

pkg/api/schema/database_protoconv.go
```diff
@@ -81,7 +88,16 @@ func KVMetadataFromProto(md *KVMetadata) *store.KVMetadata {
 		return nil
 	}
 
-	return store.NewKVMetadata().AsDeleted(md.Deleted)
+	kvmd := store.NewKVMetadata()
+
+	kvmd.AsDeleted(md.Deleted)
+
+	if md.Expiration != nil {
```
byo, line 95: Is expiration mutually exclusive with deletion? Maybe we could check it here?

byo:
In https://github.com/codenotary/immudb/blob/b485e1e8953b1110f8db5fa5ac13388efc0ec6f5/pkg/database/database.go#L904 - shouldn't we check here for `ErrKeyNotFound` ? Otherwise any item in the history that has expired will return an error, even if it was set in the past and overwritten by a new value.

byo:
TxScan grpc method - could this be used to read expired keys too ?

jeroiraz:
> TxScan grpc method - could this be used to read expired keys too ?

It won't be possible to retrieve the actual value. But the key, metadata and hash of the value are provided.

embedded/store/kv_metadata.go
jeroiraz: good point, implemented a generic solution to add new attributes in the future

pkg/api/schema/database_protoconv.go
jeroiraz: I think they are not mutually exclusive, expiration is stronger than deletion

byo: How should then we interpret a record where there's both deletion and expiration set in metadata?
Should the previous record just be deleted (and expiration being just a delayed delete is covered by such deletion) or does the deletion itself expire and the record reappears after expiration time?

embedded/store/immustore_test.go
```diff
@@ -1016,6 +1016,60 @@ func TestImmudbStoreRWTransactions(t *testing.T) {
 		require.NoError(t, err)
 		require.NotNil(t, valRef)
 		require.True(t, valRef.KVMetadata().Deleted())
+		require.NotNil(t, valRef.KVMetadata())
+		require.False(t, valRef.KVMetadata().IsExpirable())
+	})
+
+	t.Run("non-expired keys should be reachable", func(t *testing.T) {
```
byo, line 1023: We should have some tests to read history with expired entries

embedded/store/key_reader.go
```diff
@@ -19,29 +19,37 @@ import (
...
 	"encoding/binary"
+	"time"
 
 	"github.com/codenotary/immudb/embedded/tbtree"
 )
 
 type Snapshot struct {
 	st             *ImmuStore
 	snap           *tbtree.Snapshot
+	ts             time.Time
```
byo, line 30: Post-merge: This doesn't look like a part of a snapshot but rather an execution context (which may contain things like snapshot, current timestamp, in the future also locally declared variables for example)

(also this time should be used as a result of the NOW() built-in function so that multiple calls to NOW() function in one statement returns the same value)

embedded/store/kv_metadata.go
```diff
@@ -15,43 +15,200 @@ limitations under the License.
...
+	"errors"
+	"fmt"
+	"time"
+)
+
+var ErrNonExpirable = errors.New("non expirable")
+
+const (
+	deletedAttrCode   attributeCode = iota
+	expiresAtAttrCode attributeCode = iota
```
byo, line 30: Post-merge: Since those values are used in on-disk format, I'd use concrete values here:

```suggestion
	deletedAttrCode   attributeCode = 0
	expiresAtAttrCode attributeCode = 1
```

otherwise it would be too easy to break things by adding new entry in the middle for example

embedded/store/kv_metadata.go
```diff
@@ -15,43 +15,200 @@ limitations under the License.
...
+
+		if len(b[i:]) < attrCodeSize {
+			return ErrCorruptedData
+		}
+
+		attrCode := attributeCode(b[i])
+		i += attrCodeSize
 
-	md.deleted = b[0]&deletedFlag != 0
+		switch attrCode {
```
byo, line 179: Post-merge: We should check for duplicates here, maybe even require that attributes appear in increasing order of attrCode? It's to avoid confusion if an attribute is duplicated

embedded/store/tx.go
```diff
@@ -441,7 +441,7 @@ func (tx *Tx) readFrom(r *appendable.Reader) error {
 				return err
```
byo, line 441: (commend for few lines above) Post-merge: We don't check if `mdLen <= maxKVMetadataLen`, this could cause out-of-bounds panic in line 439 (`mdbs[:mdLen]`)

pkg/database/database.go
```diff
@@ -901,7 +901,7 @@ func (d *db) History(req *schema.HistoryRequest) (*schema.Entries, error) {
 		}
 
 		md, val, err := d.st.ReadValue(tx, key)
-		if err != nil {
+		if err != nil && err != store.ErrKeyNotFound {
```
byo, line 904: I think we need something like that:
```suggestion
		if error.Is(err, store.ErrKeyNotFound) {
			list.Entries[i] = &schema.Entry{
				Tx:       txID,
				Key:      req.Key,
				Metadata: schema.KVMetadataToProto(.... something indicating an expired record ...),
				Value:    []byte{},
			}
			continue
		}
		if err != nil {
```

or at least protection against using nil in `TrimPrefix` below

byo (commented):
Few more comments, most can be addressed after merging this PR to let people test this feature. The only thing left is the history issue to fix.

pkg/api/schema/database_protoconv.go
jeroiraz: Note: deletion and expiration apply to individual entries (not to a key).

deleted entries may still be obtained e.g. history by key or when using embedded store.
expired entries won't be fully readable anymore (actual value is not returned), not even using the embedded store.

embedded/store/kv_metadata.go
jeroiraz: yes, can be added later on (tampering changing the state will be detected)

embedded/store/key_reader.go
jeroiraz: sure, will continue to be adapted as we continue maturing sql implementation

pkg/database/database.go
jeroiraz: as discussed offline, a dedicated expiration error was defined, making it possible to retrieve actual metadata

PR #1018, opened 2021-12-08
Fix/history crash on deleted

byo:


SimoneLazzaris:
Problem with deleted records is fixed!

jeroiraz (approved):

PR #1021, opened 2021-12-08
chore(embedded/sql): support for escaped strings

jeroiraz:
The PR was modified to provide sql standard escaping with single quotes

```sql
create table mytable(title varchar[50], primary key title)
insert into mytable(title) values ('my escaped ''string''')

select * from mytable

+---------------------------+
| (DEFAULTDB MYTABLE TITLE) |
+---------------------------+
| "my escaped 'string'"     |
+---------------------------+
```

While we can support both escaping methods, the previous one was not correctly covering all the cases.

### DIFF
--- a/embedded/sql/engine_test.go
+++ b/embedded/sql/engine_test.go
@@ -787,7 +787,7 @@ func TestUpdate(t *testing.T) {
 	_, _, err = engine.Exec("CREATE DATABASE db1", nil, nil)
 	require.NoError(t, err)
 
-	_, _, err = engine.Exec("UPDATE table1 SET title = 'title11' WHERE title = 'title", nil, nil)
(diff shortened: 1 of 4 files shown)


byo:
I think the standard way to escape single quotes in SQL is by doubling them (https://www.databasestar.com/sql-escape-single-quote/), it's not a standard notation in SQL to use `\` escaping character. I guess many standard tools will follow that convention (maybe even golang stdlib).

SimoneLazzaris:
I think that as there is at least _one_ way to insert special characters (like the quote in my case) we're safe. The backslash is quite accepted as an escape char, so I think that it is reasonable.

SimoneLazzaris (dismissed):
LGTM

SimoneLazzaris (approved):
Tested, it works

byo (approved):
:+1:

PR #1022, opened 2021-12-09
chore(pkg/server/sessions): tuning sessions params

mmeloni:
This PR modify involved session params as follows:
server
```
  IMMUDB_MAX_SESSION_AGE_TIME=0 (infinity)
  IMMUDB_MAX_SESSION_INACTIVITY_TIME=1h
  IMMUDB_SESSION_TIMEOUT=2h
  IMMUDB_SESSIONS_GUARD_CHECK_INTERVAL=1m
```
client
```
HeartBeatFrequency = 1m
```


cmd/immudb/command/init.go
```diff
@@ -60,10 +60,11 @@ func (cl *Commandline) setupFlags(cmd *cobra.Command, options *server.Options) {
 	cmd.Flags().String("s3-secret-key", "", "s3 secret access key")
 	cmd.Flags().String("s3-bucket-name", "", "s3 bucket name")
 	cmd.Flags().String("s3-path-prefix", "", "s3 path prefix (multiple immudb instances can share the same bucket if they have different prefixes)")
-	cmd.Flags().Duration("max-session-idle-time", 5*time.Second, "max session idle time")
-	cmd.Flags().Duration("max-session-age-time", 300*time.Second, "max session age time")
-	cmd.Flags().Duration("session-timeout", 7*time.Second, "session timeout")
-	cmd.Flags().Duration("sessions-guard-check-interval", 1*time.Second, "sessions guard check interval")
+	cmd.Flags().Duration("max-session-inactivity-time", 1*time.Hour, "max session inactivity time is a duration after which an active session is declared inactive by the server. A session is kept active if server is still receiving requests from client (keep-alive or other methods)")
```
byo, line 63: To consider: maybe this should be a much lower value? If client's heartbeat is sent once every minute I think 2-3 minutes will be enough. For other languages, if the heartbeat in the background is not possible, we could document that the default value on the server should be updated.

Similarly to `session-timeout` value - once the session becomes inactive, we could think about closing it after another 2-3 minutes.

@jeroiraz what do you think?

pkg/server/sessions/manager.go
```diff
@@ -195,37 +196,29 @@ func (sm *manager) expireSessions() {
...
 		if sess.GetCreationTime().Add(sm.options.MaxSessionAgeTime).Before(now) {
-			sess.setStatus(Dead)
-			sm.logger.Debugf("session %s exceeded MaxSessionAgeTime and became Dead", ID)
+			sess.setStatus(dead)
+			sm.logger.Debugf("session %s exceeded MaxSessionAgeTime and became dead", ID)
 		}
-		if sess.GetStatus() == Idle {
-			idleSessCount++
+		if sess.GetStatus() == inactive {
+			inactiveSessCount++
```
byo, line 211: Minor: inactiveSessCount should only be incremented if the session is not set to dead in the `if` below

byo (dismissed):
Only minor comment + discussion about default values. Other than that LGTM

cmd/immudb/command/init.go
mmeloni: dine

byo: :+1: `max-session-inactivity-time` sounds much better now but how about the `session-timeout`?

PR #1029, opened 2021-12-10
fix(embedded/tbtree): set fixed snapshot ts

jeroiraz:
This PR fixes tx read conflicts by ensuring all entries in the snapshot get the same logical insertion time i.e. txID


byo (dismissed):
LGTM, please add some test for this case though

PR #1031, opened 2021-12-10
Optimization safe readvalue

jeroiraz:
The initial implementation for data expiration was preventing access to values of expired entries but with a significative lost in performance.

This PR aims to solve the performance issue while still ensuring data of expired entries is not reachable.

KVMetadata struct contains a private readonly attribute used to allow or forbid changes. When reading committed data this attribute is automatically set to true, making impossible to dynamically change it using exposed apis.

### DIFF
--- a/embedded/store/immustore.go
+++ b/embedded/store/immustore.go
@@ -1626,7 +1626,7 @@ func (s *ImmuStore) ReplicateTx(exportedTx []byte, waitForIndexing bool) (*TxHea
 		var md *KVMetadata
 
 		if mdLen > 0 {
-			md = NewKVMetadata()
+			md = NewKVMetadata(true)
 
 			err := md.ReadFrom(exportedTx[i : i+mdLen])
 			if err != nil {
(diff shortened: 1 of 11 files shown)


embedded/store/immustore.go
```diff
@@ -1693,29 +1693,18 @@ func (s *ImmuStore) ReadTx(txID uint64, tx *Tx) error {
 
 // ReadValue returns the actual associated value to a key at a specific transaction
 // ErrExpiredEntry is be returned if the specified time has already elapsed
-// Note; the tx is read to be safe in cases provided tx was manipulated at runtime
-func (s *ImmuStore) ReadValue(txID uint64, txHolder *Tx, key []byte) ([]byte, error) {
-	if txHolder == nil {
+func (s *ImmuStore) ReadValue(entry *TxEntry) ([]byte, error) {
+	if entry == nil {
```
byo, line 1697: We should check if it is read only, right?

jeroiraz: yes, also ensuring readonly metadata cannot be externally instantiated

PR #1037, opened 2021-12-13
fix(pkg/server/sessions): remove transaction on read conflict error

mmeloni:


pkg/server/sessions/internal/transactions/transactions.go
```diff
@@ -67,6 +67,7 @@ func (tx *transaction) GetMode() schema.TxMode {
 func (tx *transaction) Rollback() error {
 	tx.mutex.Lock()
 	defer tx.mutex.Unlock()
+	defer func() { tx.sqlTx = nil }()
```
jeroiraz, line 70: could be after the comparison

jeroiraz (approved):
some code related to tx handling might be refactored in the future

PR #1045, opened 2021-12-18
bugfix: Transfer timestamp in microseconds to fix timestamp column values.

tauu:
This PR fixes and issues with timestamp columns in sql tables. Although timestamps are stored in microseconds, the sql client sdk converted time.Time argument values into nanoseconds ( pkg/client/sql.go , line 332 ). Afterwards the nanosecond value was converted back into a time.Time value before writing it to the database. But in this instance it was interpreted as microseconds ( pkg/api/schema/row_value.go, line 198 ). Hence a totally wrong timestamp is stored. Similarly the client sdk retrieved timestamps in nanoseconds when queried from the database.

To fix this issue and make timestamp handling more consistent, the PR adjusts all locations setting a value for SQLValue_Ts.Ts to ensure it is always assigned a microsecond value. As the conversion formula was copied to multiple locations, this was also unified to always call the functions sql.timeFromInt64 / sql.timeToInt64.

An additional PR for the webconsole will be necessary, as it apparently expects timestamps in nanoseconds and shows wrong dates at the moment.

### DIFF
--- a/embedded/sql/engine.go
+++ b/embedded/sql/engine.go
@@ -858,7 +858,7 @@ func EncodeValue(val interface{}, colType SQLValueType, maxLen int) ([]byte, err
 			// len(v) + v
 			var encv [EncLenLen + 8]byte
 			binary.BigEndian.PutUint32(encv[:], uint32(8))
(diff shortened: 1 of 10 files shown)


byo (approved):
LGTM!

PR #1052, opened 2021-12-20
support for basic insert conflict handling

jeroiraz:
This PR introduces support for basic conflict handling in INSERT INTO statements.

```sql
INSERT INTO mytable(id, title) VALUES (1, 'title1')
INSERT INTO mytable(id, title) VALUES (1, 'title1') ON CONFLICT DO NOTHING
```

Second statement will succeed even the provided primary key already exists.
If the PK is auto_incremental column, LastInsertedID will match the provided one or the latest auto-generated one.

As part of this PR restrictions of auto_incremental pk and insert/upsert statement were removed. It's possible to explicitly specify the PK value as part of a sql statement, but the provided value must be greater than current one (when a row is inserted).


embedded/sql/stmt.go
```diff
@@ -518,33 +533,57 @@ func (stmt *UpsertIntoStmt) execAt(tx *SQLTx, params map[string]interface{}) (*S
 			}
 
 			if rval.IsNull() {
-				if col.notNull {
+				if col.notNull || col.autoIncrement {
```
mmeloni, line 536: if value provided is null and col is autoincrement why we raise an error? It should be autogenerated...

jeroiraz: it shouldn't be explicitly specified as NULL, I think it could make issues harder to find if a value is autogenerated when null value is explicitly set... but we could check how other databases behave on this...

jeroiraz: checked in pgsql it behaves the same manner

mmeloni (approved):

PR #1065, opened 2021-12-27
std lib fixes

mmeloni:
last insert id is the first inserted by immudb. This is required until we will returns a reader to collect all ids of inserted row eg.
registering connection in order to make possible conn recycling.


pkg/stdlib/connector.go
```diff
@@ -27,13 +28,20 @@ type driverConnector struct {
...
 
 // Connect implement driver.Connector interface
-func (c driverConnector) Connect(ctx context.Context) (driver.Conn, error) {
-	cliOptions, err := ParseConfig(c.name)
-	if err != nil {
-		return nil, err
+func (c *driverConnector) Connect(ctx context.Context) (driver.Conn, error) {
+	var immuClientOption *client.Options
+	c.driver.configMutex.Lock()
+	immuClientOption = c.driver.clientOptions[c.name]
```
jeroiraz, line 34: immuClientOption := c.driver.clientOptions[c.name]

probably it's more compact

pkg/stdlib/driver.go
```diff
@@ -44,44 +45,60 @@ func OpenDB(cliOpts *client.Options) *sql.DB {
 	return sql.OpenDB(c)
 }
 
+/*func Open(dns string) *sql.DB {
```
jeroiraz, line 48: can be removed?

pkg/stdlib/rows_test.go
```diff
@@ -370,9 +370,13 @@ func TestRowsAffected_LastInsertId(t *testing.T) {
 		er: &schema.SQLExecResult{
 			Txs: []*schema.CommittedSQLTx{
 				{
+					UpdatedRows: 1,
 					LastInsertedPKs: map[string]*schema.SQLValue{
 						"table1": {Value: &schema.SQLValue_N{N: 1}},
 					},
+					FirstInsertedPKs: map[string]*schema.SQLValue{
```
jeroiraz, line 377: some validation on this new attribute may be missing

pkg/api/schema/sql_exec_result.go
```diff
@@ -0,0 +1,17 @@
+package schema
+
+// LastInsertedPk returns a map in which the keys are the names of the tables and the values are the primary keys of the last inserted rows.
+func (er *SQLExecResult) LastInsertedPk() map[string]*SQLValue {
```
jeroiraz, line 4: I don't find this two methods necessary. I'd just use the attribute execResult.LastInsertedPKs
if execResult is nil it means there was an error first

pkg/stdlib/driver.go
mmeloni: I will restore it
db, err := sql.Open("immudb", "immudb://immudb:immudb@127.0.0.1:3322/defaultdb?sslmode=disable")
or 
db := Open("immudb://immudb:immudb@127.0.0.1:3322/defaultdb?sslmode=disable")

are equivalent

pkg/stdlib/rows_test.go
mmeloni: what you mean? It make no sense add a new method to return FirstInsertId if in standard lib LastInsertId is used in FirstInsertId...

jeroiraz (dismissed):

PR #1069, opened 2021-12-28
restricted and optimized ExecAll

jeroiraz:
ExecAll operation provides the capability to atomically insert key-value pairs, references and sorted set entries. Even the possibility to bind references to the entries inserted into the ongoing transaction.

Due to the presence of references, the indexing must be up to date, degrading performance when execAll request is self-contained i.e. references are only created to entries written in the same request.

This PR leverages NoWait parameter to avoid waiting for the indexing to be up to date (faster processing time) at the expense of restricting references to be made to entries specified in the ongoing request.

As a side effect, not only a reference of references may be created in the database but also a reference entry can be switched to a plain key-value entry or the other way around. So to mitigate this situation, validations are incorporated when a reference or key is resolved.

### DIFF
--- a/pkg/database/all_ops.go
+++ b/pkg/database/all_ops.go
@@ -41,10 +41,14 @@ func (d *db) ExecAll(req *schema.ExecAllRequest) (*schema.TxHeader, error) {
 		return nil, ErrIsReplica
 	}
 
-	lastTxID, _ := d.st.Alh()
-	err := d.st.WaitForIndexingUpto(lastTxID, nil)
-	if err != nil {
-		return nil, err
+	const unsafe bool = true
+
+	if !unsafe {
+		lastTxID, _ := d.st.Alh()
+		err := d.st.WaitForIndexingUpto(lastTxID, nil)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	callback := func(txID uint64, index store.KeyIndex) ([]*store.EntrySpec, error) {
@@ -82,26 +86,28 @@ func (d *db) ExecAll(req *schema.ExecAllRequest) (*schema.TxHeader, error) {
 					return nil, store.ErrIllegalArguments
 				}
 
-				// check key does not exists or it's already a reference
-				entry, err := d.getAt(EncodeKey(x.Ref.Key), 0, 0, index, tx)
-				if err != nil && err != store.ErrKeyNotFound {
-					return nil, err
-				}
-				if entry != nil && entry.ReferencedBy == nil {
-					return nil, ErrFinalKeyCannotBeConvertedIntoReference
-				}
-
-				// reference arguments are converted in regular key value items and then atomically inserted
-				_, exists := kmap[sha256.Sum256(x.Ref.ReferencedKey)]
-
-				if !exists || x.Ref.AtTx > 0 {
-					// check referenced key exists and it's not a reference
-					refEntry, err := d.getAt(EncodeKey(x.Ref.ReferencedKey), x.Ref.AtTx, 0, index, tx)
-					if err != nil {
+				if !unsafe {
+					// check key does not exists or it's already a reference
+					entry, err := d.getAt(EncodeKey(x.Ref.Key), 0, 0, index, tx)
+					if err != nil && err != store.ErrKeyNotFound {
 						return nil, err
 					}
-					if refEntry.ReferencedBy != nil {
-						return nil, ErrReferencedKeyCannotBeAReference
+					if entry != nil && entry.ReferencedBy == nil {
+						return nil, ErrFinalKeyCannotBeConvertedIntoReference
+					}
+
+					// reference arguments are converted in regular key value items and then atomically inserted
+					_, exists := kmap[sha256.Sum256(x.Ref.ReferencedKey)]
+
+					if !exists || x.Ref.AtTx > 0 {
+						// check referenced key exists and it's not a reference
+						refEntry, err := d.getAt(EncodeKey(x.Ref.ReferencedKey), x.Ref.AtTx, 0, index, tx)
+						if err != nil {
+							return nil, err
+						}
+						if refEntry.ReferencedBy != nil {
+							return nil, ErrReferencedKeyCannotBeAReference
+						}
 					}
 				}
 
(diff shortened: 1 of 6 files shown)


pkg/database/all_ops.go
```diff
@@ -82,29 +88,39 @@ func (d *db) ExecAll(req *schema.ExecAllRequest) (*schema.TxHeader, error) {
...
 
-				// check key does not exists or it's already a reference
-				entry, err := d.getAt(EncodeKey(x.Ref.Key), 0, 0, index, tx)
-				if err != nil && err != store.ErrKeyNotFound {
-					return nil, err
-				}
-				if entry != nil && entry.ReferencedBy == nil {
-					return nil, ErrFinalKeyCannotBeConvertedIntoReference
+				if req.NoWait && (x.Ref.AtTx != 0 || !x.Ref.BoundRef) {
+					return nil, ErrNoWaitOperationMustBeSelfContained
```
byo, line 92: Maybe we should add some more descriptive info here:

```suggestion
					return nil, fmt.Errorf(
					    "%w: can only set references to keys added within same transaction, please use bound references with AtTx set to 0",
					    ErrNoWaitOperationMustBeSelfContained,
					)
```

pkg/database/all_ops.go
```diff
@@ -82,29 +88,39 @@ func (d *db) ExecAll(req *schema.ExecAllRequest) (*schema.TxHeader, error) {
...
 
-				// reference arguments are converted in regular key value items and then atomically inserted
 				_, exists := kmap[sha256.Sum256(x.Ref.ReferencedKey)]
 
-				if !exists || x.Ref.AtTx > 0 {
-					// check referenced key exists and it's not a reference
-					refEntry, err := d.getAt(EncodeKey(x.Ref.ReferencedKey), x.Ref.AtTx, 0, index, tx)
-					if err != nil {
+				if req.NoWait && !exists {
+					return nil, ErrNoWaitOperationMustBeSelfContained
```
byo, line 98: With some more descriptive info:
```suggestion
					return nil, fmt.Errorf(
					    "%w: can not create a reference to a key that was not set in the same transaction",
					    ErrNoWaitOperationMustBeSelfContained,
					)
```

pkg/database/all_ops.go
```diff
@@ -120,20 +136,30 @@ func (d *db) ExecAll(req *schema.ExecAllRequest) (*schema.TxHeader, error) {
 					return nil, store.ErrIllegalArguments
 				}
 
-				// zAdd arguments are converted in regular key value items and then atomically inserted
+				if req.NoWait && (x.ZAdd.AtTx != 0 || !x.ZAdd.BoundRef) {
+					return nil, ErrNoWaitOperationMustBeSelfContained
```
byo, line 140: This could use a better error message, similarly to the `Op_Ref` operation.

pkg/database/all_ops.go
```diff
@@ -120,20 +136,30 @@ func (d *db) ExecAll(req *schema.ExecAllRequest) (*schema.TxHeader, error) {
...
-				if !exists || x.ZAdd.AtTx > 0 {
-					// check referenced key exists and it's not a reference
-					refEntry, err := d.getAt(EncodeKey(x.ZAdd.Key), x.ZAdd.AtTx, 0, index, tx)
-					if err != nil {
-						return nil, err
-					}
-					if refEntry.ReferencedBy != nil {
-						return nil, ErrReferencedKeyCannotBeAReference
+				if req.NoWait && !exists {
+					return nil, ErrNoWaitOperationMustBeSelfContained
```
byo, line 146: This could use a better error message, similarly to the `Op_Ref` operation.

pkg/database/all_ops.go
```diff
@@ -82,29 +88,39 @@ func (d *db) ExecAll(req *schema.ExecAllRequest) (*schema.TxHeader, error) {
...
-				// reference arguments are converted in regular key value items and then atomically inserted
 				_, exists := kmap[sha256.Sum256(x.Ref.ReferencedKey)]
 
-				if !exists || x.Ref.AtTx > 0 {
-					// check referenced key exists and it's not a reference
-					refEntry, err := d.getAt(EncodeKey(x.Ref.ReferencedKey), x.Ref.AtTx, 0, index, tx)
-					if err != nil {
+				if req.NoWait && !exists {
+					return nil, ErrNoWaitOperationMustBeSelfContained
+				}
```
byo, line 99: We could also check here if the referenced entry from current transaction is not a reference itself,

also we could check if there was not previous entry with the same key name in the same ExecAll and that it was not changed from value to a reference with the current operation (or to restrict to only one value in whole ExecAll for one key but this could break backwards compatibility with some odd applications).

pkg/database/all_ops.go
```diff
@@ -120,20 +136,30 @@ func (d *db) ExecAll(req *schema.ExecAllRequest) (*schema.TxHeader, error) {
...
-					// check referenced key exists and it's not a reference
-					refEntry, err := d.getAt(EncodeKey(x.ZAdd.Key), x.ZAdd.AtTx, 0, index, tx)
-					if err != nil {
-						return nil, err
-					}
-					if refEntry.ReferencedBy != nil {
-						return nil, ErrReferencedKeyCannotBeAReference
+				if req.NoWait && !exists {
+					return nil, ErrNoWaitOperationMustBeSelfContained
+				}
```
byo, line 147: Same as with the `Op_Ref` - we could add some extra checks here to make recursive and cyclic references much harder to add.

byo (approved):
Overall looks good, however I think that with few extra checks we can still avoid having cyclic reference:

* If a reference of zadd is added it must be for an entry within the same transaction - that is checked, but we can also check if this entry (that's from the same transaction) is not a reference
* we should add a check for overwriting keys within same transaction - if a second occurrence of the same key exists in the transaction, we should either fail or additionally check if it does not change previous value to a reference

jeroiraz:
> Overall looks good, however I think that with few extra checks we can still avoid having cyclic reference:
> 
> * If a reference of zadd is added it must be for an entry within the same transaction - that is checked, but we can also check if this entry (that's from the same transaction) is not a reference
> * we should add a check for overwriting keys within same transaction - if a second occurrence of the same key exists in the transaction, we should either fail or additionally check if it does not change previous value to a reference

Both cases are covered.
The first one is validated because only the list of final keys is maintained for validations but not references.

Then, while execAll does not include an explicit validation, duplicated keys will be generated if a key is included multiple times in the operation. But it can be explicitly checked as well

byo:
> Both cases are covered. The first one is validated because only the list of final keys is maintained for validations but not references.
> 
> Then, while execAll does not include an explicit validation, duplicated keys will be generated if a key is included multiple times in the operation. But it can be explicitly checked as well

Ah, ok, got it, `kmap` only contains keys for values. So the only validation I'd add is to check on `Op_Ref` if the key for this reference exists in kmap and fail in such case - that should do the trick.

jeroiraz:
>

jeroiraz:
> > Both cases are covered. The first one is validated because only the list of final keys is maintained for validations but not references.
> > Then, while execAll does not include an explicit validation, duplicated keys will be generated if a key is included multiple times in the operation. But it can be explicitly checked as well
> 
> Ah, ok, got it, `kmap` only contains keys for values. So the only validation I'd add is to check on `Op_Ref` if the key for this reference exists in kmap and fail in such case - that should do the trick.

Yes. I'm adding the checking but in the pre-validations of entries

mmeloni (approved):

PR #1077, opened 2021-12-30
validate if db is not replica then other replication attrs are unset

jeroiraz:


qrdl (approved):
LGTM

PR #1079, opened 2022-01-03
Chore/metrics improvements

byo:


embedded/store/indexer.go
```diff
@@ -23,6 +23,8 @@ import (
 
 	"github.com/codenotary/immudb/embedded/tbtree"
 	"github.com/codenotary/immudb/embedded/watchers"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
```
jeroiraz, line 27: I think we should avoid adding prometheus as a direct dependency at the embedded store. I think we should think in a different approach to get the required data from the store

byo: Refactor of monitoring will be done in https://github.com/codenotary/immudb/issues/1081

jeroiraz (dismissed):

PR #1084, opened 2022-01-04
Hot backup/restore

qrdl:
Backup/restore database without stopping immudb with new commands `hot-backup` and `hot-restore`.

Uses file or stdin/stdout (default) for backup stream.
For incremental/differential backup user can specify transaction ID to start backup from, using `--start-tx` CLI option. Alternatively, user can append new data to existing file using `--append` option.
It is possible to verify backup file with `hot-restore --verify`


cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+		PersistentPreRunE: cl.ConfigChain(cl.connect),
+		PersistentPostRun: cl.disconnect,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			file := io.Writer(os.Stdout)
+			params, err := prepareBackupParams(cmd.Flags())
+			if err != nil {
+				return err
+			}
+
+			udr, err := cl.immuClient.UseDatabase(cl.context, &schema.Database{DatabaseName: args[0]})
```
jeroiraz, line 98: database name may be part of the prepared params?

cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+				}
+				defer f.Close()
+				file = f
+			}
+
+			return cl.runHotBackup(file, params.startTx, params.progress)
+		},
+		Args: cobra.ExactArgs(1),
+	}
+	ccmd.Flags().StringP("output", "o", "-", "output file, \"-\" for stdout")
```
jeroiraz, line 117: I found the use of the slash for stdout rare... what about not specifying the flag or empty value for that case?

cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+	ccmd.Flags().Uint64("start-tx", 1, "Transaction ID to start from")
```
jeroiraz, line 118: it may be worth to specify an end-tx as well ? or a limit

cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+	ccmd.Flags().Bool("progress", false, "show progress indicator")
+	ccmd.Flags().BoolP("append", "i", false, "append to file, if it already exists (for file output only)")
```
jeroiraz, line 120: not sure if it's clear the shortcut "i" in this case, probably it doesn't require it...
I'd think "i" stands for "increment"/"incremental"?

jeroiraz: based on some lines below I think "append" where append and start-tx are not allowed to be specified at the same time, I'd think append is checking in the file the initial transaction to continue the backup..

what about using a "continue" or "resume" flag instead? where the start-tx is then read from the existent file...

cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+	}
+	params.append, err = flags.GetBool("append")
+	if err != nil {
+		return nil, err
+	}
+	params.startTx, err = flags.GetUint64("start-tx")
+	if err != nil {
+		return nil, err
+	}
+	params.progress, err = flags.GetBool("progress")
```
jeroiraz, line 140: --progress-bar ?

cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+		}
+		last, fileSignature, err := lastTxInFile(f)
+		if err != nil {
+			return nil, err
+		}
+		txn, err := cl.immuClient.TxByID(cl.context, last)
+		if err != nil {
+			return nil, fmt.Errorf("cannot find file's last transaction %d in database: %v", last, err)
+		}
+		if !bytes.Equal(fileSignature, txn.Header.EH) {
```
jeroiraz, line 177: I'd use txn.Header.Alh(), which captures the entire state of the database up to that transaction.

I'd rename fileSignature to checksum, digest or just alh (alh stands for accumulative linear hash).

cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+
+	return f, nil
+}
+
+func (cl *commandlineHotBck) runHotBackup(output io.Writer, startFrom uint64, progress bool) error {
+	state, err := cl.immuClient.CurrentState(cl.context)
+	if err != nil {
+		return err
+	}
+	txnCount := state.TxId
```
jeroiraz, line 198: txnCount -> latestTx / currTx ?

cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+				}
+				file = f
+				defer f.Close()
+			}
+
+			if params.verify {
+				return verifyFile(file)
+			}
+
+			dbExist, err := cl.isDbExists(args[0])
```
jeroiraz, line 332: db name may be part of restore params

cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+			}
+
+			var signature, payload []byte
+			var firstTx uint64
+			if dbExist {
+				firstTx, signature, payload, err = cl.verifyDb(params, args[0], file)
+				if err != nil {
+					return err
+				}
+				err = cl.immuClient.UpdateDatabase(cl.context, &schema.DatabaseSettings{DatabaseName: args[0], Replica: true})
```
jeroiraz: automatic switching replica mode on/off might lead to some issues in the future, because UpdateDatabase may expect more settings which may get lost as part of this process...

Probably it's safe to switch it off upon completion in the database is created in this restore process

cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+		// database isn't empty
+		if !params.append {
+			return 0, nil, nil, errors.New("cannot restore to non-empty database without --append flag")
+		}
+		// find the nearest transaction in backup file and position the file just after the transaction
+		txId, signature, payload, err := findTx(file, state.TxId)
+		if err != nil {
+			return 0, nil, nil, err
+		}
+		gap := txId - state.TxId
```
jeroiraz: I'm not sure I follow the gap cases, which situations are being covered?

cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+				return 0, nil, nil, errors.New("not possible to validate last transaction in DB - use --force to override")
+			}
+			return txId, signature, payload, nil // indicate to caller that first transaction to restore is already read
+		}
+
+		txn, err := cl.immuClient.TxByID(cl.context, state.TxId)
+		if err != nil {
+			return 0, nil, nil, err
+		}
+		if !bytes.Equal(signature, txn.Header.EH) {
```
jeroiraz: txn.Header.Alh()

cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+		Args: func(cmd *cobra.Command, args []string) error {
+			verify, _ := cmd.Flags().GetBool("verify")
+			if verify {
+				return cobra.ExactArgs(0)(cmd, args)
+			}
+			return cobra.ExactArgs(1)(cmd, args)
+		},
+	}
+	ccmd.Flags().StringP("input", "i", "-", "input file file, \"-\" for stdin")
+	ccmd.Flags().Bool("verify", false, "do not restore data, only verify backup")
```
jeroiraz, line 369: verify-only ?

cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+			return err
+		}
+		remainder -= chunkSize
+		offset += chunkSize
+	}
+	metadata, err := stream.CloseAndRecv()
+	if err != nil {
+		return err
+	}
+	if !bytes.Equal(signature, metadata.EH) {
```
jeroiraz, line 600: Alh

cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+		if txID >= tx {
+			break
+		}
+	}
+
+	return
+}
+
+// read transaction, position stream pointer just after the transaction
+func getTx(file io.Reader) (txId uint64, signature []byte, tx []byte, err error) {
```
jeroiraz, line 649: readTx / nextTx, so to transmit the notion of a cursor as it's not an idempotent function

cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+	}
+
+	if params.output == "-" && params.append {
+		return nil, errors.New("--append option can be used only when outputting to the file")
+	}
+
+	return &params, nil
+}
+
+func (cl *commandlineHotBck) verifyBackupFile(params *backupParams) (*os.File, error) {
```
jeroiraz, line 156: shall verification be split ? as it's creating an empty file as secondary effect if the file does not exist?

cmd/immuadmin/command/hot_backup.go
qrdl: Some programs use `-` as a special filename for stdin/stdout, although there is no standard for it. I'd expect user rather omit `-o`/`-i` param at all rather then specifying `-o -`/`-i -`

cmd/immuadmin/command/hot_backup.go
qrdl: It is not a problem to add it, I just cannot imagine why it can be needed. I see a use case for specifying `start-tx` for incremental backup, but I'm not sure about `end-tx`

cmd/immuadmin/command/hot_backup.go
qrdl: It is a leftover, this param used to be called `incremental` before, but then I've renamed it to append for consistency with restore part. I'll remove the shortcut, it is indeed confusing.

cmd/immuadmin/command/hot_backup.go
qrdl: `progress` is just shorter, and I believe understandable.
Do we have anywhere else similar type of options? In `vcn`/`cas` progress is on by default, unless user specifies `--silent` or `-o json`/`-o yaml`

cmd/immuadmin/command/hot_backup.go
qrdl: Checksum is better, will change everywhere in the code. Will change `EH` to `Alh`

cmd/immuadmin/command/hot_backup.go
```diff
@@ -0,0 +1,661 @@
...
+	}
+	err = stream.CloseSend()
+	if err != nil {
+		return fmt.Errorf("CloseSend returned %v", err)
+	}
+	txn, err := cl.immuClient.TxByID(cl.context, tx)
+	if err != nil {
+		return err
+	}
+	err = outputTx(tx, output, txn.Header.EH, content)
```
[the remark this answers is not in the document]
qrdl: Will change

cmd/immuadmin/command/hot_backup.go
qrdl: Can you please elaborate on this? How it should work?

cmd/immuadmin/command/hot_backup.go
qrdl: There are three possible scenarios. Imagine the DB that has transactions 1-10.
1. User wants to append transactions 10-20 from file, there is an overlap (transaction 10 is in both DB and restore file), so restore process can validate that checksums for txn 10 in DB and file do match, to make sure user didn't mess up.
2. User wants to append transactions 11-20, in theory it is possible as there is no gap between transactions, but because there is no overlap, checksum cannot be verified. Therefore restore process wants `--force` param to make sure user really understands.
3. User wants to append transactions 12-20, there is a gap (transaction 11 is missing), restore not possible.

cmd/immuadmin/command/hot_backup.go
qrdl: Will change

cmd/immuadmin/command/hot_backup.go
qrdl: Agree, `verify-only` is more clear, will change

cmd/immuadmin/command/hot_backup.go
qrdl: Will change

cmd/immuadmin/command/hot_backup.go
qrdl: Agree, will change

zaza81:
@qrdl  can you add documentation on how to run?

cmd/immuadmin/command/hot_backup.go
qrdl: this function was renamed to make its purpose clear

cmd/immuadmin/command/hot_backup.go
qrdl: changed

cmd/immuadmin/command/hot_backup.go
qrdl: changed

cmd/immuadmin/command/hot_backup.go
qrdl: changed

cmd/immuadmin/command/hot_backup.go
qrdl: changed

cmd/immuadmin/command/hot_backup.go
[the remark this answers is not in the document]
qrdl: changed

cmd/immuadmin/command/hot_backup.go
qrdl: changed

cmd/immuadmin/command/hot_backup.go
jeroiraz: end-tx or limit is to set some limit when the db is big enough, but might be added in the future

cmd/immuadmin/command/hot_backup.go
jeroiraz: curl uses --progress-bar. seeing 'progress' in the source code is not immediate to know if it refers to a progress bar or to some progress measure done during the process

cmd/immuadmin/command/hot_backup.go
jeroiraz: if restore is called into existent database it will fail if it's not in replica mode.
if restore is called into non-existent database, it could be automatically created as replica and may be set as non-replica when all the transactions has been applied.

I'd discourage calling update if there is not knowledge about all the settings of the database.

cmd/immuadmin/command/hot_backup.go
jeroiraz: either if there is overlap or a gap, the replication will automatically fail, validations are already in place in the store.

probably is simpler to display information about what's the latest tx in the database, and what are the available tx in the backup file?

cmd/immuadmin/command/hot_backup.go
qrdl: So you are suggesting that user need to manually switch existing database to replica mode before running restore? This isn't very user friendly. Maybe it is better to add CLI option to force switching to `replica`?

cmd/immuadmin/command/hot_backup.go
qrdl: > either if there is overlap or a gap, the replication will automatically fail, validations are already in place in the store.

Yes, but the error message is very cryptic, user cannot understand the source of the problem. I test it in order to give better diagnostics

cmd/immuadmin/command/hot_backup.go
qrdl: changed

cmd/immuadmin/command/hot_backup.go
qrdl: changed

cmd/immuadmin/command/hot_backup.go
qrdl: changed

cmd/immuadmin/command/hot_backup.go
jeroiraz: good idea about the flag, it actually could be the "force" flag...

cmd/immuadmin/command/hot_backup.go
jeroiraz: understood

cmd/immuadmin/command/hot_backup.go
qrdl: Added `force-replica` flag

jeroiraz (approved):
attributes names in restoreParams struct may not be fully descriptive e.g. progress, force, verify, replica

verifyOnly?, forceReplica (not clear difference with force atttribute), force (was related to validation?), showProgress/displayProgress/progressBar ?

qrdl:
> @qrdl can you add documentation on how to run?

@zaza81  Docs are in this PR: https://github.com/codenotary/immudb.io/pull/110

zaza81:
- I see strange chars, it's not helping very much users. I suppose this is the transaction id. 

Screenshot 2022-01-11 at 11 54 25

- I think it's better to change from to, the notation n - n was not clear to me.
Screenshot 2022-01-11 at 12 35 52

- after restoring the defaultdb from another immudb instance (same version), I get the following message:

Screenshot 2022-01-11 at 12 38 10

- then a small thing not related to this PR. When typing wrong password I get the message `invalid user name`. We should correct to `invalid username`

zaza81 (changes requested):
- I see strange chars, it's not helping very much users. I suppose this is the transaction id. 

Screenshot 2022-01-11 at 11 54 25

- I think it's better to change from to, the notation n - n was not clear to me.
Screenshot 2022-01-11 at 12 35 52

- after restoring the defaultdb from another immudb instance (same version), I get the following message:

Screenshot 2022-01-11 at 12 38 10

- then a small thing not related to this PR. When typing wrong password I get the message `invalid user name`. We should correct to `invalid username`


tested:

- [x] full hot backup to file 
- [x] full hot backup to console
- [x] incremental hot backup to file 
- [x] full hot restore to new db
- [ ] incremental hot restore to existing db

qrdl:
> I see strange chars, it's not helping very much users [...]

Because you didn't specify output file and didn't redirect the output to somewhere, you just see content of backup file, because it is outputted to `stdout`.

> I think it's better to change from to, the notation n - n was not clear to me.

Ok, will change it to `from n to n`

> after restoring the defaultdb from another immudb instance [...]

I don't think it is related to backup/restore

> then a small thing not related to this PR. When typing wrong password I get the message invalid user name. We should correct to invalid username

It says `invalid user name or password`, so I guess `user` is related to both `name` and `password`, not just to `name`, in this case it is correct.

zaza81 (approved):

jeroiraz (approved):

mmeloni (approved):

PR #1085, opened 2022-01-05
Feat/float support

mmeloni:
this add support for float type.
Please let me know if there are corner cases not yet covered


embedded/sql/catalog.go
```diff
@@ -404,6 +404,8 @@ func (c *Column) MaxLen() int {
 		return 8
 	case TimestampType:
 		return 8
+	case FloatType:
```
jeroiraz, line 407: Float64Type ? FloatDoubleType?

embedded/sql/engine.go
```diff
@@ -865,6 +868,22 @@ func EncodeValue(val interface{}, colType SQLValueType, maxLen int) ([]byte, err
...
+		{
+			floatVal, ok := val.(float64)
+			if !ok {
+				return nil, fmt.Errorf(
+					"value is not a float: %w", ErrInvalidValue,
+				)
+			}
+
+			var encv [EncLenLen + 8]byte
+			floatBits := math.Float64bits(floatVal)
```
jeroiraz, line 883: so this encoding preserves lex order with negative values as well, right?

embedded/sql/sql_grammar.y
```diff
@@ -35,7 +35,8 @@ func setResult(l yyLexer, stmts []SQLStmt) {
     values []ValueExp
     value ValueExp
     id string
-    number uint64
+    integer uint64
+	float float64
```
jeroiraz, line 39: extra space at line 39

embedded/sql/stmt.go
```diff
@@ -1668,6 +1669,75 @@ func (v *Blob) Compare(val TypedValue) (int, error) {
...
+func (v *Float) IsNull() bool {
+	return false
+}
+
+func (v *Float) inferType(cols map[string]ColDescriptor, params map[string]SQLValueType, implicitDB, implicitTable string) (SQLValueType, error) {
+	return FloatType, nil
+}
+
+func (v *Float) requiresType(t SQLValueType, cols map[string]ColDescriptor, params map[string]SQLValueType, implicitDB, implicitTable string) error {
+	if t != TimestampType {
```
jeroiraz, line 1689: FloatType

jeroiraz: and probably Integer can be converted into Float, but that would be in integer method

embedded/sql/stmt.go
```diff
@@ -1912,6 +1982,8 @@ func (p *Param) substitute(params map[string]interface{}) (ValueExp, error) {
 		{
 			return &Timestamp{val: v}, nil
 		}
+	case float64:
+		return &Float{val: v}, nil
```
jeroiraz, line 1986: so to maintain same style:

{
     return &Float{val: v}, nil
}

jeroiraz (commented):
please add a case for float in method zeroForType in file grouped_row_reader.go

please keep the rest as it's (timestamp is commented), but it's already fixed in another ongoing branch.

jeroiraz:
test cases covering NaN and Infinity may be missing

embedded/sql/engine_test.go
```diff
@@ -341,6 +343,198 @@ func TestTimestampCasts(t *testing.T) {
...
+	err = engine.SetDefaultDatabase("db1")
+	require.NoError(t, err)
+
+	_, _, err = engine.Exec("CREATE TABLE IF NOT EXISTS float_index (id INTEGER AUTO_INCREMENT, ft FLOAT, PRIMARY KEY id)", nil, nil)
+	require.NoError(t, err)
+
+	_, _, err = engine.Exec("CREATE INDEX ON float_index(ft)", nil, nil)
+	require.NoError(t, err)
+
+	floatSerie := []float64{-1.0, 3.345, -0.5, 0.0, -100.8, 0.5, 1.0, math.MaxFloat64, -math.MaxFloat64}
```
byo, line 514: Other interesting test cases to check:

* +0.0 vs -0.0 - those can have different representations in floating points although those should end up next to each other in topological sorting order (also to confirm - compiler may convert -0.0 value to 0.0 thus it should be confirmed that that's actually a negative 0 value)
* denormalized values - I think `math.SmallestNonzeroFloat64` should be such value - it should sort correctly compared to 0.0 and other numbers, also a good test would be with greatest positive denormalized value and smallest positive normalized value
* +/- Inf
* +/- NaN

byo (commented):
Additional we could add those casts:
* float <-> integer
* float <-> timestamp (this is needed to allow sub-second precision for cast results)

byo:
Missing:

- [ ] immuclient doesn't show float values
- [ ] webconsole support

PR #1089, opened 2022-01-11
feat(s3): Add support for AWS V4 signatures

byo:


SimoneLazzaris (approved):
Tested with an instance on a VPC, on a region without V2 authentication support, it works.
If I test without specifying a location (thus using V2 auth protocol) it's not working (400 bad request)

PR #1096, opened 2022-01-13
sql union operator

jeroiraz:
This PR introduces support for UNION operator.

All subqueries must select the same number of columns, with the same types. While names can differ, the final return will assign the naming as in the first subquery.

```sql
SELECT * FROM mytable WHERE id < 10 UNION SELECT * FROM mytable WHERE id > 10

SELECT MAX(age) FROM mytable1 UNION ALL SELECT MIN(age) FROM mytable2
```

### DIFF
--- a/embedded/sql/union_row_reader.go
+++ b/embedded/sql/union_row_reader.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 CodeNotary, Inc. All rights reserved.
+Copyright 2022 CodeNotary, Inc. All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/embedded/sql/union_row_reader.go
+++ b/embedded/sql/union_row_reader.go
@@ -1,0 +1,159 @@
+/*
+Copyright 2021 CodeNotary, Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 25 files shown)


byo:
@jeroiraz please rebase on current master

embedded/sql/union_row_reader.go
```diff
@@ -0,0 +1,159 @@
+/*
+Copyright 2021 CodeNotary, Inc. All rights reserved.
```
byo, line 2:
```suggestion
Copyright 2022 CodeNotary, Inc. All rights reserved.
```

pkg/integration/transaction_test.go
```diff
@@ -107,7 +107,7 @@ func TestTransaction_Rollback(t *testing.T) {
 
 	res, err := tx1.SQLQuery(context.TODO(), "SELECT * FROM table1", nil)
 	require.Error(t, err)
```
byo, line 109: How about `ErrorIs` here?

pkg/stdlib/tx_test.go
```diff
@@ -75,7 +75,7 @@ func TestConn_BeginTx(t *testing.T) {
 	_, err = db.Exec(fmt.Sprintf("INSERT INTO %s (id, amount, total, title, content, isPresent) VALUES (1, 1000, 6000, 'title 1', x'%s', true)", table, blobContent))
 	require.Error(t, err)
```
byo, line 76: How about `ErrorIs` here?

byo (approved):
LGTM, only some minor comments added

PR #1102, opened 2022-01-21
sql catalog queries

jeroiraz:
This PR extends SQL grammar and processing to solve catalog related queries and database creation.

Note: database selection from SQL is not supported with deprecated token-based authentication

```sql
CREATE DATABASE db1;

USE db1;

SELECT * FROM DATABASES();

+--------------------+
| (* DATABASES NAME) |
+--------------------+
| "defaultdb"        |
| "db1"              |
| "db2"              |
+--------------------+

SELECT * FROM  TABLES();

+-------------------------+
| (DEFAULTDB TABLES NAME) |
+-------------------------+
| "mytable"               |
| "mytable2"              |
+-------------------------+

SELECT * FROM INDEXES('mytable');

+---------------------------+--------------------------+----------------------------+-----------------------------+
| (DEFAULTDB INDEXES TABLE) | (DEFAULTDB INDEXES NAME) | (DEFAULTDB INDEXES UNIQUE) | (DEFAULTDB INDEXES PRIMARY) |
+---------------------------+--------------------------+----------------------------+-----------------------------+
| "mytable"                 | "mytable[id]"            | true                       | true                        |
| "mytable"                 | "mytable[title]"         | false                      | false                       |
+---------------------------+--------------------------+----------------------------+-----------------------------+

SELECT * FROM COLUMNS('mytable');

+---------------------------+--------------------------+--------------------------+--------------------------------+------------------------------+------------------------------------+-----------------------------+-----------------------------+----------------------------+
| (DEFAULTDB COLUMNS TABLE) | (DEFAULTDB COLUMNS NAME) | (DEFAULTDB COLUMNS TYPE) | (DEFAULTDB COLUMNS MAX LENGTH) | (DEFAULTDB COLUMNS NULLABLE) | (DEFAULTDB COLUMNS AUTO INCREMENT) | (DEFAULTDB COLUMNS INDEXED) | (DEFAULTDB COLUMNS PRIMARY) | (DEFAULTDB COLUMNS UNIQUE) |
+---------------------------+--------------------------+--------------------------+--------------------------------+------------------------------+------------------------------------+-----------------------------+-----------------------------+----------------------------+
| "mytable"                 | "id"                     | "INTEGER"                |                              8 | true                         | true                               | true                        | true                        | true                       |
| "mytable"                 | "title"                  | "VARCHAR"                |                            200 | true                         | false                              | true                        | false                       | false                      |
+---------------------------+--------------------------+--------------------------+--------------------------------+------------------------------+------------------------------------+-----------------------------+-----------------------------+----------------------------+
 
```


byo:
@jeroiraz please rebase on current master

embedded/sql/catalog.go
```diff
@@ -272,6 +269,30 @@ func (i *Index) prefix() string {
 	return SIndexPrefix
 }
 
+func (i *Index) Name() string {
+	return indexName(i.table.name, i.cols)
+}
+
+func indexName(tableName string, cols []*Column) string {
+	var buf bytes.Buffer
```
byo, line 277: why not use `string.Builder` https://pkg.go.dev/strings#Builder ?

embedded/sql/engine.go
```diff
@@ -1110,14 +1118,32 @@ func (e *Engine) ExecPreparedStmts(stmts []SQLStmt, params map[string]interface{
...
+	changedDB := false
+
 	for _, stmt := range stmts {
 		if stmt == nil {
 			return nil, nil, ErrIllegalArguments
 		}
 
+		_, isDBSelectionStmt := stmt.(*UseDatabaseStmt)
+
+		if changedDB && e.multidbHandler != nil && isDBSelectionStmt {
```
byo, line 1130: I don't understant this condition, why we're checking for `e.multidbHandler != nil` here?

Shouldn't `UseDatabaseStmt` always fail if there's no `e.multidbHandler` set ?

embedded/sql/engine_test.go
```diff
@@ -1000,6 +1005,75 @@ func TestTransactions(t *testing.T) {
...
+	})
+
+	_, _, err = engine.Exec("CREATE DATABASE db1", nil, nil)
+	require.NoError(t, err)
+
+	err = engine.SetDefaultDatabase("db1")
+	require.NoError(t, err)
+
+	_, _, err = engine.Exec(`CREATE TABLE table1 (
+										id INTEGER,
```
byo, line 1038: Something's messed up with the indentation here

embedded/sql/engine_test.go
```diff
@@ -4714,3 +4785,280 @@ func TestUnionOperator(t *testing.T) {
...
+
+		_, err = r.Read()
+		require.ErrorIs(t, err, ErrNoMoreRows)
+
+		err = r.Close()
+		require.NoError(t, err)
+	})
+}
+
+type multidbHandler struct {
```
byo, line 4853:
```suggestion
type multidbHandlerMock struct {
```

embedded/sql/sql_grammar.y
```diff
@@ -569,17 +570,32 @@ ds:
...
+        $$ = &ListIndexesStmt{tableRef: $1, as: $4}
+    }
 
 tableRef:
     IDENTIFIER
     {
         $$ = &tableRef{table: $1}
     }
-|
-    IDENTIFIER '.' IDENTIFIER
```
byo, line 579: This is breaking SQL syntax compatibility - we should confirm if this is allowed.

Alternatively we could just use some simple prefix for schema-based queries,

```sql
SELECT * FROM @table.INDEXES;
SELECT * FROM @DATABASES;
```

that would make the syntax unambiguous.

embedded/sql/values_row_reader.go
```diff
@@ -0,0 +1,197 @@
+/*
+Copyright 2021 CodeNotary, Inc. All rights reserved.
```
byo, line 2:
```suggestion
Copyright 2022 CodeNotary, Inc. All rights reserved.
```

embedded/sql/values_row_reader.go
```diff
@@ -0,0 +1,197 @@
...
+
+func (vr *valuesRowReader) Close() error {
+	if vr.closed {
+		return ErrAlreadyClosed
+	}
+
+	vr.closed = true
+
+	if vr.onCloseCallback != nil {
+		defer vr.onCloseCallback()
```
byo, line 193:
```suggestion
		vr.onCloseCallback()
```

pkg/server/db_operations.go
```diff
@@ -1,3 +1,18 @@
+/*
+Copyright 2021 CodeNotary, Inc. All rights reserved.
```
byo, line 2:
```suggestion
Copyright 2022 CodeNotary, Inc. All rights reserved.
```

embedded/sql/engine.go
jeroiraz: the embedded still supports multiple databases, but if there is a multidbHandler set, the multi-database operations are delegated to it. 
But the selection of the database can not happen in the middle of a transaction

byo: If database can not be changed then why it's not this one?

```suggestion
		if changedDB {
```

(that flag would have to be then reset right after `currTx.commit()` few lines below).

embedded/sql/stmt.go
```diff
@@ -182,6 +198,14 @@ func (stmt *CreateDatabaseStmt) inferParameters(tx *SQLTx, params map[string]SQL
...
+		if tx.explicitClose {
+			return nil, fmt.Errorf("%w: database creation can not be done within a transaction", ErrNonTransactionalStmt)
+		}
+
+		return nil, tx.engine.multidbHandler.CreateDatabase(tx.ctx, stmt.DB)
+	}
+
 	id := uint32(len(tx.catalog.dbsByID) + 1)
 
 	db, err := tx.catalog.newDatabase(id, stmt.DB)
```
byo, line 211: In case of no `multidbHandler` it will be executed even when we're in the transaction context - is that OK? Maybe we should constraint this too?

embedded/sql/stmt.go
```diff
@@ -2342,26 +2390,19 @@ func (i periodInstant) resolve(tx *SQLTx, params map[string]interface{}, asc, in
...
-	var db *Database
-
-	if stmt.db != "" {
-		rdb, err := tx.catalog.GetDatabaseByName(stmt.db)
-		if err != nil {
-			return nil, err
-		}
-
-		db = rdb
+	if stmt.db != "" && stmt.db != tx.currentDB.name {
```
byo, line 2393: The check for `if tx.currentDB == nil {` should be moved above this one.

Otherwise `tx.currentDB.name` will crash here if database is not selected.

embedded/sql/engine.go
```diff
@@ -1116,9 +1125,32 @@ func (e *Engine) ExecPreparedStmts(stmts []SQLStmt, params map[string]interface{
 			return nil, nil, ErrIllegalArguments
 		}
 
+		_, isDBSelectionStmt := stmt.(*UseDatabaseStmt)
+
+		// handle the case when working in non-autocommit mode outside a transaction block
+		if isDBSelectionStmt && (currTx != nil && !currTx.closed) && !currTx.explicitClose {
+			err = currTx.commit()
```
byo, line 1132: I'm still not sure if this is a good idea,

IIRC this would only happen in such situation:

* autocommit is set to false
* there are some statements already executed - thus this is like an implicit transaction
* some use db statement is given

What worries me here is that this is a place where implicit commit happens. What if use database or any of the statements that follow fail? That would mean that we've executed part of the statements and failed with others.

jeroiraz: when autocommit is disabled, the exact number of transactions may be unknown upfront. In case there is an ongoing tx and one of the following statements if next: `BEGIN TRANSACTION`,  `USE DATABASE`, the ongoing tx is committed and a new one is initialized

byo (dismissed):
LGTM, but make sure we're above the coverage badge limit ;)

PR #1106, opened 2022-01-27
chore(stats): Add btree cache prometheus stats

byo:
This is a very simple implementation, to be changed in https://github.com/codenotary/immudb/issues/1081

### DIFF
--- a/embedded/tbtree/tbtree.go
+++ b/embedded/tbtree/tbtree.go
@@ -34,6 +34,8 @@ import (
 	"github.com/codenotary/immudb/embedded/cache"
 	"github.com/codenotary/immudb/embedded/multierr"
 	"github.com/codenotary/immudb/pkg/logger"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
 var ErrIllegalArguments = errors.New("illegal arguments")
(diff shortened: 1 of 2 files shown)


embedded/tbtree/tbtree.go
```diff
@@ -478,26 +480,54 @@ func (t *TBtree) cachePut(n node) {
...
+var metricsCacheEvict = promauto.NewCounterVec(prometheus.CounterOpts{
+	Name: "immudb_btree_cache_evict",
+}, []string{"id"})
+
 func (t *TBtree) nodeAt(offset int64) (node, error) {
 	t.nmutex.Lock()
 	defer t.nmutex.Unlock()
 
+	size := t.cache.EntriesCount()
+	metricsCacheSizeStats.WithLabelValues(t.path).Set(float64(size))
```
jeroiraz, line 510: is it needed every time a node is read?

jeroiraz (dismissed):

embedded/tbtree/tbtree.go
byo: nope, I think we can update it only when modifying the cache - but in general it should be very lightweight operation, prometheus internals are highly optimized for such frequently changing metrics.

I've added some comments about that to https://github.com/codenotary/immudb/issues/1081

PR #1109, opened 2022-01-27
Chore expose store parameters

jeroiraz:


embedded/cache/lru_cache.go
```diff
@@ -49,6 +51,17 @@ func NewLRUCache(size int) (*LRUCache, error) {
 	}, nil
 }
 
+func (c *LRUCache) Resize(size int) {
```
byo, line 54: I don't see this method used in the code other than tests, is this a preparation for future changes about live updates of parameters?

embedded/cache/lru_cache.go
```diff
@@ -49,6 +51,17 @@ func NewLRUCache(size int) (*LRUCache, error) {
...
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	for size < c.lruList.Len() {
+		c.evict()
```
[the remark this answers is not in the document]
jeroiraz: good point. probably is not safe to rely on the data is in the cache so much. However the cache is meant to be a private component and allowing to be resized could be decided by the component itself. e.g. it's safe to evict nodes from the btree cache and it won't heart to reduce cache size. On the other hand, appendable interface is not exposing the cache nor a way to dynamically resize it

embedded/cache/lru_cache.go
jeroiraz: haven't labeled the PR as WIP,  yes, it's under preparation

jeroiraz:
Covered by #1112

PR #1114, opened 2022-02-07
add <> as alternative to != in sql queries

tauu:
This PR adds support for using `<>` in a not equal comparison in SQL queries. The current syntax `!=` for the operator is not affected by the change and can still be used.

The ANSI SQL standard defines `<>` as symbol for a not equal operator, but not `!=` (see e.g. https://en.wikipedia.org/wiki/Where_(SQL) ). Nevertheless most database engines (e.g. mysql, postgres)  support both `<>` and `!=` and  both can be used interchangeably.  To improve compatibility with automatic SQL statement generators, it would be advantageous to also support both. For example gorm creates SQL statements with `<>` in where clauses.

A Unit test for parsing `<>` as well as `!=` in a where clause is included in the PR.

### DIFF
--- a/embedded/sql/parser.go
+++ b/embedded/sql/parser.go
@@ -115,6 +115,7 @@ var boolValues = map[string]bool{
 var cmpOps = map[string]CmpOperator{
 	"=":  EQ,
 	"!=": NE,
+	"<>": NE,
 	"<":  LT,
 	"<=": LE,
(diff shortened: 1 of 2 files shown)


jeroiraz:
Super. Thanks a lot @tauu

jeroiraz (approved):

PR #1120, opened 2022-02-11
Create home directory for user `immu`

mageekguy:
The home directory of `immu` user is missing in the image, so the following error `open /home/immu/token_admin: no such file or directory` occurred when `./bin/immuadmin login immudb` is done, and it is impossible to use `immuadmin`.
This PR just create the missing directory and set the right owner on it to fix this problem.

### DIFF
--- a/Dockerfile.immuadmin
+++ b/Dockerfile.immuadmin
@@ -21,6 +21,8 @@ RUN addgroup --system --gid $IMMU_GID immu && \
     mkdir -p "$IMMUADMIN_TOKENFILE" && \
     chown -R immu:immu "$IMMUADMIN_TOKENFILE" && \
     chmod +x /usr/local/bin/immuadmin
+	
+RUN mkdir -p /home/immu && chown immu:immu /home/immu
 
 USER immu
 ENTRYPOINT ["/usr/local/bin/immuadmin"]


jeroiraz:
thanks @mageekguy !

Dockerfile.immuadmin
```diff
@@ -21,6 +21,8 @@ RUN addgroup --system --gid $IMMU_GID immu && \
     mkdir -p "$IMMUADMIN_TOKENFILE" && \
     chown -R immu:immu "$IMMUADMIN_TOKENFILE" && \
     chmod +x /usr/local/bin/immuadmin
+	
+RUN mkdir -p /home/immu && chown immu:immu /home/immu
```
jeroiraz, line 25: the flag `--no-create-home` when adding immu user, because token file is supposed to be stored at the path specifiied by IMMUADMIN_TOKENFILE.

@mmeloni, there was an ongoing issue with TOKENFILE path, was it already merged?

mageekguy: I don’t see any ongoing issue with the creation of immu home directory.
However, if the directory should be defined by IMMUADMIN_TOKENFILE, the directory should be created according to default value of this variable, and it seems that it is not the case.

mmeloni: Hi all,
in this PR https://github.com/codenotary/immudb/pull/1098 it's possible to specify the token path. If not provided it's stored under [XDG_STATE_HOME](https://github.com/adrg/xdg) . Not yet merged but it will

PR #1141, opened 2022-02-25
btree block alignment

jeroiraz:


embedded/appendable/appendable.go
```diff
@@ -54,6 +59,10 @@ type Appendable interface {
 	CompressionLevel() int
 }
 
+func PaddingLen(sz, blockSize int) int {
+	return blockSize - sz%blockSize
```
byo, line 63: This will return incorrect result if `sz` is a multiple of `blockSize` - in such case the `PaddingLen` result will be `blockSize` where it should be  `0`

PR #1142, opened 2022-03-01
chore(metrics): Add better flush / compaction metrics for btree

byo:
* Add correct compaction metrics (previously merged with flush)
* Add label for nodes count to distinguish between inner and leaf nodes
* Add counters for number of values
* Add progress reporting during compaction


jeroiraz (approved):
`buildWriteProgressOutput` is the basis for completing the abstraction and remove the direct dependency of prometheus from the embedded package in the future

PR #1146, opened 2022-03-10
index maintenance

jeroiraz:
This PR aims to simplify index maintenance by automated erasure of unreferenced data (garbage generated as newer snapshots of the index are flushed to disk).

Database index setting `CleanupPercentage` is used to determine the relative amount of garbage to be removed.

Additionally, and new endpoint (and client method) called `FlushIndex` was introduced to make an explicit index cleanup.

Compared to index compaction, the cleanup process takes place into the same ongoing files and flushing is a blocking operation i.e. indexing of new entries is not possible while flushing. However, it can be as light-weight as needed in order to prevent normal operations be affected


embedded/tbtree/tbtree.go
```diff
@@ -1073,48 +1089,82 @@ func (t *TBtree) flushTree(synced bool) (wN int64, wH int64, err error) {
...
+			t.path, t.root.ts(), cleanupPercentage, err)
 	}
 
 	t.insertionCountSinceFlush = 0
-	t.log.Infof("Index '%s' {ts=%d} successfully flushed", t.path, t.root.ts())
+	t.log.Infof("Index '%s' {ts=%d, cleanup_percentage=%d} successfully flushed",
+		t.path, t.root.ts(), cleanupPercentage)
 
 	if sync {
 		err = t.cLog.Sync()
```
byo, line 1119: nLog and hLog should be synced too

embedded/tbtree/tbtree.go
```diff
@@ -1211,7 +1261,7 @@ func (t *TBtree) Compact() (uint64, error) {
 		return 0, ErrCompactionThresholdNotReached
 	}
 
-	_, _, err := t.flushTree(false)
+	_, _, err := t.flushTree(0, false)
```
byo, line 1264: Shouldn't this be 
```suggestion
	_, _, err := t.flushTree(100, false)
```
?

embedded/tbtree/tbtree.go
```diff
@@ -1434,7 +1481,7 @@ func (t *TBtree) IncreaseTs(ts uint64) error {
 	t.insertionCountSinceSync++
 
 	if t.insertionCountSinceFlush >= t.flushThld {
-		_, _, err := t.flushTree(false)
+		_, _, err := t.flushTree(0, false)
```
byo, line 1484: Why it is not:

```suggestion
		_, _, err := t.flushTree(t.cleanupPercentage, false)
```

?

embedded/tbtree/tbtree.go
```diff
@@ -1615,7 +1665,6 @@ func (n *innerNode) updateOnInsertAt(key []byte, value []byte, ts uint64) (n1 no
 	}
 
 	n._ts = ts
-	n.mut = true
```
byo, line 1618: Why we don't mark here that the node was mutated ?

embedded/tbtree/tbtree.go
```diff
@@ -1834,24 +1904,37 @@ func (n *innerNode) split() (node, error) {
...
 			n._ts = n.nodes[i].ts()
 		}
+
+		if minOff > n.nodes[i].minOffset() && !n.nodes[i].mutated() {
+			minOff = n.nodes[i].minOffset()
+		}
+	}
+
+	if minOff == int64(math.MaxInt64) {
+		n._minOff = 0
```
byo, line 1935: This is to cover the case when there are no child nodes?

Because otherwise, current node would have to be marked as mutated too and would be written by the next flush anyway.

embedded/appendable/multiapp/multi_app.go
```diff
@@ -410,6 +414,43 @@ func (mf *MultiFileAppendable) SetOffset(off int64) error {
...
+		return ErrAlreadyClosed
+	}
+
+	if mf.offset() < off {
+		return fmt.Errorf("%w: discard beyond existent data boundaries", ErrIllegalArguments)
+	}
+
+	appID := appendableID(off, mf.fileSize)
+
+	for i := int64(0); i < appID; i++ {
```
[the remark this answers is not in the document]
jeroiraz: the metadata is stored in the beginning of each single-appendable and the offset it's handled internally, so offset 0 means right after the metadata

embedded/tbtree/snapshot.go
```diff
@@ -307,54 +324,61 @@ func (n *innerNode) writeTo(nw, hw io.Writer, writeOpts *WriteOpts) (nOff int64,
...
-				_ts:     c.ts(),
-				off:     c.offset(),
-			}
-		}
+			for i, c := range n.nodes {
+				_, isNodeRef := c.(*nodeRef)
 
-		n.nodes = nodes
+				if isNodeRef {
+					n.nodes[i] = c
```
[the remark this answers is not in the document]
jeroiraz: yes, it's. it's a left over after removing the allocation of a new node list

embedded/tbtree/snapshot.go
```diff
@@ -307,54 +324,61 @@ func (n *innerNode) writeTo(nw, hw io.Writer, writeOpts *WriteOpts) (nOff int64,
...
+					continue
+				}
 
-		n.t.cachePut(n)
```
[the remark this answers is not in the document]
jeroiraz: it will be inserted by its parent or if not, it will be the root of the tree

embedded/tbtree/tbtree.go
jeroiraz: yes, there are synced before writing into the commit log. Sync could be right before cLog sync as well...

embedded/tbtree/tbtree.go
jeroiraz: the node is already mutated.

if none of its children is serialized, then minOff will be set to 0

embedded/tbtree/tbtree.go
jeroiraz: after compaction, current nLog will be physically deleted anyway

embedded/tbtree/tbtree.go
jeroiraz: because only mutated nodes are updated. A mutated node is created when trying to modify a non-mutated one (a copy is created in mutated state)

embedded/tbtree/snapshot.go
```diff
@@ -424,47 +448,51 @@ func (l *leafNode) writeTo(nw, hw io.Writer, writeOpts *WriteOpts) (nOff int64,
...
 	}
 
 	wN = int64(size)
 	nOff = writeOpts.BaseNLogOffset
 
 	if writeOpts.commitLog {
-		l.off = writeOpts.BaseNLogOffset
-		l.mut = false
+		l.off = nOff
+		l._minOff = nOff
```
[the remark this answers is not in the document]
jeroiraz: indeed, but it was used to hold the min referenced offset even when the node was mutated. Simplified it, so minOffset as well as Offset are only valid when the node is not mutated.
The current min referenced offset is hold at the tree level now

embedded/tbtree/tbtree.go
byo: OK, I thought we don't have the old-style compaction anymore.

embedded/tbtree/tbtree.go
byo: OK, didn't see previous calls to Sync - those are in different part of the code.

byo (dismissed):
LGTM

One thing I noticed though - when doing flush through immuadmin, cleanup percentage is picked up correctly. But even when I was changing the default option in the code, the flush that was automatically made by flush threshold was always made with percentage=0

PR #1150, opened 2022-03-18
Feat/constrained writes

byo:


embedded/store/immustore.go
```diff
@@ -875,6 +898,26 @@ func (s *ImmuStore) commit(otx *OngoingTx, expectedHeader *TxHeader, waitForInde
...
 
+	// first check is performed on the snapshot taken when opening the transaction,
+	// this will prevent acquiring the write lock and putting garbage into
+	// appendables if we can already determine that constraints are not met.
+	//
+	// A corner case is if the DB fails to meet constraints now but would fulfill
+	// those during final commit phase - but in such case, we are allowed to reason
+	// about the DB state in any point in time between both snapshots are checked.
+	if hasConstrainedKVEntries(otx.entries) {
+		snap, err := s.indexer.SnapshotSince(s.indexer.Ts())
```
jeroiraz, line 909: this constraint validation may be done directly over the tbtree. See unsafeIndex struct in this same file

byo: Fixed, thanks!

embedded/store/ongoing_tx.go
```diff
@@ -34,9 +34,10 @@ type OngoingTx struct {
 }
 
 type EntrySpec struct {
-	Key      []byte
-	Metadata *KVMetadata
-	Value    []byte
+	Key         []byte
+	Metadata    *KVMetadata
+	Value       []byte
+	Constraints *KVConstraints
```
jeroiraz, line 40: I think Constraints should be handled at OngoingTx. Because it will be posible to unify MVCC validations for RW-transactions quite easily later on

byo: Moved to OngoingTx

pkg/api/schema/schema.proto
```diff
@@ -94,10 +94,22 @@ message OpenSessionResponse {
...
+
+	// If set to value > 0, only store new value if the existing value does not exist or was not modified after given transaction
+	uint64 notModifiedAfterTX = 3;
+}
+
 message KeyValue {
 	bytes key = 1;
 	bytes value = 2;
 	KVMetadata metadata = 3;
+	KVConstraints constraints = 4;
```
jeroiraz, line 112: probably constraints could be at SetRequest... it'd be even possible to set a constraint over a key which is not inserted or updated but read... It may add a bit of complexity on the message types but on the other hand it would be equivalent to constraints in read-write transactions... wdyt?

byo: Yes, that's a very interesting idea, the only drawback I could see here is that we would have to send key values twice for constrained entries - if keys were long, that would mean a significant increase of transferred data. Should we worry about that?

jeroiraz: for the constraints it may be possible to use the hash of the key...

embedded/store/immustore.go
```diff
@@ -71,6 +72,13 @@ var ErrUnexpectedError = errors.New("unexpected error")
 var ErrUnsupportedTxVersion = errors.New("unsupported tx version")
 var ErrNewerVersionOrCorruptedData = errors.New("tx created with a newer version or data is corrupted")
 
+var ErrInvalidConstraints = errors.New("invalid constraints")
+var ErrInvalidConstraintsTooMany = fmt.Errorf("%w: too many constraints", ErrInvalidConstraints)
+var ErrInvalidConstraintsNull = fmt.Errorf("%w: null constraint", ErrInvalidConstraints)
+var ErrInvalidConstraintsNullKey = fmt.Errorf("%w: %s", ErrInvalidConstraints, ErrNullKey)
+var ErrInvalidConstraintsMaxKeyLenExceeded = fmt.Errorf("%w: %s", ErrInvalidConstraints, ErrorMaxKeyLenExceeded)
+var ErrInvalidConstraintsDuplicateKey = fmt.Errorf("%w: %s", ErrInvalidConstraints, ErrDuplicatedKey)
```
jeroiraz, line 80: ErrInvalidConstraintsDuplicatedKey

embedded/store/kvconstraints.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/codenotary/immudb/embedded/tbtree"
+)
+
+type KVConstraints struct {
+	Key                []byte
```
jeroiraz, line 26: wouldn't expect the key to belong to a constraint datatype...

jeroiraz: it'd be great to have a constraint interface... so it may be extensible in the future i.e. mustExist, mustNoExist would be concrete implementations of such interface

embedded/store/ongoing_tx.go
```diff
@@ -182,6 +184,27 @@ func (tx *OngoingTx) Set(key []byte, md *KVMetadata, value []byte) error {
...
+
+	if c == nil {
+		return ErrIllegalArguments
+	}
+
+	if len(c.Key) == 0 {
+		return ErrNullKey
+	}
+
+	if len(c.Key) > tx.st.maxKeyLen {
```
jeroiraz, line 200: this is showing up that the validation of the constraint datatype is not be self-contained and probably it's better for readability to have input parameters as (key, constraint)...

embedded/store/kvconstraints.go
byo: As discussed offline, I think each constraint should be a functor that takes the current db state and responds with true or false depending on whether the constraint is met or not. Such function could be parametrized, e.g. MustExist is parametrised with the key, same with MustNotExist, for NotModifiedAfterTX it would be parametrised with key and transaction id.

Let's maybe first focus on the GRPC protocol since that's something hard to change later.

@jeroiraz what do you think about this approach?:

```proto
message WriteConstraint {

	message KeyMustExist {
		bytes key = 1;
	}
	
	message KeyMustNotExist {
		bytes key = 1;
	}
	
	message KeyNotModifiedAfterTX {
		bytes key = 1;
		uint64 txID = 2;
	}

	oneof constraint {
		KeyMustExist keyMustExist = 1;
		KeyMustNotExist keyMustNotExist = 2;
		KeyNotModifiedAfterTX keyNotModifiedAfter = 3;
	}
}
```
and then in messages:
```proto
message SetRequest {
	repeated KeyValue KVs = 1;
	bool  noWait = 2;
	repeated WriteConstraint constraints = 3;
}
```

Internally we could represent this as an implementation of abstract interface so that it's easy to extend with more constraints such as those based on prefix (e.g. there must be no key with given prefix).

embedded/store/ongoing_tx.go
```diff
@@ -182,6 +185,24 @@ func (tx *OngoingTx) Set(key []byte, md *KVMetadata, value []byte) error {
 	return nil
 }
 
+func (tx *OngoingTx) AddKVConstraint(c WriteConstraint) error {
```
jeroiraz, line 188: probably it's better to have a single naming for this constraints both in method and type names

embedded/store/ongoing_tx_test.go
```diff
@@ -0,0 +1,74 @@
...
+		st: &ImmuStore{
+			maxKeyLen: 10,
+		},
+	}
+
+	err := otx.AddKVConstraint(nil)
+	require.ErrorIs(t, err, ErrIllegalArguments)
+
+	err = otx.AddKVConstraint(&WriteContraintKeyMustExist{})
+	require.ErrorIs(t, err, ErrInvalidConstraints)
```
jeroiraz, line 35: ErrInvalidConstraint ?

embedded/store/ongoing_tx.go
byo: Changed to Precondition everywhere

jeroiraz (approved):

PR #1151, opened 2022-03-18
Feat/proof 1 1 compatibility mode

byo:


pkg/api/schema/schema.proto
```diff
@@ -416,6 +418,72 @@ message IndexSettings {
...
+message ConditionalUint64 {
+	uint64 value = 1;
+}
+
+message ConditionalBool {
+	bool value = 1;
+}
+
+message ReplicationSettings {
+	bool   replica = 1;
```
jeroiraz, line 437: conditional settings may be used here as well

pkg/server/db_options.go
```diff
@@ -392,3 +465,35 @@ func (s *ImmuServer) loadDBOptions(database string, createIfNotExists bool) (*db
 
 	return options, nil
 }
+
+func (s *ImmuServer) logDBOptions(database string, opts *dbOptions) {
```
jeroiraz, line 469: wouldn't this produce too much logging lines?

jeroiraz (approved):

PR #1154, opened 2022-03-23
Extend tx-related endpoints with entries resolution and filtering

jeroiraz:
This PR extends TxByID, VerifiedTxByID and TxScan endpoints with the capability of filtering and resolving entries.

```go
// return tx header together with resolved kv entries in a single grpc request
tx, err := client.TxByIDWithSpec(context.Background(),
			&schema.TxRequest{
				Tx: hdr.Id,
				EntriesSpec: &schema.EntriesSpec{
					KvEntriesSpec: &schema.EntryTypeSpec{
						Action: schema.EntryTypeAction_RESOLVE,
					},
				},
			})
```


mmeloni:
I would add also txs batch retrivial support, with pagination.
It will be extremely useful when search cache need to start from scratch and also in hot backup.

byo:
> I would add also txs batch retrivial support, with pagination. It will be extremely useful when search cache need to start from scratch and also in hot backup.

jeroiraz:
> As discussed offline:
> 
> #### If no spec is present for given kind of entries, the default should be to exclude at all (both from the parsed field and the raw data in the `entries` field).
> This is necessary to avoid serious issues with backwards compatibility in the future, here's the scenario:
> 
> * immudb client uses this API call and wants to parse raw entries for KV entries but disables zset entries by using EXCLUDE set
> * if such call is made, the `entries` field will only contain raw entries for KV ones
> * in the future if we add new type of entry (like graph for example) - this means that due to lack of its spec, those entries will also be added to the `entries` list - and this would break user's assumption that only KV entries are returned
> 
> This leaves a question on how to get backward compatibility right - currently the new access method is using the same API endpoint as before meaning that with no specs at all it should return all raw entries and nothing more.

Current implementation should work as described

jeroiraz:
> I would add also txs batch retrivial support, with pagination. It will be extremely useful when search cache need to start from scratch and also in hot backup.

sure,  stream endpoint is also useful, we can implement a new endpoint

pkg/api/schema/schema.proto
```diff
@@ -349,24 +352,54 @@ message VerifiableZAddRequest {
...
+}
+
+message EntryTypeSpec {
+	EntryTypeAction action = 1;
+}
+
+enum EntryTypeAction {
+	EXCLUDE     = 0;
+	ONLY_DIGEST = 1;
+	RAW_VALUE   = 2;
```
byo, line 374: how about only the referenced key info? like I want to get parsed reference but not query for the referenced value (this of course could be a future improvement)

byo (approved):
Did a brief check and it looks good, however I suggest a second review too.

pkg/api/schema/schema.proto
jeroiraz: yes, we can add it later on

pkg/database/database_test.go
```diff
@@ -510,14 +511,264 @@ func TestTxByID(t *testing.T) {
...
+	t.Run("values should not be resolved but digests returned in entries field", func(t *testing.T) {
+		tx, err := db.TxByID(&schema.TxRequest{Tx: txhdr1.Id})
+		require.NoError(t, err)
+		require.NotNil(t, tx)
+		require.Len(t, tx.Entries, 1)
+		require.Empty(t, tx.KvEntries)
+		require.Empty(t, tx.ZEntries)
+
+		for _, e := range tx.Entries {
+			require.Len(t, e.Value, 0)
```
byo, line 570: We could also check here (and simillar checks below) if digest is not empty (or it's of the sha256 size):

```suggestion
			require.Len(t, e.Value, 0)
			require.Len(t, e.HValue, sha256.Size)
```

could even check if the sha256 is correct

pkg/integration/tx_entries_test.go
```diff
@@ -0,0 +1,227 @@
...
+	"github.com/codenotary/immudb/pkg/api/schema"
+	ic "github.com/codenotary/immudb/pkg/client"
+	"github.com/codenotary/immudb/pkg/server"
+	"github.com/codenotary/immudb/pkg/server/servertest"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+)
+
+func Test_GetTransactionEntries(t *testing.T) {
+	options := server.DefaultOptions()
```
byo, line 34: Minor: Maybe use some temporary folder name? Using `data` sounds dangerous here.

byo (commented):
One comment to resolve regarding backwards compatibility, few other minor comments.

Also make sure that we have a task to document new changes.

mmeloni (approved):

PR #1156, opened 2022-03-28
flushing cleanup percentage as float value

jeroiraz:


embedded/tbtree/tbtree.go
```diff
@@ -1021,7 +1021,7 @@ func (t *TBtree) flushTree(cleanupPercentage int, synced bool) (wN int64, wH int
 	)
 	defer finishOutputFunc()
 
-	expectedNewMinOffset := t.minOffset + ((t.committedNLogSize-t.minOffset)*int64(cleanupPercentage))/100
+	expectedNewMinOffset := t.minOffset + int64((float32(t.committedNLogSize-t.minOffset)*cleanupPercentage)/100)
```
byo, line 1024: casting to `float32` may not be of enough precision here, how about `float64` ?

PR #1157, opened 2022-04-01
Feat db runtime

jeroiraz:
This PR introduces support for database loading and unloading without the need of restarting immudb server.

A new attribute `autoload` as added into database settings, instructing if the database is automatically loaded when the server starts.

Endpoints for loading and unloading databases is also added.

```
./immuadmin database list                            
4 database(s)
-  -------------  -----     
#  Database Name  Status  
-  -------------  -----     
1  defaultdb      LOADED   
2  db1            UNLOADED  
3  db2            LOADED  
4  db3            LOADED  
-  -------------  -----

immudb  2022/03/31 10:27:41 INFO: SQL Engine ready for database 'defaultdb' {replica = false}
immudb  2022/03/31 10:27:41 INFO: Database 'db1' is closed (autoload is disabled)
```

New endpoints (LoadDatabase, UnloadDatabase) for loading and unloading databases have been added:
```
./immuadmin database unload db1
database 'db1' successfully unloaded

./immuadmin database load db1
database 'db1' successfully loaded
```


cmd/immuadmin/command/database.go
```diff
@@ -102,9 +112,49 @@ func (cl *commandline) database(cmd *cobra.Command) {
...
+	addDbUpdateFlags(createCmd)
+
+	openCmd := &cobra.Command{
+		Use:               "open",
+		Short:             "Open database",
+		Example:           "open {database_name}",
+		PersistentPreRunE: cl.ConfigChain(cl.connect),
+		PersistentPostRun: cl.disconnect,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			_, err := cl.immuClient.OpenDatabase(cl.context, &schema.OpenDatabaseRequest{
```
byo, line 124: I think `open` / `close` naming could be misleading here - like in this context - I would assume that `OpenDatabase` is about connecting to given database.

Maybe load / unload ?

pkg/server/db_options.go
```diff
@@ -59,12 +59,26 @@ type dbOptions struct {
...
 	UpdatedBy string    `json:"updatedBy"`
 	UpdatedAt time.Time `json:"updatedAt"`
 }
 
+type featureState int
+
+const (
+	unspecifiedState featureState = iota
+	enabledState
+	disabledState
```
byo, line 75: Since this i persisted I think it's better to set explicit values here to avoid conincidental reordering / modifications of those values:

```suggestion
	unspecifiedState featureState = 0
	enabledState featureState = 1
	disabledState featureState = 2
```

pkg/server/server.go
```diff
@@ -831,6 +819,103 @@ func (s *ImmuServer) CreateDatabaseV2(ctx context.Context, req *schema.CreateDat
...
+
+	dbOpts, err := s.loadDBOptions(req.Database, false)
+	if err == store.ErrKeyNotFound {
+		return nil, database.ErrDatabaseNotExists
+	}
+	if err != nil {
+		return nil, fmt.Errorf("%w: while loading database settings", err)
+	}
+
+	db, err = database.OpenDB(req.Database, s.databaseOptionsFrom(dbOpts), s.Logger)
```
byo, line 863: We can end up in a race condition here - multiple callers could try to simultaneously open a database here. Between the check for `IsClosed` and OpenDB there's no lock.

cmd/immuadmin/command/database.go
```diff
@@ -37,39 +37,49 @@ func addDbUpdateFlags(c *cobra.Command) {
...
-		Short:   "Issue all database commands",
-		Aliases: []string{"d"},
-		//PersistentPreRunE: cl.ConfigChain(cl.connect),
+	dbCmd := &cobra.Command{
+		Use:               "database",
+		Short:             "Issue all database commands",
+		Aliases:           []string{"d"},
 		PersistentPostRun: cl.disconnect,
-		ValidArgs:         []string{"list", "create", "update", "use", "clean"},
+		ValidArgs:         []string{"list", "create", "open", "close", "update", "use", "flush", "compact"},
```
byo, line 49:
```suggestion
		ValidArgs:         []string{"list", "create", "load", "unload", "update", "use", "flush", "compact"},
```

pkg/api/schema/schema.proto
```diff
@@ -611,10 +631,23 @@ message SetActiveUserRequest {
...
+}
+
+message DatabaseListResponseV2 {
+	repeated DatabaseWithSettings databases = 1;
+}
+
+message DatabaseWithSettings {
+	string name = 1;
+	DatabaseNullableSettings settings = 2;
+	bool open = 3;
```
byo, line 648:
```suggestion
	bool loaded = 3;
```

pkg/api/schema/schema.proto
```diff
@@ -983,13 +1016,34 @@ service ImmuService {
...
+	}
+
+	rpc UnloadDatabase(UnloadDatabaseRequest) returns (UnloadDatabaseResponse) {
+		option (google.api.http) = {
+			post: "/db/unload"
+			body: "*"
+		};
+	}
+
+	rpc DatabaseList (google.protobuf.Empty) returns (DatabaseListResponse){
```
byo, line 1033: Please add deprecation comment

pkg/server/db_options.go
```diff
@@ -59,12 +59,26 @@ type dbOptions struct {
...
+const (
+	unspecifiedState featureState = 0
+	enabledState     featureState = 1
+	disabledState    featureState = 2
+)
+
+func (fs featureState) isEnabled() bool {
+	return fs == unspecifiedState || fs == enabledState
+}
+
```
byo, line 81: Maybe add some function to set the state so that concrete values are hidden?
```suggestion

func (fs *featureState) setEnabled(enabled bool) {
    if enabled {
        *fs = enabledState
    } else {
        *fs = disabledState
    }
}
```

PR #1160, opened 2022-04-01
chore(pkg/api): uniform v2 endpoints

jeroiraz:


byo (approved):
LGTM,

Only one thing I was wondering about - `Nullable` is more about the fact that some parts of the GRPC protocol can be null values, from the user perspective it's more about being `Optional` - maybe that name would be better?

PR #1163, opened 2022-04-05
fix(pkg/client): Prevent updates with incorrect database settings

byo:

### DIFF
--- a/embedded/store/immustore.go
+++ b/embedded/store/immustore.go
@@ -70,6 +70,7 @@ var ErrOffsetOutOfRange = tbtree.ErrOffsetOutOfRange
 var ErrUnexpectedError = errors.New("unexpected error")
 var ErrUnsupportedTxVersion = errors.New("unsupported tx version")
 var ErrNewerVersionOrCorruptedData = errors.New("tx created with a newer version or data is corrupted")
+var ErrInvalidOptions = errors.New("invalid options")
 
 var ErrSourceTxNewerThanTargetTx = errors.New("source tx is newer than target tx")
 var ErrLinearProofMaxLenExceeded = errors.New("max linear proof length limit exceeded")
@@ -173,8 +174,9 @@ type refVLog struct {
 }
 
 func Open(path string, opts *Options) (*ImmuStore, error) {
-	if !validOptions(opts) {
-		return nil, ErrIllegalArguments
+	err := opts.Validate()
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", ErrIllegalArguments, err)
 	}
 
 	finfo, err := os.Stat(path)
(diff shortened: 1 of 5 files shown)


embedded/store/immustore.go
```diff
@@ -173,8 +174,9 @@ type refVLog struct {
 }
 
 func Open(path string, opts *Options) (*ImmuStore, error) {
-	if !validOptions(opts) {
-		return nil, ErrIllegalArguments
+	err := opts.Validate()
```
jeroiraz, line 177: originally it was catching up the case when opts was nil, now it will panic...

byo: It will still catch this case, `Validate` is doing this check too.

jeroiraz (dismissed):

byo:
@jeroiraz I heavily modified server_tests.go file - after my changes it started showing issues with tests. I think I've found the reason of sporadic errors with `address already in use` errors here too. Please re-review.

embedded/store/options.go
```diff
@@ -159,49 +160,125 @@ func DefaultIndexOptions() *IndexOptions {
...
+	}
+	if opts.WriteTxHeaderVersion > MaxTxHeaderVersion {
+		return fmt.Errorf("%w: invalid WriteTxHeaderVersion", ErrInvalidOptions)
+	}
+
+	// options below are only set during initialization and stored as metadata
+	if opts.MaxTxEntries <= 0 {
+		return fmt.Errorf("%w: invalid MaxTxEntries", ErrInvalidOptions)
+	}
+	if opts.MaxKeyLen <= 0 {
```
jeroiraz, line 215: min and max values may be unified in a single validation

embedded/store/options.go
```diff
@@ -159,49 +160,125 @@ func DefaultIndexOptions() *IndexOptions {
...
 
-		opts.VLogMaxOpenedFiles > 0 &&
-		opts.TxLogMaxOpenedFiles > 0 &&
-		opts.CommitLogMaxOpenedFiles > 0 &&
+	if opts.MaxConcurrency <= 0 {
+		return fmt.Errorf("%w: invalid MaxConcurrency", ErrInvalidOptions)
+	}
 
-		opts.TxLogCacheSize >= 0 &&
+	if opts.MaxIOConcurrency <= 0 {
```
jeroiraz, line 172: min and max values may be unified in a single validation

embedded/store/options.go
```diff
@@ -159,49 +160,125 @@ func DefaultIndexOptions() *IndexOptions {
...
-		opts.MaxNodeSize > 0 &&
-		opts.RenewSnapRootAfter >= 0 &&
-		opts.NodesLogMaxOpenedFiles > 0 &&
-		opts.HistoryLogMaxOpenedFiles > 0 &&
-		opts.CommitLogMaxOpenedFiles > 0
+	if opts.TimeFunc == nil {
+		return fmt.Errorf("%w: invalid TimeFunc", ErrInvalidOptions)
+	}
+
+	if opts.WriteTxHeaderVersion < 0 {
```
jeroiraz, line 204: min and max values may be unified in a single validation

PR #1164, opened 2022-04-06
Feat load/unload/delete database

jeroiraz:
This PR introduces support for database loading and unloading without the need of restarting immudb server.

A new attribute autoload as added into database settings, instructing if the database is automatically loaded when the server starts.

Endpoints for loading and unloading databases is also added.

```
./immuadmin database list                            
4 database(s)
-  -------------  -----     
#  Database Name  Status  
-  -------------  -----     
1  defaultdb      LOADED   
2  db1            UNLOADED  
3  db2            LOADED  
4  db3            LOADED  
-  -------------  -----

immudb  2022/03/31 10:27:41 INFO: SQL Engine ready for database 'defaultdb' {replica = false}
immudb  2022/03/31 10:27:41 INFO: Database 'db1' is closed (autoload is disabled)
```

New endpoints (LoadDatabase, UnloadDatabase) for loading and unloading databases have been added:

```
./immuadmin database unload db1
database 'db1' successfully unloaded

./immuadmin database load db1
database 'db1' successfully loaded
```

This PR also includes a new endpoint for physical database deletion `DeleteDatabase` (a delete command was added into immuadmin tool)


cmd/immuadmin/command/database.go
```diff
@@ -37,39 +37,49 @@ func addDbUpdateFlags(c *cobra.Command) {
...
-		Use:     "database",
-		Short:   "Issue all database commands",
-		Aliases: []string{"d"},
-		//PersistentPreRunE: cl.ConfigChain(cl.connect),
+	dbCmd := &cobra.Command{
+		Use:               "database",
+		Short:             "Issue all database commands",
+		Aliases:           []string{"d"},
 		PersistentPostRun: cl.disconnect,
-		ValidArgs:         []string{"list", "create", "update", "use", "clean"},
```
byo, line 49: We're changing clean to compact - maybe we should support both action names for backward compatibility?

There could be some admin scripts running compaction periodically already out there.

cmd/immuadmin/command/database.go
```diff
@@ -102,9 +112,69 @@ func (cl *commandline) database(cmd *cobra.Command) {
...
+	}
 
-	cu := &cobra.Command{
+	deleteCmd := &cobra.Command{
+		Use:               "delete",
+		Short:             "Delete database",
+		Example:           "delete {database_name}",
+		PersistentPreRunE: cl.ConfigChain(cl.connect),
+		PersistentPostRun: cl.disconnect,
+		RunE: func(cmd *cobra.Command, args []string) error {
```
byo, line 163: This is a destructive operation. Maybe we should do some simple safety action here?

Like require an extra argument '--yes-i-know-what-i-am-doing`?

pkg/client/session.go
```diff
@@ -36,9 +38,7 @@ func (c *immuClient) OpenSession(ctx context.Context, user []byte, pass []byte,
 		return err
 	}
 
-	if c.ServiceClient == nil {
-		c.ServiceClient = schema.NewImmuServiceClient(c.clientConn)
-	}
+	c.ServiceClient = schema.NewImmuServiceClient(c.clientConn)
```
byo, line 41: Why this change is needed? There's no way to set custom ServiceClient that way

byo (commented):
Did a brief check, one comment regarding immuadmin command to delete the database.

Besides that it looks good, I'd still like to do some more tests, but maybe it would be better to merge into master first.

cmd/immuadmin/command/database.go
jeroiraz: actually the command was changes couple of releases ago, this was a left over...

pkg/client/session.go
jeroiraz: hm, the client API should be revised... while the ServiceClient could be externally set, a new client connection was stablished anyways, and assumed to be there in other parts of the code i.e. when closing the session

PR #1167, opened 2022-04-09
fix(pkg/server): typo in log message

bengentil:
Signed-off-by: Benjamin Gentil <benjamin@gentil.io>


jeroiraz:
thanks @bengentil !

PR #1168, opened 2022-04-11
chore(embedded/store): use store layer for constraint validations

jeroiraz:

### DIFF
--- a/embedded/store/immustore_test.go
+++ b/embedded/store/immustore_test.go
@@ -2349,7 +2349,9 @@ func TestImmudbStoreCommitWithPreconditions(t *testing.T) {
 	require.NoError(t, err)
 
 	// wait for entry to be expired
-	for {
+	for i := 0; ; i++ {
+		require.Less(t, i, 20, "entry expiration failed")
+
 		time.Sleep(100 * time.Millisecond)
 
 		_, err = immuStore.Get([]byte("expirableKey"))
--- a/embedded/store/immustore_test.go
+++ b/embedded/store/immustore_test.go
@@ -1194,10 +1194,19 @@ func TestImmudbStoreRWTransactions(t *testing.T) {
 		// already expired
 		_, err = immuStore.Get([]byte("expirableKey"))
 		require.ErrorIs(t, err, ErrKeyNotFound)
+		require.ErrorIs(t, err, ErrExpiredEntry)
 
-		// expired entries are never returned
-		_, err = immuStore.GetWith([]byte("expirableKey"))
+		// expired entries can not be resolved
+		valRef, err = immuStore.GetWith([]byte("expirableKey"))
+		require.NoError(t, err)
+		_, err = valRef.Resolve()
 		require.ErrorIs(t, err, ErrKeyNotFound)
+		require.ErrorIs(t, err, ErrExpiredEntry)
+
+		// expired entries are not returned
+		_, err = immuStore.GetWith([]byte("expirableKey"), IgnoreExpired)
+		require.ErrorIs(t, err, ErrKeyNotFound)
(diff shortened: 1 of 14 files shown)


embedded/store/preconditions.go
```diff
@@ -108,8 +108,8 @@ func (cs *PreconditionKeyNotModifiedAfterTx) Validate(st *ImmuStore) error {
 	return nil
 }
 
-func (cs *PreconditionKeyNotModifiedAfterTx) Check(idx *indexer) (bool, error) {
-	_, tx, _, err := idx.Get(cs.Key)
+func (cs *PreconditionKeyNotModifiedAfterTx) Check(idx KeyIndex) (bool, error) {
+	valRef, err := idx.Get(cs.Key)
```
byo, line 112: Deleted entries are considered not modified after given TX - I'm not sure if this is a good approach.

Deletion of an entry could result in valid constraint check that leads to removal of some intermediate changes made to the key.

e.g.:

* Key has value X set at tx 1, two writers read that value
* Writer 1 adds update to the key in tx 2 (using precondition not modified since tx 1)
* Someone deletes the key in tx 3
* Writer 2 updates the key with precondition not modified since tx 1 - this succeeds

Shouldn't writer 2 receive an error that the constraint has failed?

jeroiraz: aham, probably the constraint should support the spec if deleted ones are ignored or not, it would be consistent with the optional filtering when reading and scanning over keys

byo: yup, agree, however I think that the default behavior would be to consider deletion as entry modification. That way you can still do atomic operations on the database that also consider deletion as one of operations.

This also has strange corner cases when we consider data expiration. I think that by default it should not matter if the entry expired or not. The important fact would be if it was set as expirable after given TX.

In case of non-indexable entries I don't think we can support such case.

embedded/store/immustore_test.go
```diff
@@ -2239,6 +2251,158 @@ func (la *FailingAppendable) Append(bs []byte) (off int64, n int, err error) {
...
+	require.NoError(t, err)
+
+	err = otx.Set([]byte("expirableKey"), md, []byte("expirableValue"))
+	require.NoError(t, err)
+
+	hdr, err := otx.Commit()
+	require.NoError(t, err)
+
+	// wait for entry to be expired
+	for {
```
byo, line 2352:
```suggestion
	for i:=0;; i++ {
        require.Less(i, 20, "Entry expiration failed")
```

embedded/store/key_reader.go
```diff
@@ -33,25 +34,31 @@ type Snapshot struct {
 
 type valueRefInterceptor func(key []byte, valRef ValueRef) ValueRef
 
-// filter out entries when filter evaluates to true
-type FilterFn func(valRef ValueRef, t time.Time) bool
+// filter out entries when filter evaluates to a non-nil error
+type FilterFn func(valRef ValueRef, t time.Time) error
```
byo, line 38: For future: Currently it's not needed but shouldn't we plan to also support errors when filtering happens? i.e. if we check something on disk and read fails. That would return an error too, but it would not indicate that the entry should be filtered our but rather that there was an error determining if it should be filtered out.

This would be important in line 366: https://github.com/codenotary/immudb/pull/1168/files#diff-c632c3c2ae2f4bdfbf8378e41bdcfecc38a5a87142a5a3ba769d721e4a0eb4cbR366 - where getting any error from filter means the entry is skipped where we should skip it only if the error indicates that the entry should be skipped.

byo (approved):
Two minor comments, one for a bit improved test, one for future considerations. Those are not blockers for this PR

PR #1181, opened 2022-04-15
chore: Use specific version of git-chglog

byo:
Use specific version of git-chglog run in docker
to consistently generate the CHANGELOG.md file


byo:
@jeroiraz After this change, running `make CHANGELOG.md` should not produce any changes in the `CHANGELOG.md` file - can you check that ?

byo:
Even with unified docker environment, there are still differences between my and @jeroiraz's outputs - changelog should then be generated in a more controlled manner in Github pipelines.

PR #1182, opened 2022-04-15
fix(pkg/auth): Do not send duplicated authorization header

byo:

### DIFF
--- a/pkg/auth/clientinterceptors.go
+++ b/pkg/auth/clientinterceptors.go
@@ -19,17 +19,14 @@ package auth
 import (
 	"context"
 
-	"google.golang.org/grpc/metadata"
-
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
 )
 
 // ClientStreamInterceptor gRPC client interceptor for streams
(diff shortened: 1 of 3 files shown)


pkg/auth/clientinterceptors.go
```diff
@@ -19,50 +19,36 @@ package auth
...
-type TokenAuthStruct struct {
-	Token string
-}
-
-// GetRequestMetadata callback which returns the Bearer token to be set in request metadata
-func (t TokenAuthStruct) GetRequestMetadata(ctx context.Context, in ...string) (map[string]string, error) {
-	var token string
+func updateAuthHeader(ctx context.Context, token string) context.Context {
 	if md, ok := metadata.FromOutgoingContext(ctx); ok && len(md.Get("authorization")) > 0 {
+		// Token from existing metadata overwrites the one stored in
```
jeroiraz, line 44: existent token shouldn't be replaced by the one provided as parameter then? if so, I'd suggest changing the comment from `stored in` by `provided in current call` or something like that. And the method could do a fast return if no change is made into the metadata.

However, I would expect the provided as parameter having more precedence actually...

byo: Clarified comments a bit.

This is a very fragile peace of code. Most of the work was to find a way to update it without breaking tests and compatibility with existing imuclient. We should definitely clean up the code setting up interceptors when thinking about the new version of client libraries.

jeroiraz (approved):

PR #1188, opened 2022-04-19
chore(github): Update github workflow on master / version push

byo:
* go Use 1.17 compiler to build binaries
* check if binaries can be built with go 1.13
* build binaries with `make dist` and notarize all binaries
  compiled that way

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,5 +1,9 @@
 name: pushCI
 
+env:
+  GO_VERSION: 1.17
+  MIN_SUPPORTED_GO_VERSION: 1.13
+
 on:
   push:
     branches:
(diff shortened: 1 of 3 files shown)


.github/workflows/push.yml
```diff
@@ -8,6 +12,17 @@ on:
       - 'v*'
 
 jobs:
+
+  old-go:
+    name: Ensure immudb compiles with the oldest supported go version
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-go@v1
```
tomekkolo, line 20: most probably does not make any difference, but there is already v3 https://github.com/actions/setup-go

tomekkolo (approved):

tomekkolo (approved):

PR #1189, opened 2022-04-19
fix(embedded/tbtree): split into multiple nodes

jeroiraz:


jeroiraz:
TODO: fix max key and value lengths at creation time (store into metadata section). It may require index regeneration

embedded/tbtree/tbtree.go
```diff
@@ -1651,60 +1664,61 @@ func (t *TBtree) snapshotClosed(snapshot *Snapshot) error {
...
-	nodes[insertAt+1] = c2
+	for i, c := range cs {
+		ns[insertAt+i] = c
+	}
 
-	if insertAt+2 < len(nodes) {
-		copy(nodes[insertAt+2:], n.nodes[insertAt+1:])
+	if insertAt+len(cs) < len(ns) {
+		copy(ns[insertAt+len(cs):], n.nodes[insertAt+1:])
 	}
```
byo, line 1702: This whole could be written as:
```go
copy(ns, n.nodes[:insertAt])
copy(ns[insertAt:], cs)
copy(ns[insertAt+len(cs)], n.nodes[insertAt+1:])
```

embedded/tbtree/tbtree.go
```diff
@@ -1734,16 +1748,17 @@ func (n *innerNode) copyOnInsertAt(key []byte, value []byte, ts uint64) (n1 node
...
-	newNode.nodes[insertAt+1] = c2
+	for i, c := range cs {
+		newNode.nodes[insertAt+i] = c
+	}
 
-	if insertAt+2 < len(newNode.nodes) {
-		copy(newNode.nodes[insertAt+2:], n.nodes[insertAt+1:])
+	if insertAt+len(cs) < len(newNode.nodes) {
+		copy(newNode.nodes[insertAt+len(cs):], n.nodes[insertAt+1:])
 	}
```
byo, line 1757: Same here:

```go
copy(newNode.nodes, n.nodes[:insertAt])
copy(newNode.nodes[insertAt:], cs)
copy(newNode.nodes[insertAt+len(cs):], n.nodes[insertAt+1:])
```

embedded/tbtree/tbtree.go
```diff
@@ -1907,17 +1922,22 @@ func (n *innerNode) split() (node, error) {
...
-
 	newNode.updateTs()
 
 	n.nodes = n.nodes[:splitIndex]
-
 	n.updateTs()
 
-	metricsBtreeInnerNodeEntries.WithLabelValues(n.t.path).Observe(float64(len(n.nodes)))
-	metricsBtreeInnerNodeEntries.WithLabelValues(n.t.path).Observe(float64(len(newNode.nodes)))
```
byo, line 1918: Some metrics were removed here

embedded/tbtree/tbtree.go
```diff
@@ -2385,13 +2405,19 @@ func (l *leafNode) split() (node, error) {
 	newLeaf.updateTs()
 
 	l.values = l.values[:splitIndex]
-
 	l.updateTs()
 
-	metricsBtreeLeafNodeEntries.WithLabelValues(l.t.path).Observe(float64(len(l.values)))
-	metricsBtreeLeafNodeEntries.WithLabelValues(l.t.path).Observe(float64(len(newLeaf.values)))
+	ns1, err := l.split()
```
byo, line 2410: How can we be sure that this and the `newLeaf.split()` below won't end up with an infinite recursive split?

I'm also worried about this ending up with very uneven split - like if the first half must be split again, it will end up with 3 children, first two containing ~1/4 entries, the last one containing 1/2.

embedded/tbtree/tbtree.go
jeroiraz: split is a recursive function and the base case should cover this metrics... could you double check it?

embedded/tbtree/tbtree.go
jeroiraz: termination is guaranteed by ensuring at least two entries can fit into a inner node

embedded/tbtree/tbtree.go
```diff
@@ -498,6 +509,20 @@ func OpenWith(path string, nLog, hLog, cLog appendable.Appendable, opts *Options
...
+		maxKeySize = opts.maxKeySize
+	}
+
+	maxValueSize, ok := metadata.GetInt(MetaMaxValueSize)
+	if !ok {
+		maxValueSize = opts.maxValueSize
+	}
+
+	if maxNodeSize < minNodeSize(maxKeySize, maxValueSize) {
+		return nil, fmt.Errorf("%w: max node size is too small for specified max key and max value sizes", ErrIllegalArguments)
```
byo, line 523: We should add some extra log info here with information on where to find a solution in such case

embedded/tbtree/tbtree.go
```diff
@@ -498,6 +509,20 @@ func OpenWith(path string, nLog, hLog, cLog appendable.Appendable, opts *Options
...
+	if !ok {
+		maxKeySize = opts.maxKeySize
+	}
+
+	maxValueSize, ok := metadata.GetInt(MetaMaxValueSize)
+	if !ok {
+		maxValueSize = opts.maxValueSize
+	}
+
+	if maxNodeSize < minNodeSize(maxKeySize, maxValueSize) {
```
byo, line 522: Maybe rename `minNodeSize` to `requiredMinNodeSize`?

embedded/tbtree/tbtree.go
```diff
@@ -646,13 +672,20 @@ func greatestKeyOfSize(size int) []byte {
 	return k
 }
 
+func minNodeSize(maxKeySize, maxValueSize int) int {
+	// space for at least two children is required for inner nodes
+	// 31 bytes are fixed in leafNode serialization while 29 bytes are fixed in innerNodes
```
byo, line 677: Is there a way to calculate this 31 instead of hardcoding that value or to put it closer to the serialization code?

It may lead to very unpleasant issues if we forget to update it when the btree structure changes again.

embedded/tbtree/tbtree.go
byo: Yup, you're right, if we do recursive split for new two nodes, node sizes will be observed

embedded/tbtree/tbtree.go
byo: What about the uneven distribution of child nodes in case any of new nodes needs a recursive split? Maybe better to base the split position based on bytes rather than entries?

embedded/tbtree/tbtree.go
jeroiraz: we should describe in the documentation how to address this error. but it should not be associated to a specific documentation site

embedded/tbtree/tbtree.go
jeroiraz: I think it will be better to handle the changes with care. I added a direct reference into the size calculation functions for lead and inner nodes describing that this function needs to be revised if any change is made

embedded/tbtree/tbtree.go
byo: As discussed offline - those recursive splits will be very rare and will only happen in a very odd key length distributions (such as majority of small keys and few huge ones). It should be considered a safely recovery (and thus an unusual situation) rather than a common, optimizable scenario.

byo (approved):

PR #1191, opened 2022-04-22
Fix/cleanup percentage

byo:


jeroiraz (approved):
internal flush-related methods may be simplified and improved in the future

PR #1192, opened 2022-04-22
chore(pkg/api): txbyid with keepReferencesUnresolved option

jeroiraz:
`TxById` and `VerifiableTxById` can be resolved without requiring the index when  `KeepReferencesUnresolved` is set to `true` in the request.

e.g.
```go
tx, err := client.TxByIDWithSpec(context.Background(),
    &schema.TxRequest{
        Tx: hdr.Id,
        EntriesSpec: &schema.EntriesSpec{
            KvEntriesSpec: &schema.EntryTypeSpec{
                Action: schema.EntryTypeAction_RESOLVE,
            },
            ZEntriesSpec: &schema.EntryTypeSpec{
                Action: schema.EntryTypeAction_RESOLVE,
            },
        },
        KeepReferencesUnresolved: true,
})
```


pkg/api/schema/schema.proto
```diff
@@ -377,10 +377,11 @@ message VerifiableZAddRequest {
...
 message TxRequest {
-	uint64      tx = 1;
-	EntriesSpec entriesSpec = 2;
-	uint64 sinceTx  = 3;
-	bool   noWait   = 4;
+	uint64      tx                = 1;
+	EntriesSpec entriesSpec       = 2;
+	uint64      sinceTx           = 3;
+	bool        noWait            = 4;
+	bool        keepReferencesUnresolved = 5;
```
byo, line 384: How about putting it in the `EntriesSpec` structure - for each type separately?

jeroiraz: though on that, it would require some additional changes on when the snapshot was requested

byo (approved):

PR #1194, opened 2022-04-25
Fix immuclient auto login

jeroiraz:


cmd/immuclient/immuc/init.go
```diff
@@ -113,7 +113,7 @@ func (i *immuc) Execute(f func(immuClient client.ImmuClient) (interface{}, error
 	}
 
 	needsLogin := strings.Contains(err.Error(), "token has expired") ||
-		strings.Contains(err.Error(), "please login first") ||
```
byo, line 116: Maybe we should handle both cases?

byo (approved):
LGTM, a small comment left

PR #1197, opened 2022-04-26
fix(Dockerfile): Fix HOME variable for podman

bengentil:
With an empty HOME variable, immudb fail to start on podman
with the following message:
"Error: user: Current requires cgo or $HOME set in environment"

This fails here: https://cs.opensource.google/go/go/+/master:src/os/user/lookup_stubs.go;l=58
Because `HOME="${IMMUDB_HOME}"` on line 71 is not set without a prior `ARG`

For some reason docker always sets a HOME variable so it doesn't have this issue.

Signed-off-by: Benjamin Gentil <benjamin@gentil.io>


byo:
Thanks @bengentil, merged to immudb master

PR #1212, opened 2022-05-12
non-transactional statements handling

jeroiraz:
TODO: extract multidb handler from the sql engine

Note:
Database creation  (`CREATE DATABASE`) and selection  (`USE DATABASE`) statements are non-transactional in both, embedded and with external multidb handler (e.g. when using the server for processing these statements).
Therefore, this two statements are not allowed inside a `BEGIN TRANSACTION`/`COMMIT`|`ROLLBACK` block

### DIFF
--- a/embedded/tools/stress_tool_sql/stress_tool_sql.go
+++ b/embedded/tools/stress_tool_sql/stress_tool_sql.go
@@ -153,7 +153,7 @@ func main() {
 		panic(err)
 	}
 
-	err = engine.SetDefaultDatabase("defaultdb")
+	_, _, err = engine.Exec("USE DATABASE defaultdb;", map[string]interface{}{}, nil)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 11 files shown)


embedded/tools/stress_tool_sql/stress_tool_sql.go
```diff
@@ -153,7 +153,7 @@ func main() {
 		panic(err)
 	}
 
-	err = engine.SetDefaultDatabase("defaultdb")
+	_, _, err = engine.Exec("USE DATABASE defaultdb;", map[string]interface{}{}, nil)
```
byo, line 156: Minor: why not use `engine.SetCurrentDatabase("defaultdb")`?

byo (approved):
LGTM, only a minor comment left (that can be ignored)

embedded/tools/stress_tool_sql/stress_tool_sql.go
jeroiraz: it was intentionally, to ensure everything can be done with SQL statements, but yes, if using `SetCurrentDatabase` programmatically some steps would be avoided

PR #1214, opened 2022-05-12
feat(embedded/sql): create database if not exists

jeroiraz:
```sql
-- CREATE DATABASE [IF NOT EXISTS] name;

CREATE DATABASE db1;

-- statement equiv. to no-op if database already exists
CREATE DATABASE IF NOT EXISTS db1;

```


pkg/api/schema/schema.proto
```diff
@@ -464,6 +464,7 @@ message DatabaseSettings {
 message CreateDatabaseRequest {
 	string name = 1;
 	DatabaseNullableSettings settings = 2;
+	bool ifNotExists = 3;
 }
 
 message CreateDatabaseResponse {
```
byo, line 470: How about adding extra boolean to the response to indicate whether the database already existsed?

byo (approved):
LGTM, added comment about a small improvement that we could add to the response message when creating the database.

PR #1234, opened 2022-05-27
Fix/snapshot leaks

byo:


embedded/sql/engine.go
```diff
@@ -1267,16 +1267,25 @@ func (e *Engine) QueryPreparedStmt(stmt DataSource, params map[string]interface{
 	// TODO: eval params at once
 	nparams, err := normalizeParams(params)
 	if err != nil {
+		if tx == nil {
```
jeroiraz, line 1270: what about deferring the close and checking if there was an error... the benefit is to avoid having a close for each error handling...

```
defer func() {
    if err != nil {
        qtx.Cancel()
    }
}```

File: embedded/sql/stmt.go [hunk 2]
```diff
@@ -2156,17 +2156,21 @@ func (stmt *SelectStmt) Resolve(tx *SQLTx, params map[string]interface{}, _ *Sca
 	}
 
 	if stmt.joins != nil {
-		rowReader, err = newJointRowReader(rowReader, stmt.joins)
+		jointRowReader, err := newJointRowReader(rowReader, stmt.joins)
```
jeroiraz, line 2159: I'm wondering if we cannot defer the close of rowReader similarly as suggested in previous file...

embedded/sql/union_row_reader.go
```diff
@@ -63,9 +63,7 @@ func newUnionRowReader(rowReaders []RowReader) (*unionRowReader, error) {
 }
 
 func (ur *unionRowReader) onClose(callback func()) {
-	for _, r := range ur.rowReaders {
-		r.onClose(callback)
-	}
+	ur.rowReaders[0].onClose(callback)
```
jeroiraz, line 66: hm, I'd expect to call each callback

byo: It's only registering the close callback here.

And in fact we want to register this only into one child that will be closed as the last one. That's why the first child is selected and the closing of readers is done in reverse order.

This was causing issues in many places actually.

The problem is that all readers first need to be closed and only then the callback can be executed. Otherwise the close callback will be executed when readers are still active - in our the callback was closing the transaction. But while closing it, it failed on closing its associated snapshot because it still had active readers.

PR #1239, opened 2022-06-01
Fix/sessions deadlock

byo:


pkg/server/sessions/manager.go
```diff
@@ -65,30 +63,38 @@ func NewManager(options *Options) (*manager, error) {
 	if options == nil {
```
jeroiraz, line 63: just wondering if the input param could just be `Options`

pkg/server/sessions/manager.go
```diff
@@ -65,30 +63,38 @@ func NewManager(options *Options) (*manager, error) {
...
 	guard := &manager{
 		sessions: make(map[string]*Session),
 		ticker:   time.NewTicker(options.SessionGuardCheckInterval),
 		done:     make(chan bool),
 		logger:   logger.NewSimpleLogger("immudb session guard", os.Stdout),
-		options:  options,
+		options:  *options,
+	}
+
+	if guard.options.MaxSessionAgeTime == 0 {
```
jeroiraz, line 80: this settings may be moved into options struct

pkg/server/sessions/manager.go
```diff
@@ -65,30 +63,38 @@ func NewManager(options *Options) (*manager, error) {
...
+
 	return guard, nil
 }
 
 func (sm *manager) NewSession(user *auth.User, db database.DB) (*Session, error) {
 	sm.sessionMux.Lock()
 	defer sm.sessionMux.Unlock()
 
-	if len(sm.sessions) == MaxSessions {
+	if len(sm.sessions) >= sm.options.MaxSessions {
```
jeroiraz, line 97: when was the case it could be > ?

pkg/server/sessions/manager.go
```diff
@@ -42,7 +40,7 @@ type manager struct {
 	ticker     *time.Ticker
 	done       chan bool
```
jeroiraz, line 41: running, ticker and done may be prefixed with sessionGuard e.g. sessionGuardRunning, sessionGuardTicker sessionGuardDone. Or even better extract into it's own struct...

pkg/server/sessions/manager.go
byo: That could be inconsistent with other parts of the code (like other modules) and all those functions like `DefaultOptions`, `With...` modifiers - all those return by pointer.

We could of course change this to a functional-like approach - where we would work on values everywhere (like each `With...` call returns a new Options object with modifications applied instead of modifying the object itself). But then I think we should rather consistently change it in the overall codebase.

jeroiraz:
probably transactions under sessions/internal/transactions can just be moved up to sessions package...

pkg/server/sessions/manager.go
jeroiraz: or actually, it's the session manager itself and the concept of session guard is not needed... ?

jeroiraz: anyways, this are non-blocking changes

pkg/server/sessions/manager.go
byo: Fixed in a599c8d448c9b98bf8e2a9d451ebba9a3f38574e

jeroiraz (approved):

pkg/server/sessions/manager.go
byo: Hmm, currently it should not be possible due to options validation, but it would happen if someone set MaxSessions option to something negative - in such case you have unbound session count

PR #1240, opened 2022-06-03
Fix/session ids

byo:


jeroiraz (dismissed):
LGTM, I'd keep the padding in the encoded session id...

PR #1242, opened 2022-06-04
Feat scan offset

jeroiraz:
OFFSET support
Solves #1243

```sql
SELECT * FROM table1 LIMIT 2 OFFSET 1
```


pkg/api/schema/schema.proto
```diff
@@ -376,6 +377,7 @@ message ZScanRequest {
 	Score maxScore = 9;
 	uint64 sinceTx = 10;
 	bool  noWait = 11;
+	uint64 offset = 12;
```
byo, line 380: Would be good to have some small comment here (that will propagate to generated GRPC and SWAGGER code)

embedded/tbtree/reader.go
```diff
@@ -230,18 +230,18 @@ func (r *Reader) Read() (key []byte, value []byte, ts, hc uint64, err error) {
...
 
-		if len(r.prefix) > 0 && len(leafValue.key) >= len(r.prefix) {
-			leafPrefix := leafValue.key[:len(r.prefix)]
-
-			// prefix match
-			if bytes.Equal(r.prefix, leafPrefix) {
-				return cp(leafValue.key), cp(leafValue.value), leafValue.ts, leafValue.hCount, nil
-			}
+		if r.skipped < r.offset {
+			r.skipped++
```
byo, line 240: How does it work with entries marked as deleted or the expired ones?

byo: Are we still introducing skipping entries on the btree level ? I think that currently there's no valid use case for that thus it would only be a technical depth ATM.

PR #1248, opened 2022-06-09
More deterministic tests

byo:


pkg/client/streams_integration_test.go
```diff
@@ -35,16 +35,25 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
-func TestImmuServer_SimpleSetGetStream(t *testing.T) {
-	_, err := net.DialTimeout("tcp", fmt.Sprintf(":%d", DefaultOptions().Port), 1*time.Second)
-	if err != nil {
-		t.Skip(fmt.Sprintf("Please launch an immudb server at port %d to run this test.", DefaultOptions().Port))
+func externalImmudbClient(t *testing.T) (*immuClient, context.Context) {
```
jeroiraz, line 38: hm, it seems intricate, this tests are skipped, right?

byo: It seems like some tests require external immudb running in the background, otherwise those are skipped.

The problem is that some other tests actually start immudb on the default port, previous detection if external immudb is running was checking if that port is open. But if it was opened because of other test running (like the GC cycle did not yet clenaup objects from the previous tests and the TCP listener is still open), the test was running against such a temporary server. The result usually is that at the time the test really tries to connect to external immudb, it complains that the port is already closed.

jeroiraz (dismissed):

PR #1249, opened 2022-06-10
Feat/immuclient json output

byo:


cmd/immuclient/command/commandline.go
```diff
@@ -27,6 +33,24 @@ func (cl *commandline) ConfigChain(post func(cmd *cobra.Command, args []string)
 		if err = cl.config.LoadConfig(cmd); err != nil {
 			return err
 		}
+
+		if viper.GetBool("value-only") {
```
jeroiraz, line 37: as discussed online, I'd interpret `value-only` as a data filtering flag rather than a data serialization format.
I think here we support two data formats: plain-text and json.
`value-only` flag may be primary useful when using plain-text format but in the future a more elaborated data filtering can be implemented

PR #1256, opened 2022-06-28
fix(cmd/immudb): include metrics endpoint related flags

jeroiraz:


byo (approved):
LGTM, but some tests are still failing

PR #1257, opened 2022-06-28
Tx Timestamp and SQL NOW function

jeroiraz:
This PR includes a fixed timestamp associated to ongoing transactions.

SQL `now()` function evals to the fixed transaction timestamp. Thus making `now()` an idempotent function in the scope of the same transaction.

```sql
-- column `ts` of the three inserted rows will contain the same value

INSERT INTO mytable(id, ts) VALUES (1, NOW()), (2, NOW()), (3, NOW());

-- or equivalently

BEGIN TRANSACTION;
    INSERT INTO mytable(id, ts) VALUES (1, NOW());
    INSERT INTO mytable(id, ts) VALUES (2, NOW()), (3, NOW());
COMMIT;
```

### DIFF
--- a/embedded/sql/engine_test.go
+++ b/embedded/sql/engine_test.go
@@ -428,10 +428,15 @@ func TestNowFunctionEvalsToTxTimestamp(t *testing.T) {
 		for i := 0; i < 10; i++ {
 			row, err := r.Read()
 			require.NoError(t, err)
-			require.EqualValues(t, tx.Timestamp().UTC(), row.ValuesBySelector[EncodeSelector("", "db1", "tx_timestamp", "ts")].Value())
+			require.EqualValues(t, tx.Timestamp(), row.ValuesBySelector[EncodeSelector("", "db1", "tx_timestamp", "ts")].Value())
 		}
 
+		_, err = r.Read()
+		require.ErrorIs(t, err, ErrNoMoreRows)
+
 		currentTs = tx.Timestamp()
+
+		time.Sleep(1 * time.Microsecond)
 	}
 }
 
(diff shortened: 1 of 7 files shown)


embedded/sql/engine_test.go
```diff
@@ -386,6 +386,55 @@ func TestTimestampCasts(t *testing.T) {
...
+		r, err := engine.Query("SELECT * FROM tx_timestamp WHERE ts = @ts", map[string]interface{}{"ts": tx.Timestamp()}, nil)
+		require.NoError(t, err)
+		defer r.Close()
+
+		for i := 0; i < 10; i++ {
+			row, err := r.Read()
+			require.NoError(t, err)
+			require.EqualValues(t, tx.Timestamp().UTC(), row.ValuesBySelector[EncodeSelector("", "db1", "tx_timestamp", "ts")].Value())
+		}
+
```
byo, line 433: Please check here if attempt to read one more results in ErrNoMoreRows

byo (commented):
This will also require clarification in the documentation as it is changing immudb's behavior.

embedded/sql/engine_test.go
jeroiraz: not completely sure about that, if tx processing is fast enough, there could be more than one transaction "at the same time".. lets discuss it

byo: True, more Txs can end up with the same timestamp. But how then would we make sure that the tx timestamp is changing?
I think we can use `time.Sleep` here since it has the lower-bound of delay guaranteed (like sleep 2 microseconds between rounds).

byo (approved):

PR #1263, opened 2022-07-04
Fix/scan revision

### LINKED ISSUE #1262: client.Get vs client.Scan for the same key provide different results
I am using immudb version 1.3.1.
In a database I have one key (entry) and I try to retrieve it with Get and Scan commands.
 
What happened
Although I extracted, from the database, the same key:
Get output -> says the entry is at Revision: 1 
Scan output -> says the entry is at Revision: 0

After an update on the aforementioned key:
Get output -> says that entry is at Revision: 2
Scan output -> says that entry is at Revision: 0

So, it is obvious that Revision value returned by the Scan command is incorrect, actually always is the default value of the data type uint64 (0).

What you expected to happen
Scan command to return consistent information related to a key - provide correct info related to the Revision.

How to reproduce it :
```
package main

import (
	"context"
	"log"

	immudb "github.com/codenotary/immudb/pkg/client"
)

func main() {
	opts := immudb.DefaultOptions().WithAddress("localhost").WithPort(3322)
	client := immudb.NewClient().WithOptions(opts)
	err := client.OpenSession(context.TODO(), []byte(`immudb`), []byte(`immudb`), "defaultdb")
	if err != nil {
		log.Fatal(err)
	}

	defer client.CloseSession(context.TODO())

	key := "key1"
	val := "val11"

	_, err = client.Set(context.TODO(), []byte(key), []byte(val))
	if err != nil {
		log.Fatal(err)
	}

	entry1, err := client.Get(context.TODO(), []byte(key))
	if err != nil {
		log.Fatal(err)
	}
	log.Printf("Get.  Revision:%d  Key:%s Value:%s \n", entry1.GetRevision(), entry1.GetKey(), entry1.GetValue())

	entries, err := client.Scan(context.TODO(), &schema.ScanRequest{
		Prefix: []byte(key),
	})
	if err != nil {
		log.Fatal(err)
	}

	for _, entry2 := range entries.GetEntries() {
		log.Printf("Scan. Revision:%d  Key:%s Value:%s \n", entry2.GetRevision(), entry2.GetKey(), entry2.GetValue())
	}
}
```

byo:
Correctly calculate key revision when doing a Scan request. Closes #1262.

### DIFF
--- a/pkg/database/database_test.go
+++ b/pkg/database/database_test.go
@@ -1790,60 +1790,81 @@ func TestRevisionGetConsistency(t *testing.T) {
 	db, closer := makeDb()
 	defer closer()
 
-	keyTx, err := db.Set(&schema.SetRequest{KVs: []*schema.KeyValue{{
-		Key:   []byte("key"),
-		Value: []byte("value"),
-	}}})
-	require.NoError(t, err)
+	var keyTxId uint64
 
-	_, err = db.SetReference(&schema.ReferenceRequest{
-		Key:           []byte("reference"),
-		ReferencedKey: []byte("key"),
-	})
-	require.NoError(t, err)
+	// Repeat the test for different revision numbers
+	for i := 0; i < 10; i++ {
 
-	_, err = db.SetReference(&schema.ReferenceRequest{
-		Key:           []byte("reference"),
-		ReferencedKey: []byte("key"),
-		AtTx:          keyTx.Id,
-		BoundRef:      true,
-	})
-	require.NoError(t, err)
-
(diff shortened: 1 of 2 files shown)


pkg/database/database_test.go
```diff
@@ -1786,6 +1786,66 @@ func TestGetAtRevision(t *testing.T) {
 	})
 }
 
+func TestRevisionGetConsistency(t *testing.T) {
```
jeroiraz, line 1789: probably a test with more than key updates would be useful

jeroiraz (approved):

Issue #1254, opened 2022-06-27
Add `ServerInfo` gRPC endpoint and client functions


joe-mann:
```go
ServerInfo(context.Context, *schema.ServerInfoRequest) (*schema.ServerInfoResponse, error)
```

To contain the same fields as `schema.HealthResponse` initially, with possibility for expansion later.

- [ ] Add `rpc ServerInfo`
- [ ] Expose `ServerInfo` in Go client
- [ ] Expose `ServerInfo` in other [placeholder] clients
- [ ] Expose `ServerInfo` in immuclient
- [ ] Deprecate `rpc Health`
- [ ] Deprecate `HealthCheck` in Go client
- [ ] Deprecate usages of `rpc Health` and `rpc HealthCheck` in Go client
- [ ] Deprecate usages of `rpc Health` and `rpc HealthCheck` in other [placeholder] clients

PR #1266, opened 2022-07-05
Implement `rpc ServerInfo`

joe-mann:
Part of https://github.com/codenotary/immudb/issues/1254.

Implements `rpc ServerInfo` whose response currently just contains the version string.


pkg/api/schema/schema.proto
```diff
@@ -321,6 +321,13 @@ message VerifiableGetRequest {
 	uint64 proveSinceTx = 2;
 }
 
+message ServerInfoRequest {}
+
+message ServerInfoResponse {
+	bool status = 1;
```
byo, line 327: Please add some comments what the fields in the `ServerInfoResponse` message mean.

pkg/server/server.go
```diff
@@ -696,9 +696,15 @@ func (s *ImmuServer) UpdateMTLSConfig(ctx context.Context, req *schema.MTLSConfi
 	return nil, ErrNotSupported
 }
 
+// ServerInfo returns health and status information about the server instance.
```
byo, line 699: Please add / move the comment into the `.proto` file - that way it will propagate to generated code too.

pkg/server/server_test.go
```diff
@@ -790,6 +790,15 @@ func testServerHistoryError(ctx context.Context, s *ImmuServer, t *testing.T) {
 
 }
 
+func testServerInfo(ctx context.Context, s *ImmuServer, t *testing.T) {
+	h, err := s.ServerInfo(ctx, &schema.ServerInfoRequest{})
+	require.NoError(t, err)
+
+	if !h.GetStatus() {
+		t.Fatalf("ServerInfo, expected %v, got %v", true, h.GetStatus())
+	}
```
byo, line 799: Could be written as:

```suggestion
	require.True(t, h.GetStatus())
```

byo (commented):
I think  this PR is still missing changes to the immudb client SDK - even if its present on the server-side it would still require a low-level operation made on the client side.

Also I'm not sure about the `status` field in the response. Shouldn't this endpoint only result some static information about the server such as the version, maybe some git commit etc in the future?

joe-mann:
> I think this PR is still missing changes to the immudb client SDK - even if its present on the server-side it would still require a low-level operation made on the client side.

Yep, I deliberately didn't include those additions in this PR to keep it a little smaller. But I can make them here if you prefer - all the same to me.

> Also I'm not sure about the `status` field in the response. Shouldn't this endpoint only result some static information about the server such as the version, maybe some git commit etc in the future?

The purpose of this PR was to expose both of the existing `status` and `version` fields. However, I now see that `status` is always `true`, so I don't see the harm in removing it (a successful response serving the same purpose of an explicit `status = true`).

pkg/api/schema/docs.md
```diff
@@ -3,147 +3,149 @@
 
 ## Table of Contents
 
-- [schema.proto](#schema.proto)
```
byo, line 6: This looks like some protobuf generator version mismatch - that doc itself seems to be OK - it's only internal anchor names that changed, however it produces a large diff. Can you check if there's some trivial way to avoid this?

byo (dismissed):
LGTM, just one comment regarding protobuf compiler mismatch producing larger than usual docs update.

byo (dismissed):

byo (approved):

PR #1268, opened 2022-07-07
Fix/support begin without transaction

byo:
This PR adds an alternative syntax `BEGIN TRANSACTION` statement to start a new transaction with a simple `BEGIN` statement.

This change is required to maintain compatibility with some psql clients such as [psycopg2](https://pypi.org/project/psycopg2/)


jeroiraz:
as discussed offline, it may be better to keep a unique way to initiate and finalize transactions i.e. keep using `BEGIN TRANSACTION` or its shortcut `BEGIN` but keeping `START TRANSACTION` for future consideration

jeroiraz (dismissed):

PR #1274, opened 2022-07-11
Use `protoc-gen-*` tools as versioned in `go.mod`

joe-mann:
This PR fixes the intended behaviour of `make build/codegen`.

We currently version 4 tools (by importing them to `tools.go`):

- `protoc-gen-go`;
- `protoc-gen-grpc-gateway`;
- `protoc-gen-swagger`; and,
- `protoc-gen-doc`

However, the user must explicitly `go install`s these 4 tools. Failing to do this results in an error if the tools have not been installed, or a different version of the tools already installed causes unexpected generation results.

This PR does 2 things:

- Fix the versions of tools to those currently used to generate code; and,
- Add shims to call `go run protoc-gen-*` instead of `protoc-gen-*`, causing the version of the tool specified in `go.mod` to be used.

The effects are:

- It's no longer necessary to install these tools before running `make build/codegen`. This also means that if you already have different versions installed (e.g. from different projects), you're not overriding them.
- All developers who run `make build/codegen` will see the same generation results. The only exception here in the version of `protoc` being used, but this is not easily versioned (not a Go tool), and in any case has little/no effect on the generated code (often just a version number in the header).


go.mod
```diff
@@ -8,8 +8,10 @@ require (
 	github.com/Masterminds/sprig v2.22.0+incompatible // indirect
 	github.com/aead/chacha20poly1305 v0.0.0-20201124145622-1a5aba2a8b29 // indirect
 	github.com/fatih/color v1.13.0
+	github.com/fsnotify/fsnotify v1.5.4 // indirect
```
byo, line 11: Some new dependencies unrelated to the PR were added - can you check if they go away with `go mod tidy` ?

go.mod
```diff
@@ -19,32 +21,40 @@ require (
...
 	github.com/rakyll/statik v0.1.7
 	github.com/rogpeppe/go-internal v1.8.0
 	github.com/rs/xid v1.3.0
 	github.com/schollz/progressbar/v2 v2.15.0
+	github.com/spf13/afero v1.8.2 // indirect
+	github.com/spf13/cast v1.5.0 // indirect
 	github.com/spf13/cobra v1.2.1
 	github.com/spf13/pflag v1.0.5
-	github.com/spf13/viper v1.12.0
+	github.com/spf13/viper v1.9.0
```
byo, line 48: This version downgrade looks suspicious, isn't that by mistake?

tools.go
```diff
@@ -1,13 +1,14 @@
-// +build tools
+//go:build tools
```
byo, line 1: `//go:build` was introduced in 1.17 but we're still bound to compile with 1.15 - isn't that going to disrupt the compilation process if 1.15 is used?

go.mod
joe-mann: `github.com/spf13/viper` and `google.golang.org/genproto` appear to share incompatible dependencies, and the downgrade happened when pinning one of the tools to the last-used version.

But both are now updated.

tools.go
joe-mann: Tested with 1.15, and it compiles fine.

byo (approved):

byo (approved):

byo (approved):

PR #1275, opened 2022-07-12
Rollback after failed exec

ldej:
Performing a `Rollback` after a failed `SQLExec` results in a `no transaction found` error

### DIFF
--- a/embedded/sql/engine.go
+++ b/embedded/sql/engine.go
@@ -1196,8 +1196,11 @@ func (e *Engine) execPreparedStmts(stmts []SQLStmt, params map[string]interface{
 
 		ntx, err := stmt.execAt(currTx, nparams)
 		if err != nil {
-			currTx.Cancel()
-			return nil, committedTxs, stmts[execStmts:], err
+			if tx == nil {
+				currTx.Cancel()
+				return nil, committedTxs, stmts[execStmts:], err
+			}
+			return currTx, committedTxs, stmts[execStmts:], err
 		}
 
 		if !currTx.closed && !currTx.explicitClose && e.autocommit {


pkg/integration/transaction_test.go
```diff
@@ -105,29 +106,56 @@ func TestTransaction_Rollback(t *testing.T) {
...
+		);`, nil)
+		require.NoError(t, err)
+
+		err = tx.SQLExec(context.TODO(), `
+			INSERT INTO table1 (id) VALUES (1);
+			INSERT INTO table1 (id) VALUES (1);
+		`, nil)
+		require.EqualError(t, err, "key already exists")
+
+		err = tx.Rollback(context.TODO())
```
ldej, line 156: This fails

byo: I think we could even extend this test - after failure in a single statement you should still be able to execute another valid statement and successfully commit the transaction.

Such failed statement in the middle should not invalidate the whole transaction but a single statement.

But then it may be much harder to implement.

What I believe is happening here is that a broken INSERT that violates constraints could make the internal state of the transaction invalid and the simplest way to handle that situation would be to destroy the transaction completely.

A proper fix would be to implement some internal rollback within a single transaction (thus some form of nested transactions).

ldej: You are correct that a failing `INSERT` results in a closed transaction. This only makes sense when the transaction is created within `execPreparedStmts`. However, when a transaction is provided as a parameter for `execPreparedStmts`, it shouldn't close the transaction.

I am not sure if you want to continue executing statements after the failure of a single statement. In that case you could use the `UPSERT` or `IF NOT EXISTS` statements. What benefit would continuing give?

ldej:
I solved this by only closing (`Cancel`) a transaction if it has been created within `execPreparedStmts`. I looked at how this was solved in `QueryPreparedStmt`.

embedded/sql/engine.go
```diff
@@ -1196,8 +1196,11 @@ func (e *Engine) execPreparedStmts(stmts []SQLStmt, params map[string]interface{
 
 		ntx, err := stmt.execAt(currTx, nparams)
 		if err != nil {
-			currTx.Cancel()
-			return nil, committedTxs, stmts[execStmts:], err
+			if tx == nil {
+				currTx.Cancel()
+				return nil, committedTxs, stmts[execStmts:], err
+			}
+			return currTx, committedTxs, stmts[execStmts:], err
```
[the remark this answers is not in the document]
ldej: Hmm, you are going to have to help me with that. 

Maybe it's better to just `Cancel` the transaction, but when `Rollback` is called and the transaction is already closed, to not return `no transaction found`?

pkg/integration/transaction_test.go
byo: After discussing this with @ldej:

The proposed solution is to add a new flag (e.g. `broken`) into SQLTx class. That flag when set to true would block any operations on Tx data - both reads and writes (`newKeyReader`, `get`, `set`, `existKeyWith`, `commit`) - allowing only the `Cancel` operation. That way we can ensure there are no operations that would rely on the potentially broken state of the transaction.

PR #1280, opened 2022-07-15
Add S3 support and multiple-replicas to Helm Chart

andreimerfu:
This pull request adds two important features to the helm chart deployment strategy.

- ✅ S3 Storage support
- ✅ Multiple read-replicas support

These features are so important for in-cluster deployment using AWS Cloud.

### DIFF
--- a/helm/templates/service.yaml
+++ b/helm/templates/service.yaml
@@ -24,12 +24,46 @@ metadata:
   name: {{ include "immudb.fullname" . }}-grpc
   labels:
     {{- include "immudb.labels" . | nindent 4 }}
-  annotations:  
-    traefik.ingress.kubernetes.io/service.serversscheme: h2c
 spec:
   type: {{ .Values.service.type }}
   ports:
     - port: {{ .Values.service.ports.grpc }}
+      targetPort: grpc
+      protocol: TCP
+      name: grpc
+  selector:
+    {{- include "immudb.selectorLabels" . | nindent 4 }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "immudb.fullname" . }}-replica-http
+  labels:
+    {{- include "immudb.labels" . | nindent 4 }}
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+    - port: {{ .Values.replicaService.ports.http }}
+      targetPort: http
+      protocol: TCP
+      name: http
+    - port: {{ .Values.replicaService.ports.metrics }}
+      targetPort: metrics
+      protocol: TCP
+      name: metrics
+  selector:
+    {{- include "immudb.selectorLabels" . | nindent 4 }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "immudb.fullname" . }}-replica-grpc
+  labels:
+    {{- include "immudb.labels" . | nindent 4 }}
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+    - port: {{ .Values.replicaService.ports.grpc }}
       targetPort: grpc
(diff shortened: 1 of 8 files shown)


helm/templates/statefulset.yaml
```diff
@@ -31,13 +28,45 @@ spec:
       volumes:
       - name: immudb-storage
         persistentVolumeClaim:
-          claimName: {{ include "immudb.fullname" . }}
+          claimName: immudb-claim
```
byo, line 31: All nodes in the cluster (both the leader and all replicas) will use the same persistent volume, immudb was not designed to handle such situation that could end up with data corruption.

For this purpose, each replica should have its own persistent volume assigned using [Volume Claim Template](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#volume-claim-templates)

helm/values.yaml
```diff
@@ -13,9 +13,13 @@ image:
 imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
-volume:
-  class: ""
-  size: 5Gi
+
+persistence:
```
byo, line 17: This would change the way persistent volumes are declared, could potentially break existing installations when upgrading helm version. Can we add some compatibility mode here ? e.g. if the `volume.size` is present use the old definition rather than the new one?

helm/values.yaml
```diff
@@ -13,9 +13,13 @@ image:
...
+  enabled: true
+  storageClass: gp2
```
byo, line 19: Please leave the default `storageClass` empty.

helm/values.yaml
```diff
@@ -39,16 +43,20 @@ service:
...
+replicaService:
+  type: ClusterIP
+  ports:
+    grpc: 3334
+    http: 9090
+    metrics: 9498
+
 ingress:
-  enabled: false
+  enabled: "false"
```
byo, line 54: Why is this change needed?

byo (changes requested):

helm/templates/service.yaml
```diff
@@ -24,8 +24,6 @@ metadata:
   name: {{ include "immudb.fullname" . }}-grpc
   labels:
     {{- include "immudb.labels" . | nindent 4 }}
-  annotations:  
-    traefik.ingress.kubernetes.io/service.serversscheme: h2c
```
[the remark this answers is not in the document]
SimoneLazzaris: Yes, it is needed for traefik-based ingresses to handle gRPC calls correctly

helm/templates/service.yaml
```diff
@@ -35,3 +33,39 @@ spec:
...
+      targetPort: http
+      protocol: TCP
+      name: http
+    - port: {{ .Values.replicaService.ports.metrics }}
+      targetPort: metrics
+      protocol: TCP
+      name: metrics
+  selector:
+    {{- include "immudb.selectorLabels" . | nindent 4 }}
+---
```
SimoneLazzaris, line 56: Using just one service for both master and replicas makes it impossible to split read and write requests. It would be best to define two different services: one for talking with the main instance, one for talking with replicas.

helm/templates/statefulset.yaml
```diff
@@ -1,6 +1,3 @@
-{{- if gt (.Values.replicaCount | toString | atoi) 1 }}
-{{- fail "At the moment, you can just have 1 instance of immudb. We are working to raise that limit."}}
-{{- end }}
 apiVersion: apps/v1
 kind: StatefulSet
 metadata:
```
SimoneLazzaris, line 3: Again, using just one statefulset makes it impossible to distinguish between master and replicas.

PR #1284, opened 2022-07-16
fix: access tls value in global scope within ingress annotations

tauu:
The helm chart contains a bug within the ingress template. It uses `.Values.ingress.tls.enabled` to check if tls is enabled within the annotations settings. As the annotations section is surrounded with a `with` definition, the scope is changed within it to `.Values.ingress.annotations`. Therefore  `.Values.ingress.tls.enabled` actually looks up `.Values.ingress.annotations.Values.ingress.tls.enabled`, which does not exist.

The fix adds a `$` to actually use the global scope when checking if tls is enabled.

### DIFF
--- a/helm/templates/ingress.yaml
+++ b/helm/templates/ingress.yaml
@@ -16,7 +16,7 @@ metadata:
   {{- with .Values.ingress.annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
-    {{- if .Values.ingress.tls.enabled }}
+    {{- if $.Values.ingress.tls.enabled }}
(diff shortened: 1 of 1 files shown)


byo:
Hi @tauu, thanks for your contribution.

The fix is ready to be merged :+1:

Before it's merged though, can you please:

1. Change the commit message to `fix(helm): access tls value in global scope within ingress annotations` (to produce better changelog entries)
2. Add `Signed-off-by: ...` clause (it's added automatically with `git commit -s`)

Also if possible please gpg-sign your commits.

Issue #1278, opened 2022-07-14
Invalid client state after connection refused


grossvater:
**What happened**
When `OpenSession()` returns "connection refused" (e.g server not running), next call to `OpenSession()` fails with "session already opened". Trying then to close the session with `CloseSession()` will result in panic.

**What you expected to happen**
The second call to `OpenSession()` should return the same error, "connection refused".

**How to reproduce it (as minimally and precisely as possible)**
```
user := "immudb"
passwd := "immudb"
db := "testdb"

ctx := context.Background()
opts := immudb.DefaultOptions().
	WithAddress("localhost").
	WithPort(1000)
client := immudb.NewClient().WithOptions(opts)

err := client.OpenSession(ctx, []byte(user), []byte(passwd), db)
if err == nil {
	panic("who's listening?!")
}
fmt.Println(err)

err = client.OpenSession(ctx, []byte(user), []byte(passwd), db)

// session already opened
fmt.Println(err.Error())

defer func() {
	if err := recover(); err != nil {
		fmt.Printf("panic recovered: %s\n", err)
	}
}()

// close will panic
_ = client.CloseSession(ctx)
```

PR #1285, opened 2022-07-18
fix(pkg/client): Invalid client state after connection refused

michaelbacci:
fix [#1278](https://github.com/codenotary/immudb/issues/1278)


pkg/client/session.go
```diff
@@ -16,6 +16,13 @@ import (
 )
 
 func (c *immuClient) OpenSession(ctx context.Context, user []byte, pass []byte, database string) (err error) {
+	defer func() {
```
byo, line 19: Please add separate defer statements for each of the variables zeroed below. Those should be set up right after assigning value to them.

Also each of them will require some kind of deinitialization:

E.g. for the `SessionId` the defer right after assignment:

```go
	c.SessionID = resp.GetSessionID()
	defer func(){
		if err != nil {
			c.ServiceClient.CloseSession(ctx, new(empty.Empty))
			c.SessionID = nil
		}
	}()
```

For the `HeartBeater` it will require `c.HearBeater.Stop()`.

The `ServiceClient` member does not need deinitialization.

There's one more value to deinitialize in case of an error: `c.clientConn`, that one should be deinitialized with `c.clientConn.Close()` and set to nil too.

byo (changes requested):
Thank you @michaelbacci for your contribution. The fix is definitely in the right direction. I've added comment on how to make it even more robust and prevent resource leaks by freeing them up on error.

pkg/client/session.go
```diff
@@ -39,6 +46,11 @@ func (c *immuClient) OpenSession(ctx context.Context, user []byte, pass []byte,
 	}
 
 	c.ServiceClient = schema.NewImmuServiceClient(c.clientConn)
+	defer func(){
+		if err != nil {
+			_, _ = c.ServiceClient.CloseSession(ctx, new(empty.Empty))
```
jeroiraz, line 51: it's rare to defer the session closing before opening it. I'd expect the close statement after the line where it's opened

pkg/client/session.go
```diff
@@ -16,6 +16,13 @@ import (
 )
 
 func (c *immuClient) OpenSession(ctx context.Context, user []byte, pass []byte, database string) (err error) {
+	defer func(){
+		if err != nil && err != ErrSessionAlreadyOpen {
+			_ = c.clientConn.Close()
```
jeroiraz, line 21: I'd expect the close the connection after the line where the connection is first established

pkg/client/session.go
```diff
@@ -50,8 +62,18 @@ func (c *immuClient) OpenSession(ctx context.Context, user []byte, pass []byte,
 	}
 
 	c.SessionID = resp.GetSessionID()
+	defer func(){
```
jeroiraz, line 65: probably it'd be better to set a local variable and set the attribute in the client when all steps succeeded

pkg/client/session.go
jeroiraz: not sure if the error comparison will match `err != ErrSessionAlreadyOpen`

pkg/client/session.go
michaelbacci: Hi @jeroiraz here the logic that I apply as suggested (as I interpret it) by @byo : for each declared variable that can be reseted by an error, I add the condition that reset the relative variable just the next line after; so they are related.

An exception is `c.clientConn` that is outside scope and that's why I put it to the top!

pkg/client/session.go
michaelbacci: 1. See my answer above
2. I've already tested: `require.Equal(t, ErrSessionAlreadyOpen, errors.FromError(ErrSessionAlreadyOpen))`

pkg/client/session.go
michaelbacci: Yep, there are several approach, I follow @byo suggestion but I can change: Who vote ?

michaelbacci:
Code updated.

pkg/client/session.go
byo: Yup, I think this particular `defer` should be a bit lower - right after `serviceClient.OpenSession` call.

pkg/client/session.go
byo: I like the current approach with separate variables :+1:

michaelbacci:
There is an issue with the coverage:   
the variable `${{ secrets.COVERALLS_TOKEN }}` seem missing so the CI can't right execute the coverage command `goveralls -coverprofile=coverage.out -service=gh-ci -repotoken ${{ secrets.COVERALLS_TOKEN }}`

byo (dismissed):

pkg/client/session.go
```diff
@@ -34,13 +31,26 @@ func (c *immuClient) OpenSession(ctx context.Context, user []byte, pass []byte,
...
 	}
+	defer func(){
+		if err != nil {
+			_ = clientConn.Close()
+		}
+	}()
 
-	c.ServiceClient = schema.NewImmuServiceClient(c.clientConn)
+	serviceClient := schema.NewImmuServiceClient(clientConn)
+	defer func(){
```
byo, line 47: That defer should go below `serviceClient.OpenSession`, correct?

byo:
Hi @michaelbacci , sorry it took so long - but I finally managed to get it through the coverage point :)

There's one remaining part in the code regarding the deferred closing of the session - can you take a look?
We also keep the immudb history linear - please rebase on the current master branch. Also if its not a problem please gpg-sign your commits.

byo (approved):

byo:
Hi @michaelbacci PR is all green and ready to go :). If you'd like to keep the original GPG signature, please rebase on top of current master and force-push.

byo:
Hi @michaelbacci the master branch has moved forward again, since we keep git history linear for the master branch that would require another rebase from your side. I think that it would be easier to rebase and merge through github then. Commits would then become gpg-signed by github but will still be viewed as verified. Let me know if that's OK with you, I'll merge this PR today if you have no objections.

PR #1287, opened 2022-07-19
Chore/perf improvements

byo:


pkg/database/database.go
```diff
@@ -1439,6 +1477,44 @@ func (d *db) isReplica() bool {
 	return d.options.replica
 }
 
+func (d *db) newTxHolder() *store.Tx {
+	d.txHolderPoolMutex.Lock()
+	defer d.txHolderPoolMutex.Unlock()
+
+	d.txHolderPoolUsed++
```
jeroiraz, line 1484: some limit in the pool may be needed

byo: What should happen if the pool is exhausted? Should be fail with error about reaching max read concurrency?

That would change the behavior compared to what we do right now where there's no limit (although this means we can have unbound growth of memory usage).

Alternatively we could wait for the holder to become ready here - but I wouldn't like to go that way without propagation of context cancellation first.

embedded/store/key_reader.go
```diff
@@ -169,7 +167,6 @@ func (s *Snapshot) NewKeyReader(spec *KeyReaderSpec) (*KeyReader, error) {
 		filters:        spec.Filters,
 		refInterceptor: refInterceptor,
 		offset:         spec.Offset,
-		_tx:            s.st.NewTxHolder(),
```
jeroiraz, line 172: I wouldn't expose the txHolder into the exposed APIs but internally fetch the holder from the pool...

embedded/store/txpool.go
```diff
@@ -0,0 +1,123 @@
...
+type txPool struct {
+	pool      []*Tx
+	used      int
+	max       int
+	nentries  int
+	maxKeyLen int
+	m         sync.Mutex
+}
+
+func newTxPool(nentries int, maxKeyLen int, poolSize int, preallocated bool) (TxPool, error) {
```
jeroiraz, line 39: probably a candidate for an options parameter

pkg/database/database.go
```diff
@@ -473,19 +495,18 @@ func (d *db) Get(req *schema.KeyRequest) (*schema.Entry, error) {
 		return d.getAtRevision(EncodeKey(req.Key), req.AtRevision)
 	}
 
-	return d.getAtTx(EncodeKey(req.Key), req.AtTx, 0, d.st, d.st.NewTxHolder(), 0)
+	return d.getAtTx(EncodeKey(req.Key), req.AtTx, 0, d.st, 0)
 }
 
 func (d *db) get(key []byte, index store.KeyIndex, txHolder *store.Tx) (*schema.Entry, error) {
```
jeroiraz, line 501: txHolder parameter may be removed?

embedded/store/key_reader.go
byo: But that would mean that we'll take the TxHolder for the pool that was meant to limit the commit concurrency. Is that something we're OK with? Thsi may lead to a situation where simultaneous readers could penalize / starve committers.

byo: Tx Holder is no longer needed in key reader

cmd/immuadmin/command/database.go
```diff
@@ -37,6 +39,8 @@ func addDbUpdateFlags(c *cobra.Command) {
 	c.Flags().String("replication-follower-username", "", "set username used for replication")
 	c.Flags().String("replication-follower-password", "", "set password used for replication")
 	c.Flags().Uint32("write-tx-header-version", 1, "set write tx header version (use 0 for compatibility with immudb 1.1, 1 for immudb 1.2+)")
+	c.Flags().Uint32("max-commit-concurrency", store.DefaultMaxConcurrency, "set the maximum commit concurrency")
```
byo, line 42: To discuss:

I called this `max-commit-concurrency` since currently that is what this parameter is doing inside store.

jeroiraz: we might even change the internal naming as well

pkg/database/sql.go
```diff
@@ -92,7 +92,11 @@ func (d *db) VerifiableSQLGet(req *schema.VerifiableSQLGetRequest) (*schema.Veri
 		}
 	}
 
-	tx := d.st.NewTxHolder()
+	tx, err := d.allocTx()
```
jeroiraz, line 95: it might be moved some lines below, right before its usage

pkg/database/sorted_set.go
```diff
@@ -258,32 +256,33 @@ func (d *db) VerifiableZAdd(req *schema.VerifiableZAddRequest) (*schema.Verifiab
 		return nil, store.ErrIllegalArguments
 	}
 
-	txMetatadata, err := d.ZAdd(req.ZAddRequest)
+	lastTx, err := d.allocTx()
```
jeroiraz, line 259: it may be moved some lines below, right before its use

pkg/database/reference.go
```diff
@@ -126,32 +124,35 @@ func (d *db) VerifiableSetReference(req *schema.VerifiableReferenceRequest) (*sc
 		return nil, store.ErrIllegalArguments
 	}
 
-	txMetatadata, err := d.SetReference(req.ReferenceRequest)
+	// Preallocate tx buffers
+	lastTx, err := d.allocTx()
```
jeroiraz, line 128: it may be moved some lines below, right before its use

embedded/store/txpool.go
```diff
@@ -0,0 +1,126 @@
...
+package store
+
+import (
+	"sync"
+
+	"github.com/codenotary/immudb/embedded/htree"
+)
+
+type txPoolOptions struct {
+	PoolSize     int
```
jeroiraz, line 25: options struct may be public? or attributes private?

embedded/store/txpool.go
```diff
@@ -0,0 +1,126 @@
...
+	return ret, nil
+}
+
+func (p *txPool) Alloc() (*Tx, error) {
+	p.m.Lock()
+	defer p.m.Unlock()
+
+	if p.used == len(p.pool) {
+		if p.used >= p.opts.PoolSize {
+			return nil, ErrMaxConcurrencyLimitExceeded
```
jeroiraz, line 101: error raised by the pool might be due to the pool limit

embedded/store/txpool.go
```diff
@@ -0,0 +1,126 @@
...
+		txEntryPtrBuffer := make([]*TxEntry, opts.PoolSize*opts.MaxTxEntries)
+		txEntryBuffer := make([]TxEntry, opts.PoolSize*opts.MaxTxEntries)
+		headerBuffer := make([]TxHeader, opts.PoolSize)
+		keyBuffer := make([]byte, opts.PoolSize*opts.MaxTxEntries*opts.MaxKeyLen)
+
+		for i := 0; i < opts.PoolSize; i++ {
+			tx := &txBuffer[i]
+			ret.pool = append(ret.pool, tx)
+
+			tx.entries = txEntryPtrBuffer[:opts.MaxTxEntries]
```
jeroiraz, line 74: aren't all txs getting the same slice?

jeroiraz: oh, final lines are doing the trick... `txEntryPtrBuffer = txEntryPtrBuffer[opts.MaxTxEntries:]`

jeroiraz: was expecting `i` to participate, it may be easier to read as it doesn't require to read all the lines

embedded/store/txpool.go
```diff
@@ -0,0 +1,126 @@
...
+		opts: opts,
+	}
+
+	if opts.Preallocated {
+		// The pool uses only 5 allocations in total
+		// instead of allocating data separately for each tx and then each
+		// entry, we instead allocate single large arrays large enough
+		// to fulfill requirements of all internal objects
+		//
+		// Keeping the number of allocated objects small is essential
```
jeroiraz, line 61: but it may make it harder to dynamically resize the pool...

embedded/store/txpool.go
jeroiraz: does it makes the difference to white-box initialize the txholders here compared of using newTx method? which it's already alloc optimized with a similar approach?

embedded/store/immustore_test.go
```diff
@@ -41,6 +41,17 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func immustoreClose(t *testing.T, immuStore *ImmuStore) {
+	err := immuStore.Close()
+	if !t.Failed() {
+		require.NoError(t, err)
```
jeroiraz, line 47: I'm wondering if error assertion does it work even when the method is deferred..

embedded/store/txpool.go
jeroiraz: another point to consider, this shared array may lead to concurrency issues... as the requirement for this to be consistent, the array must be threadsafe...

embedded/store/immustore_test.go
byo: Yes, I used that before to catch errors with session leaks.

And for that reason I also only check if the test did not fail before - in many cases when some requirement fails and breaks the test in the middle we did not yet cleanup properly and the error here was heavily polluting the test error report.

embedded/store/txpool.go
byo: regarding, this is thread safety - we don't change the allocation or slices once the initialization is done. Elements of the same array can also be written to concurrently (good explanation is here: https://stackoverflow.com/a/49879469/18309616) - as long as each goroutine works on its own part of the array, it is safe. I was confused by this some time ago too - I thought that it would be thread safe as long as each goroutine works on a separate cache line in CPU but it seems like CPUs have a way around this.

byo: Regarding one large allocation - I'll check that. I agree that it may not bring any benefit, then I'll switch to allocating each txholder separately.

embedded/store/txpool.go
byo: Can you explain more?

I was thinking about adding more descriptive error messages - like an information what pool was exhausted to point out what configuration option should be tuned up to avoid this happening - did you also mean some more descriptive errors here?

pkg/database/reference.go
byo: Actually this was moved up on purpose. I wanted to minimize the risk of a scenario where a successful write is made but the whole operation fails because the read tx holder pool was exhausted. In such case the user would get the max concurrency limit exceeded error and could retry the write in case of such error.

pkg/database/sorted_set.go
byo: smae comment as in case of references - this is to avoid hard-to-interpret errors on high load

pkg/database/sql.go
byo: Fixed in 92492aa9.

embedded/store/txpool.go
byo: Made members private in 8d479635

embedded/store/txpool.go
byo: I did performance comparision between two implementations, there are no differences discovered thus I've switched to separate allocation for each entry in 03dcbcc1.

Zrzut ekranu z 2022-07-29 11-12-01

embedded/store/txpool.go
jeroiraz: I meant,  the error from the perspective of the pool is that there is no more available objects (tx), it may be due to concurrency or some other reason. From the store, we know it's due to a max concurrency limit reached

jeroiraz (dismissed):

embedded/store/txpool.go
byo: Changed to a dedicated error in 1cca68da - that one is mapped to more specific error in immustore or database

cmd/immuadmin/command/database.go
byo: I'd leave it for future changes - this also affects names in the GRPC protocol

PR #1312, opened 2022-08-09
test(embedded/ahtree): Update tests

byo:
Convert Edge Cases test to test suite and run each test independently.

### DIFF
--- a/embedded/ahtree/ahtree_test.go
+++ b/embedded/ahtree/ahtree_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/codenotary/immudb/embedded/appendable/multiapp"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


jeroiraz (dismissed):
LGTM. I'm wondering if with this changes the unit tests are not assuming any ordering, which would be the ideal case. If not, we may defer a restore of the state changes in the struct of the suite on each test case?

byo:
> LGTM. I'm wondering if with this changes the unit tests are not assuming any ordering, which would be the ideal case. If not, we may defer a restore of the state changes in the struct of the suite on each test case?

With this change I actually had to break those dependencies between steps - with them it was really hard to update tests separately. Sometimes this required copying some code from earlier stages to start in a correct state.

PR #1314, opened 2022-08-10
fix(verification): Restrict the length of verifiable linear proof

byo:
Since the main Merkle Tree is calculated synchronously, the
linear proof can contain up to one link.

### DIFF
--- a/embedded/store/verification.go
+++ b/embedded/store/verification.go
@@ -118,6 +118,17 @@ func VerifyDualProof(proof *DualProof, sourceTxID, targetTxID uint64, sourceAlh,
 		}
 	}
 
+	if len(proof.LinearProof.Terms) > 2 {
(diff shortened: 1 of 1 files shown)


jeroiraz:
We have already discussed it offline, we may hard restrict to sync append into the hash tree, even in the embedded store. Keeping the possibility to implement the additional validations in the proofs or its definitely removal

embedded/store/verification.go
```diff
@@ -118,6 +118,17 @@ func VerifyDualProof(proof *DualProof, sourceTxID, targetTxID uint64, sourceAlh,
 		}
 	}
 
+	if len(proof.LinearProof.Terms) > 2 {
```
arriqaaq, line 121: can we also add a test case around this? Just to confirm

PR #1315, opened 2022-08-10
Fix/bltx ordering

byo:


embedded/store/immustore.go
```diff
@@ -1118,7 +1118,6 @@ func (s *ImmuStore) precommit(otx *OngoingTx, expectedHeader *TxHeader, waitForI
 
 	if expectedHeader == nil {
 		ts = s.timeFunc().Unix()
-		blTxID = s.aht.Size()
```
jeroiraz, line 1121: it may be better to leave only the header version out of the critical section. Moving timestamp and validations during replicated commit inside it, to prevent potential edge cases when switching between master and replica modes. Please take a look at the commit in branch `fix/bltx-ordering-rep` as reference

jeroiraz: or situation where the same tx is attempted to be replicated multiple times at the same time. The disadvantage is that more garbage in payload logs may be generated in such situations

byo: I have moved the timestamp calculation to the locked section - also improved the test to ensure timestamps are ordered.

Also added check for simultaneous replicators - currently the check is done both before the mutex is held (like checking if we have the data or that transaction hash is correct) and while the mutex is locked (final check to ensure there was no additional TX in the meantime).

embedded/store/immustore.go
```diff
@@ -1207,6 +1207,15 @@ func (s *ImmuStore) precommit(otx *OngoingTx, expectedHeader *TxHeader, waitForI
...
 		ts = expectedHeader.Ts
 		blTxID = expectedHeader.BlTxID
+
+		// currTxID and currAlh were already checked before,
+		// but we have to add an additional check once the commit mutex
+		// is locked to ensure that those constraints are still valid
+		// in case of simultaneous writers
+		currTxID := s.lastCommittedTxID()
+		if currTxID != expectedHeader.ID-1 {
+			return nil, ErrTxReadConflict
```
jeroiraz, line 1217: I'd return a different error, such as TxAlreadyCommitted maybe...

jeroiraz (dismissed):

PR #1316, opened 2022-08-10
chore(pkg/logger): add json logger

arriqaaq:
adds json logger to output log messages in json format


pkg/logger/json.go
```diff
@@ -0,0 +1,249 @@
...
+	// errInvalidTypeMsg message implies an arg cannot be serialized to json
+	errInvalidTypeMsg = "cannot serialize arg(s) to json"
+
+	// defaultCallerOffset is the stack frame offset in the call stack for the caller
+	defaultCallerOffset = 3
+)
+
+var (
+	// DefaultOutput is used as the default log output.
+	DefaultOutput io.Writer = os.Stderr
```
jeroiraz, line 44: I'd use global variables only when they're meant to be externally set at compile time... otherwise it might lead to unexpected changes...

pkg/logger/json.go
```diff
@@ -0,0 +1,249 @@
...
+	l.level = level
+}
+
+// Name returns the loggers name
+func (i *jsonLogger) Name() string {
+	return i.name
+}
+
+// CloneWithLevel returns a shallow copy of the jsonLogger
+func (l *jsonLogger) CloneWithLevel(level LogLevel) Logger {
```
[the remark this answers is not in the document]
arriqaaq: I have to fix this so that it clones a new writer. I was wondering if the method CloneWithLevel is required to be part of the interface? Personally cloning seems helpful when we are deriving child loggers or something similar, but I am not sure how it would be used in the current way

pkg/logger/json.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+// Errorf prints the message and args at ERROR level
+func (l *jsonLogger) Errorf(msg string, args ...interface{}) {
+	l.log(l.Name(), LogError, msg, args...)
+}
+
+// Update the logging level
+func (l *jsonLogger) SetLogLevel(level LogLevel) {
+	l.level = level
```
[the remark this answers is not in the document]
arriqaaq: Thank you for catching this, wanted an early review on this :)

You are right, I will fix this

pkg/logger/json_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+		b := buf.Bytes()
+
+		var raw map[string]interface{}
+		if err := json.Unmarshal(b, &raw); err != nil {
+			t.Fatal(err)
+		}
+
+		assert.Equal(t, "test call", raw["message"])
+		assert.Equal(t, "foo", raw["user"])
```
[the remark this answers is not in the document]
arriqaaq: I agree with you. Since most of the golang loggers are built to log a message, and then KV pairs, I was confused whether I should add it as Info rather than Infof. I am thinking I'll keep the Infof to log a formatted message, and change the existing function to a Info, best of both worlds

pkg/logger/logger.go
```diff
@@ -55,3 +57,32 @@ func logLevelFromEnvironment() LogLevel {
...
+		// Name of the subsystem to prefix logs with
+		Name string
+
+		// The threshold for the logger. Anything less severe is supressed
+		Level LogLevel
+
+		// Where to write the logs to. Defaults to os.Stderr if nil
+		Output io.Writer
+
+		Mutex Locker
```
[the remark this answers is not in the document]
arriqaaq: Oh I thought of adding a mutex as an option, to abstract it out. But this makes no sense and will use default mutex itself

pkg/logger/json.go
arriqaaq: This is not being set anywhere I was thinking this could be useful if a user using this library wants to use a set of preinitialised defaults. Wdyt?

pkg/logger/json_test.go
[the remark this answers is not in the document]
byo: At this point I believe we're only interested in a set of predefined parameters for the log (like module name, call point, log level).
Currently all arguments passed to the `logger.Infof(...)` are only about formatting the message thus I think we will have to leave it that way to not break existing code.

I was also thinking about marking some of those arguments as named fields, e.g.:

```go
logger.Infof("test call from %s to method %s", logger.named("from", "user"), "foo")
```

or some extra builder-like approach (we're using this in TrustCenter backend):

```go
logger.Arg("name", "value").Arg("name2", 46).Infof("test call from %s to method %s", "user", "foo")
```

but that's a much a much larger code change.

Maybe let's just start with a backwards-compatible change without user-specified extra arguments?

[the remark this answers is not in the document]
arriqaaq: I agree with you @byo , I have made the Infof backward compatible, but also kept Info to log args, similar to how [zap](https://github.com/uber-go/zap) does

```
sugar.Infow("failed to fetch URL",
  // Structured context as loosely typed key-value pairs.
  "url", url,
  "attempt", 3,
  "backoff", time.Second,
)
sugar.Infof("Failed to fetch URL: %s", url)
```

pkg/logger/json.go
```diff
@@ -0,0 +1,269 @@
...
+var _ Logger = &JsonLogger{}
+
+// JsonLogger is a logger implementation for json logging.
+type JsonLogger struct {
+	name       string
+	timeFormat string
+
+	timeFnc TimeFunc
+
+	mutex  Locker
```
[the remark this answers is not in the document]
arriqaaq: yes, makes sense, I'll do that

pkg/logger/json.go
```diff
@@ -0,0 +1,269 @@
...
+	writer *writer
+
+	level *int32
```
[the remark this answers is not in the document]
arriqaaq: I had something in mind on doing nil checks if a level wasn't assigned, hence the pointer directly, but yeah, will keep it simple for now

pkg/logger/json.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	if opts.TimeFnc != nil {
+		l.timeFnc = opts.TimeFnc
+	}
+	if opts.TimeFormat != "" {
+		l.timeFormat = opts.TimeFormat
+	}
+
+	atomic.StoreInt32(l.level, int32(opts.Level))
```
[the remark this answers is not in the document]
arriqaaq: ack

pkg/logger/json.go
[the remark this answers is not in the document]
arriqaaq: @jeroiraz do we need CloneWithLevel in the interface contract?  Or can it be removed for now to keep it simple?

pkg/logger/json.go
```diff
@@ -0,0 +1,269 @@
...
+	if minLevel > level {
+		return
+	}
+
+	t := l.timeFnc()
+
+	l.mutex.Lock()
+	defer l.mutex.Unlock()
+
+	l.logJSON(t, name, level, msg, args...)
```
arriqaaq, line 112: @byo @jeroiraz I think eventually we can have a single logger (with json/plaintext as output), and can use the opts to configure which one to use. That way we wouldn't have to maintain many loggers. 

There can be another function, `logPlain` which handles memory/simple loggers. And file logger is just writing to another writer, and hence I've abstracted out in the struct to the output configurable.

This can be set for another story, but it will simplify the logger pkg

pkg/logger/json.go
arriqaaq: made this variable private

pkg/logger/json.go
arriqaaq: https://github.com/codenotary/immudb-issues/issues/80

cmd/immudb/command/root.go
```diff
@@ -59,24 +59,35 @@ Setting the logging level and other options through environment variables:
...
-		if options.Logfile != "" {
-			if flogger, file, err := logger.NewFileLogger("immudb ", options.Logfile); err == nil {
-				defer file.Close()
-				immudbServer.WithLogger(flogger)
-			} else {
-				c.QuitToStdErr(err)
-			}
+
+		// initialize logger for immudb
+		flogger, file, err := logger.NewLogger(&logger.Options{
```
jeroiraz, line 70: I'm wondering if it's safe to remove the conditional `if options.Logfile != "" {`

pkg/logger/json.go
```diff
@@ -0,0 +1,274 @@
...
+// NewJSONFileLogger returns a json logger which outputs to file
+func NewJSONFileLogger(name string, file string) (logger Logger, out *os.File, err error) {
+	if out, err = setup(file); err != nil {
+		return nil, nil, err
+	}
+	return &JsonLogger{
+		name:       name,
+		timeFormat: DefaultTimeFormat,
+		timeFnc:    time.Now,
+		mutex:      new(sync.Mutex),
```
jeroiraz, line 57:
```suggestion
```

It is not necessary to explicitly initialize sync.Mutex

pkg/logger/json.go
```diff
@@ -0,0 +1,274 @@
...
+		writer:     newWriter(out),
+		level:      int32(logLevelFromEnvironment()),
+	}, out, nil
+}
+
+// NewJSONLogger returns a json logger.
+func NewJSONLogger(opts *Options) *JsonLogger {
```
jeroiraz, line 64: opts could even be a value of type Options

pkg/logger/json.go
```diff
@@ -0,0 +1,274 @@
...
+	output := opts.Output
+	if output == nil {
+		output = defaultOutput
+	}
+
+	l := &JsonLogger{
+		name:       opts.Name,
+		timeFormat: DefaultTimeFormat,
+		timeFnc:    time.Now,
+		mutex:      new(sync.Mutex),
```
jeroiraz, line 78: It is not necessary to explicitly initialize sync.Mutex

cmd/immudb/command/root.go
arriqaaq: I've added that check inside this factory, `NewLogger` because there were too many ifs. If there is a json logger without file, there was a need to add an if else, and then a json logger with file would have a separate condition.

The problem is that the way immudb server is initialised, it sets the `NewSimpleLogger` by default in the `DefaultServer`. I see your point, the flogger can reset the logger even though it is defined, but I wanted to move to a single place where the logger is defined. WDYT?

pkg/logger/json.go
```diff
@@ -0,0 +1,274 @@
...
+		l.timeFormat = opts.TimeFormat
+	}
+
+	l.level = int32(opts.Level)
+	return l
+}
+
+var _ Logger = &JsonLogger{}
+
+// JsonLogger is a logger implementation for json logging.
```
jeroiraz, line 95: probably it's simpler to read the source code if the data type is at the top of the file before the methods?

pkg/logger/json.go
arriqaaq: I did not quite understand...

arriqaaq: You mean, pass by value than reference?

pkg/logger/json.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+// Log a message and a set of key/value pairs for a given level.
+func (l *JsonLogger) log(name string, level LogLevel, msg string, args ...interface{}) {
+	minLevel := LogLevel(atomic.LoadInt32(&l.level))
+	if level < minLevel {
+		return
+	}
+
+	t := l.timeFnc()
```
jeroiraz, line 120: time inside the critical section so to avoid (in some cases) older logs after newer ones?

pkg/logger/json.go
arriqaaq: ack

pkg/logger/logger.go
```diff
@@ -55,3 +56,47 @@ func logLevelFromEnvironment() LogLevel {
...
+		LogFormat string
+
+		// The file to write to.
+		LogFile string
+	}
+)
+
+// NewLogger is a factory for selecting a logger based on options
+func NewLogger(opts *Options) (logger Logger, out *os.File, err error) {
+	switch opts.LogFormat {
```
jeroiraz, line 90: it may be worth validating the logFormat is  json or plain text or empty. If an invalid data format is specified an error is raised

pkg/logger/logger.go
```diff
@@ -55,3 +56,47 @@ func logLevelFromEnvironment() LogLevel {
...
+		// The format in which logs will be formatted. (eg: text/json)
+		LogFormat string
+
+		// The file to write to.
+		LogFile string
+	}
+)
+
+// NewLogger is a factory for selecting a logger based on options
+func NewLogger(opts *Options) (logger Logger, out *os.File, err error) {
```
jeroiraz, line 89: opts could be of type Options, it avoids checking for null pointers

cmd/immudb/command/root.go
jeroiraz, line 70: problably the `logger` should be closeable and the file (if any) is closed when calling `Close` method.
```
logger, err := logger.NewLogger(...)
if err != nil {
  ...
}
defer logger.Close() // file is closed under the hood, if any
```

jeroiraz (commented):
check following logging line is also considered:

`You can now use immuadmin and immuclient CLIs to login with the immudb superadmin user and start using immudb.`

cmd/immudb/command/root.go
arriqaaq: Yes, I was thinking of adding it to the interface, this would help in avoiding this check of getting the file back and closing it

pkg/logger/logger.go
arriqaaq: ack

pkg/logger/logger.go
arriqaaq: Depends really, can use the null check to use default options in the future, and to avoid any default struct values for implicit data types (string = "", int = 0 , bool = false etc)

pkg/logger/json.go
arriqaaq: fixed

pkg/logger/json.go
arriqaaq: ack

cmd/immudb/command/root.go
arriqaaq: added the Close method to Logger interface

pkg/logger/json.go
arriqaaq: fixed

pkg/logger/logger.go
jeroiraz: fair point, in other parts of the codebase there are defaultOptions which can be modified to customize them

arriqaaq: ack

pkg/logger/json.go
```diff
@@ -0,0 +1,249 @@
...
+	sl.level = l.level
+	return &sl
+}
+
+type writer struct {
+	b bytes.Buffer
+	w io.Writer
+}
+
+func newWriter(w io.Writer) *writer {
```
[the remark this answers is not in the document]
arriqaaq: You are actually right, it would be much better. Since we have a file writer and I need the io writer specifically to close the file, I am keeping it different (the buffer isn't being used in case of json since the output is written to the io.Writer directly). But I will use the bufio.NewWriter in the later PR once I turn this to an interface, and unify the logger pkg

cmd/immudb/command/init.go
```diff
@@ -37,6 +37,8 @@ func (cl *Commandline) setupFlags(cmd *cobra.Command, options *server.Options) {
 	cmd.PersistentFlags().StringVar(&cl.config.CfgFn, "config", "", "config file (default path are configs or $HOME. Default filename is immudb.toml)")
 	cmd.Flags().String("pidfile", options.Pidfile, "pid path with filename. E.g. /var/run/immudb.pid")
 	cmd.Flags().String("logfile", options.Logfile, "log path with filename. E.g. /tmp/immudb/immudb.log")
+	cmd.Flags().String("logformat", options.LogFormat, "log format. E.g. text/json")
```
jeroiraz, line 40: what about "log format e.g. text/json" ?
If so, we may update the message in other flags

pkg/logger/file.go
```diff
@@ -27,30 +27,34 @@ import (
...
-	if out, err = setup(file); err != nil {
+	out, err = setup(file)
+	if err != nil {
 		return nil, nil, err
 	}
 	logger = &FileLogger{
+		out:      out,
 		Logger:   log.New(out, name, log.LstdFlags),
-		LogLevel: logLevelFromEnvironment(),
+		LogLevel: LogLevelFromEnvironment(),
```
jeroiraz, line 42: isn't the logLevel already specified in `logger.NewLogger(&logger.Options{...})`

pkg/logger/json.go
```diff
@@ -0,0 +1,283 @@
...
+	// defaultCallerOffset is the stack frame offset in the call stack for the caller
+	defaultCallerOffset = 4
+)
+
+var (
+	// defaultOutput is used as the default log output.
+	defaultOutput io.Writer = os.Stderr
+)
+
+var _ Logger = &JsonLogger{}
```
jeroiraz, line 48:
```suggestion
var _ Logger = (*JsonLogger)(nil)
```

pkg/logger/json.go
```diff
@@ -0,0 +1,283 @@
...
+	if opts == nil {
+		opts = &Options{}
+	}
+
+	output := opts.Output
+	if output == nil {
+		output = defaultOutput
+	}
+
+	if opts.LogFile != "" {
```
jeroiraz, line 74: it's not fully clear the relationship between opts.Output and opts.LogFile...

pkg/logger/json.go
```diff
@@ -0,0 +1,283 @@
...
+	w io.Writer
+}
+
+func newWriter(w io.Writer) *writer {
+	return &writer{w: w}
+}
+
+func (w *writer) Flush() (err error) {
+	var unwritten = w.b.Bytes()
+	_, err = w.w.Write(unwritten)
```
jeroiraz, line 263: using the same naming for the internal element might lead to infinite loops when making a mistake i.e. I'd recommend assigning a different name either for the outer or inner object so to avoid `w.w.Write(...)`

cmd/immudb/command/init.go
arriqaaq: ack

pkg/logger/json.go
arriqaaq: ack

pkg/logger/file.go
arriqaaq: the plan is to resolve this in another PR for unified logger, the factory methods do not accept options as a parameter, and hence why this is initialised as such. Tracking it here

https://github.com/codenotary/immudb-issues/issues/85

pkg/logger/json.go
arriqaaq: ack

pkg/logger/json.go
arriqaaq: So the Logfile is the filename in which the log is written to. Output is just any direct interface to be written to. Again, this logfile parameter is to provide backward compatibility with existing factory methods

pkg/logger/json.go
```diff
@@ -0,0 +1,283 @@
...
+	vals := map[string]interface{}{
+		"message":   msg,
+		"timestamp": t.Format(l.timeFormat),
+	}
+
+	if name != "" {
+		vals["module"] = name
+	}
+
+	var levelStr string
```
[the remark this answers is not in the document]
arriqaaq: ack, have made this change in the other PR, made it here too

https://github.com/codenotary/immudb/pull/1325/files

pkg/logger/json.go
```diff
@@ -0,0 +1,283 @@
...
+		levelStr = "info"
+	case LogDebug:
+		levelStr = "debug"
+	default:
+		levelStr = "all"
+	}
+
+	vals["level"] = levelStr
+
+	if pc, file, line, ok := runtime.Caller(defaultCallerOffset + 1); ok {
```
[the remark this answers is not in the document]
arriqaaq: ack

pkg/server/server.go
```diff
@@ -76,9 +76,9 @@ var immudbTextLogo = " _                               _ _     \n" +
 
 // Initialize initializes dependencies, set up multi database capabilities and stats
 func (s *ImmuServer) Initialize() error {
-	s.Logger.Infof("%s\n%s\n%s\n\n", immudbTextLogo, version.VersionStr(), s.Options)
```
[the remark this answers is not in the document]
arriqaaq: I don't understand this well, but can we create an issue to track this separately? Because I don't see it as a blocker for adding json log support, if it is, you can lemme know.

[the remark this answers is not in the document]
arriqaaq: ok based on our discussion, I've added some steps, please verify for the same

pkg/server/server.go
```diff
@@ -76,9 +76,9 @@ var immudbTextLogo = " _                               _ _     \n" +
...
-
-	if s.Options.Logfile != "" {
+	if s.Options.IsJSONLogger() {
```
arriqaaq, line 79: @jeroiraz does this satisfy the condition which you were mentioning?

- json format into stdout, messages are not printed (logo and CLIs messages)
- text format into stdout, messages are printed into stdout

jeroiraz: I think so. The same logic may be applied in the logo

arriqaaq: The logo is only printed in case of stdout to file

jeroiraz (approved):

PR #1320, opened 2022-08-15
appendable: retryable sync

jeroiraz:
This PR implements retryable sync on Appendable implementation.

When durability needs to be ensured, buffered data is not released until data is successfully synced on disk.


embedded/ahtree/ahtree.go
```diff
@@ -775,7 +768,17 @@ func (t *AHtree) sync() error {
 		return nil
 	}
 
-	err := t.pLog.Sync()
+	err := t.pLog.Flush()
```
arriqaaq, line 771: WDYT of combining flush and sync into another method on the appendable interface? Like `FSync` or anything else? Will just make it easier to use in future cases where a person wants to commit, preserving durability

embedded/appendable/singleapp/single_app.go
```diff
@@ -362,18 +378,75 @@ func (aof *AppendableFile) Append(bs []byte) (off int64, n int, err error) {
...
+
+	return off, n + 4, err
+}
+
+func (aof *AppendableFile) write(bs []byte) (n int, err error) {
+	for n < len(bs) {
+		available := len(aof.writeBuffer) - aof.wbufUnwrittenOffset
+
+		if available == 0 {
+			if aof.retryableSync && !aof.autoSync {
```
arriqaaq, line 396: wdyt of making this as follows

```
	if aof.retryableSync{
		if !aof.autoSync {
			// Sync must be called to free buffer space
			return n, ErrBufferFull
		} else {
			err = aof.sync()
			if err != nil {
				return
			}	
		}
	} else {
		err = aof.flush()
		if err != nil {
			return
		}
	}

```

Also I am a bit unclear on what retryableSync does. I am assuming it forces flush and then sync? If so, wdyt about adding a small comment describing the purpose of it here?

arriqaaq: Ok my bad, just read the comment above in the struct definition of what retryableSync does. Pls ignore the comment line

embedded/appendable/singleapp/single_app.go
```diff
@@ -484,7 +558,28 @@ func (aof *AppendableFile) Sync() error {
...
+
+		// prevent data lost when fsync fails
+		// buffered data may be re-written in following
+		// flushing and syncing calls.
+		aof.fileOffset -= int64(aof.wbufFlushedOffset)
+		aof.wbufFlushedOffset = 0
+	}
+	if aof.retryableSync && err == nil {
+		// buffer space is freed
+		aof.wbufFlushedOffset = 0
```
arriqaaq, line 578: I see these two fields are being reset at a few different places, can we add that as a separate function

```
// Clear frees buffer space for reuse.
func (aof *A) Clear(){
		aof.wbufFlushedOffset = 0
		aof.wbufUnwrittenOffset = 0
}

```

embedded/ahtree/ahtree.go
jeroiraz: I though about even removing Flush method from Appendable interface, however it would require further changes and that's why I keep it as it was, at least in the scope of this changes

embedded/appendable/singleapp/single_app.go
jeroiraz: In general I prefer code readability, in the end compiler optimizations may be applied and in this type of changes performance shouldn't be affected. However, I made a slightly change there

arriqaaq: It was bit hard to grasp the logic with so many conditions, hence thought having a parental `if` would help. Subjective I guess, no blocker

embedded/ahtree/ahtree.go
arriqaaq: ack

arriqaaq (approved):

Issue #1288, opened 2022-07-20
Return empty columns when use SQL stdlib


winterpi:
**What happened**
- when use SQL stdlib to make a query, it returns [ ] for columns and rows when there is no data in the table.
- but if I use client.SQLquery( ), the result is OK.

**What you expected to happen**
the columns should not be [ ] when use stdlib.Query( )

**How to reproduce it (as minimally and precisely as possible)**
```
rows, err := db.QueryContext(ctx, "select * from emptytable")
cols :=rows.Columns()   //should not by empty
rows.next()
```

**Environment**
```shell
# run "immu* version" and copy/paste the output here
```

**Additional info (any other context about the problem)**
It's actually that the columns info has been ommitted in stdlib connection.go
```
		queryResult, err = c.tx.SQLQuery(ctx, query, vals)
		if err != nil {
			return nil, err
		}
		return &Rows{rows: queryResult.Rows}, nil
```

PR #1322, opened 2022-08-16
fix(stdlib/rows): add colums to row response

arriqaaq:
- adds columns from SQLQueryResult to row response
- columns from rows now fetched from columns field, not rows
- add tests for connection (Conn) with actual db connection
- add test for empty table for QueryContext in a connection

References: #1288


pkg/stdlib/rows_test.go
```diff
@@ -35,6 +35,11 @@ func TestRows(t *testing.T) {
...
 			Values:  []*schema.SQLValue{{Value: nil}},
 		}},
+		columns: []*schema.Column{
+			{
+				Name: "c1",
+			},
+		},
 	}
 
 	ast := r.Columns()
```
jeroiraz, line 45: it would be good having a test case with empty rows and still returning the list of columns

pkg/stdlib/connection_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+	rows, err := c.QueryContext(ctx, "SELECT * FROM emptytable limit 1", nil)
+	require.NoError(t, err)
+	defer rows.Close()
+
+	cols := rows.Columns()
+	require.Equal(t, len(cols), 6)
+}
+
+/*func TestConn_Ping(t *testing.T) {
```
jeroiraz, line 194: is this test case not working?

pkg/stdlib/connection_test.go
```diff
@@ -0,0 +1,222 @@
+package stdlib
```
jeroiraz, line 1: please include the copyright notice

pkg/stdlib/rows_test.go
arriqaaq: ack

pkg/stdlib/connection_test.go
arriqaaq: no it doesn't, it was commented out before, so did not want to address the issue in this PR

jeroiraz (dismissed):

jeroiraz (approved):

PR #1323, opened 2022-08-16
fix(getRandomTable): increase RNG range for table generation

arriqaaq:
- fixes issue with flaky tests which fail due to table already exists error

References: [1324](https://github.com/codenotary/immudb/issues/1324)


jeroiraz (approved):
quick fix, at least it will occur with less probability...

arriqaaq:
Thank you :)

PR #1325, opened 2022-08-16
Unified logger package

arriqaaq:


arriqaaq:
Can work on this later, when refactoring logger

PR #1328, opened 2022-08-19
fix(embedded): use tmp folder for unit test cases

arriqaaq:


embedded/appendable/multiapp/multi_app_test.go
```diff
@@ -165,8 +171,11 @@ func TestMultiAppClosedFiles(t *testing.T) {
 }
 
 func TestMultiAppReOpening(t *testing.T) {
-	a, err := Open("testdata", DefaultOptions().WithFileSize(1))
-	defer os.RemoveAll("testdata")
+	path, err := ioutil.TempDir(os.TempDir(), "testdata")
+	require.NoError(t, err)
+	defer os.RemoveAll(path)
+
+	a, err := Open(path, DefaultOptions().WithFileSize(1))
```
jeroiraz, line 178: `testdata_copy` below should also be a temp folder

embedded/appendable/remoteapp/remote_app_test.go
```diff
@@ -835,8 +854,7 @@ func TestRemoteStorageDownloadUnrecoverableError(t *testing.T) {
 }
 
 func TestRemoteStorageOpenChunkWhenUploading(t *testing.T) {
-	defer os.RemoveAll("testdata")
-	prepareLocalTestFiles(t)
+	path := prepareLocalTestFiles(t)
```
jeroiraz, line 857:
```suggestion
	path := prepareLocalTestFiles(t)
	defer os.RemoveAll(path)
```

embedded/store/immustore_test.go
```diff
@@ -306,13 +306,13 @@ func TestImmudbStoreEdgeCases(t *testing.T) {
 		require.EqualError(t, err, "stat invalid\x00_dir_name: invalid argument")
 	})
 
-	t.Run("should fail with permiission denied", func(t *testing.T) {
```
jeroiraz, line 309: is this test case failing in the github runner? Recent changes shouldn't affect this edge case... However I was able to reproduce it locally when running test inside a container but with macOS as host

pkg/server/server_test.go
```diff
@@ -593,6 +593,7 @@ func TestServerLoaduserDatabase(t *testing.T) {
 }
 
 func TestServerLoadUserDatabases(t *testing.T) {
+	os.RemoveAll("./data_v1.1.0")
```
jeroiraz, line 596: a tmp folder could be used here as well

embedded/store/immustore_test.go
arriqaaq: this test is only failing on a mac, and also in ubuntu container. I will uncomment this and run in gitlab runner

embedded/appendable/remoteapp/remote_app_test.go
arriqaaq: ack

embedded/appendable/multiapp/multi_app_test.go
arriqaaq: ack

pkg/server/server_test.go
arriqaaq: for some reason it doesn't work as expected with tmp folder

PR #1337, opened 2022-08-23
chore(embedded/ahtree): improve validations and error handling

jeroiraz:

### DIFF
--- a/embedded/ahtree/ahtree.go
+++ b/embedded/ahtree/ahtree.go
@@ -19,6 +19,7 @@ import (
 	"crypto/sha256"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"math/bits"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


embedded/ahtree/ahtree.go
```diff
@@ -31,6 +32,7 @@ import (
 )
 
 var ErrIllegalArguments = errors.New("illegal arguments")
```
arriqaaq, line 34: wdyt about putting all these vars in a single block? 
```
var (

)
```

embedded/ahtree/ahtree.go
```diff
@@ -142,8 +145,13 @@ func Open(path string, opts *Options) (*AHtree, error) {
 }
 
 func OpenWith(pLog, dLog, cLog appendable.Appendable, opts *Options) (*AHtree, error) {
-	if !validOptions(opts) || pLog == nil || dLog == nil || cLog == nil {
-		return nil, ErrIllegalArguments
+	if pLog == nil || dLog == nil || cLog == nil {
+		return nil, fmt.Errorf("%w: invalid appendables", ErrIllegalArguments)
```
arriqaaq, line 149: may be you would want to apply the same format of error message here as used below for nil options?

```
		return fmt.Errorf("%w: nil options", ErrInvalidOptions)
		return fmt.Errorf("%w: nil appendable", ErrInvalidOptions) (something similar)

```

arriqaaq (dismissed):
minor changes, looks neat! LGTM

embedded/ahtree/ahtree.go
jeroiraz: I think it could be done, probably it another PR. Inline var declaration sometimes make it easier to lookup for error declaration without IDE support, just that

PR #1338, opened 2022-08-23
chore(embedded/appendable): improve validation and error handling

jeroiraz:


embedded/store/immustore.go
```diff
@@ -42,8 +43,9 @@ import (
 	"github.com/codenotary/immudb/pkg/logger"
 )
 
-var ErrIllegalArguments = errors.New("illegal arguments")
-var ErrAlreadyClosed = errors.New("already closed")
+var ErrIllegalArguments = embedded.ErrIllegalArguments
+var ErrInvalidOptions = fmt.Errorf("%w: invalid options", ErrIllegalArguments)
+var ErrAlreadyClosed = embedded.ErrAlreadyClosed
```
arriqaaq, line 48: do we need to redeclare this if it is already described in embedded? A single place to define all errors would be great, which you've done in embedded pkg. From there on, why don't we wrap the error using errors.Wrap? That way we can avoid lots of duplicates? wdyt?

arriqaaq: errors.Wrap is not a good idea as it is already being wrapped with errorf, I was just thinking along custom error lines

```
type myCustomError struct {
    errorMessage    string
    someRandomValue int
}

func (e *myCustomError) Error() string {
    return fmt.Sprintf("Message: %s - Random value: %d", e.errorMessage, e.someRandomValue)
}

func doAnotherThing() error {
    return &myCustomError{
        errorMessage:    "hello world",
        someRandomValue: 13,
    }
}

func doSomething() error {
    err := doAnotherThing()
    return fmt.Errorf("error doing something: %w", err)
}

func main() {
    err := doSomething()

    var customError *myCustomError
    if errors.As(err, &customError) {
        fmt.Printf("Custom error random value: %d\n", customError.someRandomValue)
    }

    fmt.Println(err)

    if errors.Is(err, myCustomError) {
        fmt.Println("Found error!")
    }
}
```

I dunno, maybe we need to revisit this later, but just a food for thought

arriqaaq (approved):

embedded/store/immustore.go
jeroiraz: the general error declaration was required to test comparing with `errors.Is(err, embedded.ErrIllegalArguments)` works even when the store wrapped the error e.g. adding more info to it.

Yes, I think error declaration should be in embedded package, also assigning an error code

PR #1345, opened 2022-08-24
appendable: unit test read/write using file and inmem buffering

jeroiraz:


SimoneLazzaris (approved):
My only comment is that maybe it's possible to merge code from `TestSingleAppSetOffsetWithRetryableSyncOff` and `TestSingleAppSetOffsetWithRetryableSyncOn` since the only differ in options setup.
But it's a matter of taste.
Apart from that LGTM

PR #1354, opened 2022-08-31
Support read replicas

fernandezcuesta:
- Support read replicas by deploying a separate statefulset of N-1 replicas
- Add additional ingresses to the read replica services
- Optionally install grafana dashboard via Grafana Operator CRD

**Test Plan:**

```
k3d cluster create -a3
helm upgrade --install immudb . --set replicaCount=3 --set ingress.enabled=true
kubectl port-forward svc/immudb-rr-grpc 2233:3322 &
kubectl port-forward svc/immudb-grpc 3322:3322 &

# This should succeed
immuclient --username immudb --password immudb exec "UPSERT INTO orders (id, customerid, productid) VALUES ( 2, 1, 1);"
# This should fail
immuclient --immudb-port 2233 --username immudb --password immudb exec "UPSERT INTO orders (id, customerid, productid) VALUES ( 2, 1, 1);"
```

### DIFF
--- a/helm/templates/dashboard.yaml
+++ b/helm/templates/dashboard.yaml
@@ -6,6 +6,6 @@ metadata:
   labels: 
     app: immudb
 spec:
-  url: https://raw.githubusercontent.com/codenotary/immudb/master/tools/monitoring/grafana-dashboard.json
+  url: https://raw.githubusercontent.com/codenotary/immudb/{{- .Values.grafanaDashboard.version | default (printf "v%s" .Chart.AppVersion) -}}/tools/monitoring/grafana-dashboard.json
 {{- end -}}
 
--- a/helm/templates/dashboard.yaml
+++ b/helm/templates/dashboard.yaml
@@ -1,0 +1,11 @@
+{{- if .Values.grafanaDashboard -}}
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata: 
+  name: immudb-grafana-dashboard
+  labels: 
+    app: immudb
+spec:
+  url: https://raw.githubusercontent.com/codenotary/immudb/master/tools/monitoring/grafana-dashboard.json
+{{- end -}}
+
--- a/helm/templates/ingress.yaml
+++ b/helm/templates/ingress.yaml
@@ -1,52 +1,61 @@
 {{- if .Values.ingress.enabled -}}
+{{- $18orHigher := semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion }}
+{{- $19orHigher := semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion }}
 {{- $fullName := include "immudb.fullname" . -}}
-{{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion)) }}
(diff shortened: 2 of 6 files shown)


helm/templates/dashboard.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.grafanaDashboard -}}
```
byo, line 1: Let's make this `.Values.grafanaDashboard.enabled` - that way the `.Values.grafanaDashboard` object can be easily extended with other options in backwards-compatible way.

helm/templates/dashboard.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata: 
+  name: immudb-grafana-dashboard
+  labels: 
+    app: immudb
+spec:
+  url: https://raw.githubusercontent.com/codenotary/immudb/master/tools/monitoring/grafana-dashboard.json
```
byo, line 9: We could use URL for specific immudb version here, e.g. `https://raw.githubusercontent.com/codenotary/immudb/v1.3.2/tools/monitoring/grafana-dashboard.json`

that could default to the same version as the immudb one that is used in the statefulset with an option to override, e.g.:

```suggestion
  url: https://raw.githubusercontent.com/codenotary/immudb/{{- .Values.grafanaDashboard.version | default print "v" .Chart.AppVersion -}}/tools/monitoring/grafana-dashboard.json
```

helm/templates/statefulset.yaml
```diff
@@ -1,43 +1,43 @@
...
     spec:
-      {{- with .Values.imagePullSecrets }}
+      {{- with $.Values.imagePullSecrets }}
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      securityContext:
-        {{- toYaml .Values.podSecurityContext | nindent 8 }}
-      volumes:
-      - name: immudb-storage
```
byo, line 32: I see the volume removed but where it has been moved to? I can't find the place.

helm/templates/ingress.yaml
```diff
@@ -1,52 +1,61 @@
...
-  name: {{ $fullName }}-http
+  name: {{ . }}-http
   labels:
-    {{- include "immudb.labels" . | nindent 4 }}
-  {{- with .Values.ingress.annotations }}
+    {{- include "immudb.labels" $ | nindent 4 }}
+  {{- with $.Values.ingress.annotations }}
   annotations:
-    {{- toYaml . | nindent 4 }}
+    {{- toYaml $ | nindent 4 }}
```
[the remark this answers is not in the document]
fernandezcuesta: `.` here refers to the `range` loop, whereas `$` refers to the context out from the loop

helm/templates/statefulset.yaml
fernandezcuesta: If I'm not mistaken, that's not needed (see example [here](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#components))

SimoneLazzaris: It is not needed *if* you provide a volumeClaimTemplate. You must provide one when using more than one replicas

jeroiraz:
@SimoneLazzaris could you review this changes?

PR #1355, opened 2022-08-31
Feat/changable db list in server

byo:


pkg/server/types.go
```diff
@@ -108,6 +108,7 @@ type ImmuServerIf interface {
 	WithStateSigner(stateSigner StateSigner) ImmuServerIf
 	WithStreamServiceFactory(ssf stream.ServiceFactory) ImmuServerIf
 	WithPgsqlServer(psrv pgsqlsrv.Server) ImmuServerIf
+	WithDbList(dbList database.DatabaseList) ImmuServerIf
```
jeroiraz, line 111: how databases are loaded may also be extracted, that way it would be possible to run the server totally independent from were they are stored (or even not stored)

jeroiraz (approved):

PR #1356, opened 2022-09-02
chore(Dockerfile): add fips build changes

arriqaaq:
- add Dockerfile for fips build
- add cmd/immudb/fips folder to enable fips check and immudb build
- add Makefile changes for generating FIPS build with cgo enabled

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -22,6 +22,10 @@ jobs:
             testWithMinio: true
 
           - os: ubuntu-latest
+            go: "1.18"
+            testWithFips: true
+
+          - os: ubuntu-latest
             go: "1.19"
             test: true
 
@@ -66,6 +70,11 @@ jobs:
           # Stop minio
           docker rm -f minio
         if: matrix.testWithMinio
+
+      - name: Test (with fips build)
+        run: |
+          make test/fips
+        if: matrix.testWithFips
 
       - name: Test Client
         run: make test-client
--- a/Makefile
+++ b/Makefile
@@ -32,13 +32,18 @@ V_COMMIT := $(shell git rev-parse HEAD)
 #V_BUILT_BY := "$(shell echo "`git config user.name`<`git config user.email`>")"
 V_BUILT_BY := $(shell git config user.email)
 V_BUILT_AT := $(shell date +%s)
-V_LDFLAGS_COMMON := -s -X "github.com/codenotary/immudb/cmd/version.Version=${VERSION}" \
+V_LDFLAGS_SYMBOL := -s
+V_LDFLAGS_BUILD := -X "github.com/codenotary/immudb/cmd/version.Version=${VERSION}" \
 					-X "github.com/codenotary/immudb/cmd/version.Commit=${V_COMMIT}" \
 					-X "github.com/codenotary/immudb/cmd/version.BuiltBy=${V_BUILT_BY}"\
 					-X "github.com/codenotary/immudb/cmd/version.BuiltAt=${V_BUILT_AT}"
+V_LDFLAGS_COMMON := ${V_LDFLAGS_SYMBOL} ${V_LDFLAGS_BUILD}
 V_LDFLAGS_STATIC := ${V_LDFLAGS_COMMON} \
 				  -X github.com/codenotary/immudb/cmd/version.Static=static \
 				  -extldflags "-static"
+V_LDFLAGS_FIPS_BUILD = ${V_LDFLAGS_BUILD} \
+				  -X github.com/codenotary/immudb/cmd/version.FIPSEnabled=true
+
 GRPC_GATEWAY_VERSION := $(shell go list -m -versions github.com/grpc-ecosystem/grpc-gateway | awk -F ' ' '{print $$NF}')
 ifdef WEBCONSOLE
 IMMUDB_BUILD_TAGS=-tags webconsole
@@ -83,19 +88,34 @@ immutest:
 
 .PHONY: immuclient-static
 immuclient-static:
-	CGO_ENABLED=0 $(GO) build -a -ldflags '$(V_LDFLAGS_STATIC) -extldflags  "-static"' ./cmd/immuclient
+	CGO_ENABLED=0 $(GO) build -a -ldflags '$(V_LDFLAGS_STATIC)' ./cmd/immuclient
+
+.PHONY: immuclient-fips
+immuclient-fips:
+	CGO_ENABLED=1 GOOS=linux GOARCH=amd64 $(GO) build -tags=fips -a -o immuclient -ldflags '$(V_LDFLAGS_FIPS_BUILD)' ./cmd/immuclient/fips
+	./build/fips/check-fips.sh immuclient
 
 .PHONY: immuadmin-static
 immuadmin-static:
-	CGO_ENABLED=0 $(GO) build -a -ldflags '$(V_LDFLAGS_STATIC) -extldflags "-static"' ./cmd/immuadmin
+	CGO_ENABLED=0 $(GO) build -a -ldflags '$(V_LDFLAGS_STATIC)' ./cmd/immuadmin
+
+.PHONY: immuadmin-fips
+immuadmin-fips:
+	CGO_ENABLED=1 GOOS=linux GOARCH=amd64 $(GO) build -tags=fips -a -o immuadmin -ldflags '$(V_LDFLAGS_FIPS_BUILD)' ./cmd/immuadmin/fips
+	./build/fips/check-fips.sh immuadmin
 
 .PHONY: immudb-static
 immudb-static: webconsole
-	CGO_ENABLED=0 $(GO) build $(IMMUDB_BUILD_TAGS) -a -ldflags '$(V_LDFLAGS_STATIC) -extldflags "-static"' ./cmd/immudb
+	CGO_ENABLED=0 $(GO) build $(IMMUDB_BUILD_TAGS) -a -ldflags '$(V_LDFLAGS_STATIC)' ./cmd/immudb
+
+.PHONY: immudb-fips
+immudb-fips: webconsole
+	CGO_ENABLED=1 GOOS=linux GOARCH=amd64 WEBCONSOLE=default $(GO) build -tags=webconsole,fips -a -o immudb -ldflags '$(V_LDFLAGS_FIPS_BUILD)' ./cmd/immudb/fips
+	./build/fips/check-fips.sh immudb
 
 .PHONY: immutest-static
 immutest-static:
-	CGO_ENABLED=0 $(GO) build -a -ldflags '$(V_LDFLAGS_STATIC) -extldflags "-static"' ./cmd/immutest
+	CGO_ENABLED=0 $(GO) build -a -ldflags '$(V_LDFLAGS_STATIC)' ./cmd/immutest
 
 .PHONY: vendor
 vendor:
@@ -105,6 +125,12 @@ vendor:
 test:
 	$(GO) vet ./...
 	LOG_LEVEL=error $(GO) test -v -failfast ./... ${GO_TEST_FLAGS}
+
+# build FIPS binary from docker image
+.PHONY: test/fips
+test/fips:
+	docker build -t fips:test-build -f build/fips/Dockerfile.build .
+	docker run --rm fips:test-build -c "make test"
 
 .PHONY: test-client
 test-client:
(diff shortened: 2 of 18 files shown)


.github/workflows/pull.yml
```diff
@@ -24,6 +24,7 @@ jobs:
           - os: ubuntu-latest
             go: "1.19"
             test: true
+            testWithFips: true
```
byo, line 27: The main version of the compiler we're building at now is 1.18 - shouldn't this test be moved to 1.18 ubunutu-latest?

Also it would be good to perform the test with both fips-enabled and iwthout-fips code by cloning the matrix entry.

build/fips/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM us-docker.pkg.dev/google.com/api-project-999119582588/go-boringcrypto/golang:1.18.5b7 as build-fips
```
byo, line 1: Please add a short comment to this dockerfile - some people may get confused that this is used to build immudb docker image.

check-fips.sh
```diff
@@ -0,0 +1,10 @@
+# Pass the path to the executable to check for FIPS compliance
```
byo, line 1: We should move this file to `build/fips` folder.

build/fips/Dockerfile
arriqaaq: ack, I will do that

.github/workflows/pull.yml
arriqaaq: actually it doesn't matter since the fips test run inside a docker container. I can move it to 1.18 though

check-fips.sh
arriqaaq: ack

.github/workflows/pull.yml
arriqaaq: `Also it would be good to perform the test with both fips-enabled and iwthout-fips code by cloning the matrix entry.`

Could you let me know more about this @byo ? Because I think these two lines does that right

```
            test: true
            testWithFips: true
```

You'd want me to do it for 1.18 too?

byo: Ah, OK, didn't get that it does two separate steps now.

What I would suggest then is to actually add a separate built step that only has the `testWithFips` set without neither `test` nor `testWithMinio` set to true so that it is easier to understand.

Dockerfile.fips
```diff
@@ -0,0 +1,52 @@
+# This version of Go is a Go+BoringCrypto release for FIPS 140-2 compliance
```
byo, line 1: This file may need to go to `build` folder once https://github.com/codenotary/immudb/pull/1366 gets merged

Dockerfile.fips
```diff
@@ -0,0 +1,52 @@
...
+LABEL org.opencontainers.image.authors="Codenotary Inc. <info@codenotary.com>"
+
+ARG IMMU_UID="65532"
+ARG IMMU_GID="65532"
+ARG IMMUDB_HOME="/usr/share/immudb"
+
+WORKDIR /usr/bin
+COPY --from=build-fips /src/immudb .
+COPY --from=build-fips /src/immuadmin .
+COPY --from=build-fips "/etc/ssl/certs/ca-certificates.crt" "/etc/ssl/certs/ca-certificates.crt"
```
byo, line 22: This should not be necessary, distroless should already contain ca-certificates.crt file  (or a compatible alternative)

Dockerfile.fips
```diff
@@ -0,0 +1,52 @@
...
+COPY . .
+RUN rm -rf /src/webconsole/dist
+RUN GOOS=linux GOARCH=amd64 WEBCONSOLE=default make immuadmin-static immudb-fips
+RUN mkdir /empty
+
+### distroless FIPS 140-2
+FROM gcr.io/distroless/base:nonroot AS distroless-fips
+LABEL org.opencontainers.image.authors="Codenotary Inc. <info@codenotary.com>"
+
+ARG IMMU_UID="65532"
```
byo, line 15: Other dockerfiles use 3322 UID/GID, is it safer to use larger values ? I could see a potential issue here if someone would like to easily switch between fips and non-fips image. You would have to assume different UID/GID for both images.

build/Dockerfile.fips
```diff
@@ -0,0 +1,52 @@
...
+LABEL org.opencontainers.image.authors="Codenotary Inc. <info@codenotary.com>"
+
+ARG IMMU_UID="65532"
+ARG IMMU_GID="65532"
+ARG IMMUDB_HOME="/usr/share/immudb"
+
+WORKDIR /usr/bin
+COPY --from=build-fips /src/immudb .
+COPY --from=build-fips /src/immuadmin .
+COPY --from=build-fips "/etc/ssl/certs/ca-certificates.crt" "/etc/ssl/certs/ca-certificates.crt"
```
byo, line 22: This should not be necessary, distroless should already contain ca-certificates.crt file (or a compatible alternative)

build/Dockerfile.fips
```diff
@@ -0,0 +1,52 @@
...
+COPY . .
+RUN rm -rf /src/webconsole/dist
+RUN GOOS=linux GOARCH=amd64 WEBCONSOLE=default make immuadmin-static immudb-fips
+RUN mkdir /empty
+
+### distroless FIPS 140-2
+FROM gcr.io/distroless/base:nonroot AS distroless-fips
+LABEL org.opencontainers.image.authors="Codenotary Inc. <info@codenotary.com>"
+
+ARG IMMU_UID="65532"
```
byo, line 15: Other dockerfiles use 3322 UID/GID, is it safer to use larger values ? I could see a potential issue here if someone would like to easily switch between fips and non-fips image. You would have to assume different UID/GID for both images.

byo (commented):
Few small comments left.

Another things to do:

- [x] test if a normal immuclient / immudb sdk can connect to fips-enabled server using certificates (if easy to do, let's add this to GH actions too)
- [ ] check if the fips-enabled server can properly sign the state if state signing key is provided

build/Dockerfile.fips
arriqaaq: actually the nonroot user is created by default for the distroless image with uid/gid 65532, so that is the reason I've used it. Any potential issue?

.github/workflows/pull.yml
arriqaaq: ack

arriqaaq: done

build/Dockerfile.fips
arriqaaq: This just uses the default nonroot user created on a distroless image with uid/gid 65532)

arriqaaq:
@byo @jeroiraz 

Verified the following

1) test if a normal immuclient / immudb sdk can connect to fips-enabled server using certificates

Used generate.sh script to generate certs

```
#fips server
docker run --net host -v $(pwd):/src -it --rm immudb:fips --mtls --certificate="/src/tools/mtls/3_application/certs/localhost.cert.pem" --clientcas="/src/tools/mtls/2_intermediate/certs/ca-chain.cert.pem" --pkey="/src//tools/mtls/3_application/private/localhost.key.pem"

```

```
#client
package main

import (
	"context"
	"log"
	"fmt"
	immudb "github.com/codenotary/immudb/pkg/client"
)

func main() {
	// Folder cotaining MTLS certificates
	pathToMTLSFolder := "./mtls"

	opts := immudb.DefaultOptions().
		WithAddress("localhost").
		WithPort(3322).
		WithMTLs(true).
		WithMTLsOptions(
			immudb.MTLsOptions{}.
				WithCertificate(pathToMTLSFolder + "/4_client/certs/localhost.cert.pem").
				WithPkey(pathToMTLSFolder + "/4_client/private/localhost.key.pem").
				WithClientCAs(pathToMTLSFolder + "/2_intermediate/certs/ca-chain.cert.pem").
				WithServername("localhost"),
		)

	client := immudb.NewClient().WithOptions(opts)
	err := client.OpenSession(context.TODO(), []byte(`immudb`), []byte(`immudb`), "defaultdb")
	if err != nil {
		log.Fatal(err)
	}

	defer client.CloseSession(context.TODO())

	// do amazing stuff
	// write an entry
	_, err = client.Set(context.Background(), []byte("my key"), []byte("my value"))
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("Sucessfully set an entry: ('%s', '%s')\n", []byte("my key"), []byte("my value"))

	// read an entry
	entry, err := client.Get(context.Background(), []byte("my key"))
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("Sucessfully got entry: ('%s', '%s') @ tx %d\n", entry.Key, entry.Value, entry.Tx)
}
```

Passes

2) check if the fips-enabled server can properly sign the state if state signing key is provided

```
#fips server

docker run --net host -v $(pwd):/src -it --rm immudb:fips --signingKey="/src/test/signer/ec1.key"
```

```
#client 

package main

import (
	"context"
	"log"
	"fmt"
	immudb "github.com/codenotary/immudb/pkg/client"
)


func main() {
	c, err := client.NewImmuClient(client.DefaultOptions().WithServerSigningPubKey("../test/signer/ec1.pub"))
	if err != nil {
		log.Fatal(err)
	}
	ctx := context.Background()
	
	_ , err = c.Login(ctx, []byte(`immudb`), []byte(`immudb`))
	if err != nil {
		log.Fatal(err)
	}
	
	if _, err = c.Set(ctx, []byte(`immudb`), []byte(`hello world`)); err != nil {
		log.Fatal(err)
	}
	fmt.Printf("Sucessfully set an entry: ('%s', '%s')\n", []byte("immudb"), []byte("hello world"))

	// read an entry
	entry, err := c.Get(context.Background(), []byte("immudb"))
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("Sucessfully got entry: ('%s', '%s') @ tx %d\n", entry.Key, entry.Value, entry.Tx)
	

	var state *schema.ImmutableState
	if state, err = c.CurrentState(ctx); err != nil {
		log.Fatal(err) // if signature is not verified here is trigger an appropriate error
	}
	
	fmt.Println("state: ",state)
}

```
Passes, and fails if wrong public key is provided, as expected

build/Dockerfile.fips
byo: Currently we don't have any plans to do so and if someone would like to use custom CA, one can either create its own docker image based on our official ones or use something like a configmap in kubernetes and avoid changing the dockerfile completely.

build/Dockerfile.fips
byo: I just found that there's the user called `nonroot in distroless` - in such case we should avoid using raw UID/GID values (like remove the ARG completely) and use the `nonroot` user name instead. the UID/GID is only necessary where there's really no named user with such UID/GID

tomekkolo:
Do not see it in PR, but immudb should log that it is run in fips mode. It would also make sense to build immuclient-fips and immuadmin-fips.

build/Dockerfile.fips
arriqaaq: ack

build/Dockerfile.fips
arriqaaq: ack

cmd/immudb/fips/fips.go
```diff
@@ -0,0 +1,30 @@
...
+package main
+
+import (
+	_ "crypto/tls/fipsonly"
+
+	immudb "github.com/codenotary/immudb/cmd/immudb/command"
+)
+
+func main() {
+	immudb.Execute()
```
arriqaaq, line 29: @tomekkolo this is the only place a log can be added to mention if it is running in FIPS mode

Makefile
```diff
@@ -181,9 +204,20 @@ clean/dist:
...
 .PHONY: dist
-dist: webconsole dist/binaries
+dist: webconsole dist/binaries dist/fips
 	@echo 'Binaries generation complete. Now vcn signature is needed.'
 
+# build FIPS binary from docker image (no arm or non-linux support)
+.PHONY: dist/fips
+dist/fips: clean
+	docker build -t fips:build -f build/fips/Dockerfile.build .
+	docker run -v ${PWD}:/src -it --user root --rm fips:build -c "WEBCONSOLE=default make immudb-fips"
```
byo, line 214: This could potentially cause some issues regarding local user permissions, maybe we could run some docker container in the background and use `docker cp` to exchange files?

byo (dismissed):
Left one comment for building fips-enabled binaries (but I wouldn't consider that a blocker), other than that looks good

cmd/immudb/fips/fips.go
arriqaaq: fixed

Makefile
arriqaaq: actually this was creating problem for the builds, and if the ubuntu containers we are building off run as root, it causes an issue, hence added the `--user root` flag. I'd like to revisit this later if it ain't a blocker because the changes require checking the CI process on how it is being cloned and run as

byo (dismissed):
LGTM

byo (approved):

PR #1366, opened 2022-09-06
Fix/simplify dockerfile

byo:


Dockerfile
```diff
@@ -7,50 +21,7 @@ RUN rm -rf /src/webconsole/dist
...
-EXPOSE 8080
-EXPOSE 5432
-
-HEALTHCHECK --interval=30s --timeout=30s --start-period=5s --retries=3 CMD [ "/usr/local/bin/immuadmin", "status" ]
-USER immu
-ENTRYPOINT ["/usr/sbin/immudb"]
-
-
-FROM scratch as scratch
+FROM scratch
```
arriqaaq, line 24: any reason we prefer scratch to gcr.io/distroless/static:nonroot?

vchaindz: scratch is a bit smaller. But we should add a distroless Dockerfile within a Docker subfolder

arriqaaq: Ack, actually there is a non static version for distroless which has ca certs too, about the size overhead

```
gcr.io/distroless/static                                                       nonroot            2.36MB
gcr.io/distroless/base                                                         nonroot          20.3MB
```

arriqaaq: I built from both scratch and distroless (static and non static)

scratch vs distroless static ---> 1mb overhead
scratch vs distroless non static ---> 20mb overhead

arriqaaq (approved):

Dockerfile
arriqaaq: Better to keep scratch for now, as Dennis mentioned, we can add a separate distroless image in the build folder, Dockerfile.distroless

PR #1376, opened 2022-09-14
Feat sync followers

jeroiraz:
This PR introduces support for synchronous followers.

It's possible to set the number of followers required to acknowledge replication before transactions can be fully committed.

The major change required at the embedded store was the support for durable pre-committed transactions. This was done without introducing performance degradation as no additional synchronization is needed. Commit log entries are delayed but transaction headers and payload logs are synced, ensuring it's possible to read such information when needed.

Replication APIs at grpc server layer are aware of sync followers and instruct commit when enough followers ack replication.

Support for discarding pre-committed transactions has been added


pkg/database/database.go
```diff
@@ -1145,7 +1188,89 @@ func (d *db) serializeTx(tx *store.Tx, spec *schema.EntriesSpec, snap *store.Sna
...
+			precommittedAlh:  newFollowerAlh,
+		}
+	}
+
+	// check up to which tx enough followers ack replication and it's safe to commit
+	mayCommitUpToTxID := uint64(0)
+	allowances := 0
+
+	// we may clean up followerStates from those who are lag behind commit
+	for _, st := range d.followerStates {
```
arriqaaq, line 1253: Just trying to understand this better

allowances will always be equal to the number of followers, this implies acks (allowances>=syncfollowers) from followers will always the same right? Is this to check at least `n replicas` are alive and up to date?

pkg/database/database.go
```diff
@@ -1145,7 +1188,89 @@ func (d *db) serializeTx(tx *store.Tx, spec *schema.EntriesSpec, snap *store.Sna
...
+		if st.precommittedTxID < mayCommitUpToTxID {
```
arriqaaq, line 1254: mayCommitUpToTxID would always be 0, unless there are negative values that are being stored in the state? Here is an example for the same https://go.dev/play/p/vs-QG-iYn6P

arriqaaq: assume we have three followers (f1, f2 and f3), what happens in scenario where the follower states are like this

```
	followerStates := map[string]int{
		"f1": 4,
		"f2": 3,
		"f3": 1,
	}
```
the mayCommitUpToTxID should be the minimum in the map right, in this scenario, `1`?

embedded/store/immustore.go
```diff
@@ -2278,15 +2364,41 @@ func (s *ImmuStore) sync() error {
...
+			return err
+		}
+
+		commitUpToTxID = txID
+		commitUpToTxAlh = alh
+	}
+
+	if commitUpToTxID != commitAllowedUpToTxID {
+		// added as a safety fuse but this situation should NOT happen
+		return fmt.Errorf("%w: may commit up to %d but actual transaction to be committed is %d",
```
arriqaaq, line 2400: currently the commitUpToTxID is always set to 0 when syncing, and when this error is returned, the replicator runs into an infinite loop

embedded/store/immustore.go
```diff
@@ -565,11 +572,9 @@ func OpenWith(path string, vLogs []appendable.Appendable, txLog, cLog appendable
 	}
 
 	if store.synced {
-		store.cLogBuf = make([]byte, cLogEntrySize*opts.MaxActiveTransactions)
-
 		go func() {
 			for {
-				committedTxID := store.lastCommittedTxID()
+				committedTxID := store.LastCommittedTxID()
```
arriqaaq, line 577: If an error occurs (and if syncFrequency is not set), this runs into an infinite loop

arriqaaq: I am assuming this happens because sync frequency is not set on defaultdb

pkg/replication/replicator.go
```diff
@@ -245,13 +251,47 @@ func (txr *TxReplicator) disconnect() {
...
+
+	md := exportTxStream.Trailer()
+
+	if len(md.Get("may-commit-up-to-txid-bin")) > 0 && len(md.Get("may-commit-up-to-alh-bin")) > 0 {
+		mayCommitUpToTxID := binary.BigEndian.Uint64([]byte(md.Get("may-commit-up-to-txid-bin")[0]))
+
+		var mayCommitUpToAlh [sha256.Size]byte
+		copy(mayCommitUpToAlh[:], []byte(md.Get("may-commit-up-to-alh-bin")[0]))
+
+		err = txr.db.AllowCommitUpto(mayCommitUpToTxID, mayCommitUpToAlh)
```
arriqaaq, line 288: This currently fails, when master is started and replica connected, it is not able to fetch the transaction (tx not found error). The pre-committed state is not consistent with master. I assume this is because we are reading from the immustore before the tx has been added to it

embedded/store/precommit_buffer.go
```diff
@@ -0,0 +1,134 @@
...
+	txID   uint64
+	alh    [sha256.Size]byte
+	txOff  int64
+	txSize int
+}
+
+type precommitBuffer struct {
+	buf []*precommittedEntry
+
+	rpos int
```
arriqaaq, line 37: could you add a comment here on what rpos and wpos is

embedded/store/immustore.go
```diff
@@ -2278,17 +2507,48 @@ func (s *ImmuStore) sync() error {
...
+			return err
+		}
+
+		commitUpToTxID = txID
+		commitUpToTxAlh = alh
+	}
+
+	if commitUpToTxID != commitAllowedUpToTxID {
+		// added as a safety fuse but this situation should NOT happen
+		return fmt.Errorf("%w: may commit up to %d but actual transaction to be committed is %d",
```
arriqaaq, line 2548: Currently starting the master with one replica `./immudb --replication-sync-followers 1` and it gets stuck at this point in a loop

cmd/immuadmin/command/database.go
```diff
@@ -33,7 +33,8 @@ import (
 func addDbUpdateFlags(c *cobra.Command) {
 	c.Flags().Bool("exclude-commit-time", false,
 		"do not include server-side timestamps in commit checksums, useful when reproducibility is a desired feature")
-	c.Flags().Bool("replication-enabled", false, "set database as a replica")
+	c.Flags().Bool("replication-enabled", false, "set database as a replica") // TODO: flag name should be changed to something like `replication-is-replica`
+	c.Flags().String("replication-sync-followers", "", "set a minimum number of followers for ack replication before transactions can be committed")
```
arriqaaq, line 37: This has to be an Uint32

arriqaaq:
There is an issue stopping the server when the replication follower are set on the database directly (for defaultdb). Following are the steps:

1) start master
```
./immudb --replication-sync-followers 1
```
2) use client to set some value in defaultdb
```
./immuclient use defaultdb && ./immuclient safeset y 8
```
Client stuck, server not able to close

embedded/store/indexer.go
```diff
@@ -317,8 +319,8 @@ func (idx *indexer) Pause() {
 	idx.stateCond.L.Unlock()
 }
 
-func (idx *indexer) doIndexing(cancellation <-chan struct{}) {
-	committedTxID := idx.store.lastCommittedTxID()
+func (idx *indexer) doIndexing() {
```
byo, line 322: Why cancellation was removed here? Was it not used anywhere?

embedded/store/precommit_buffer.go
```diff
@@ -0,0 +1,134 @@
...
+	txID   uint64
+	alh    [sha256.Size]byte
+	txOff  int64
+	txSize int
+}
+
+// precommitBuffer is a read-ahead circular buffer
+// this buffer is used to hold a portion of the clog in memory:
+//   - entries put into the buffer as they are precommitted
+//   - entries are removed from the buffer as they are committed (content was successfully written into clog)
```
byo, line 37:
```suggestion
//   - entries are removed from the buffer as they are committed (content has successfully written into clog)
```

pkg/api/schema/schema.proto
```diff
@@ -451,6 +454,16 @@ message TxList {
 
 message ExportTxRequest {
 	uint64 tx = 1;
+	bool allowPreCommitted = 2;
+	FollowerState followerState = 3;
+}
+
+message FollowerState {
+	string UUID = 1;
```
byo, line 462: Currently we don't share UUID across primary and replicas, but I assume this is rather some replica's UUID rather than cluster UUID?

pkg/api/schema/schema.proto
```diff
@@ -345,6 +345,9 @@ message ImmutableState {
 	uint64 txId = 2;
 	bytes txHash = 3;
 	Signature signature = 4;
+	// following fields are not part of the signature
+	uint64 precommittedTxId = 5;
```
byo, line 349: Why those have to re reported ?

pkg/database/database.go
```diff
@@ -236,39 +265,14 @@ func (d *db) releaseTx(tx *store.Tx) {
 }
 
 func (d *db) initSQLEngine() error {
-	// Warn about existent SQL data
```
byo, line 239: Why is this change needed?

pkg/database/database.go
```diff
@@ -1145,18 +1171,188 @@ func (d *db) serializeTx(tx *store.Tx, spec *schema.EntriesSpec, snap *store.Sna
...
+
+	// check up to which tx enough followers ack replication and it's safe to commit
+	mayCommitUpToTxID := uint64(0)
+	if len(d.followerStates) > 0 {
+		mayCommitUpToTxID = math.MaxUint64
+	}
+
+	allowances := 0
+
+	// we may clean up followerStates from those who are lag behind commit
```
byo, line 1220:
```suggestion
	// we may clean up followerStates from those who are lagging behind commit
```

pkg/database/database.go
```diff
@@ -1145,18 +1171,188 @@ func (d *db) serializeTx(tx *store.Tx, spec *schema.EntriesSpec, snap *store.Sna
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (d *db) ExportTxByID(req *schema.ExportTxRequest) (txbs []byte, mayCommitUpToTxID uint64, mayCommitUpToAlh [sha256.Size]byte, err error) {
+	d.exportTxMutex.Lock()
+	defer d.exportTxMutex.Unlock()
```
byo, line 1240: This mutex will remove ability run multiple exports in parallel.

byo (commented):
I did a first shallow go over the PR, have few comments.

I will now follow with more deep review.

embedded/store/immustore.go
```diff
@@ -1942,17 +2121,47 @@ func (s *ImmuStore) ReplicateTx(exportedTx []byte, waitForIndexing bool) (*TxHea
 		return nil, ErrIllegalArguments
 	}
 
-	return s.commit(txSpec, hdr, waitForIndexing)
+	txHdr, err := s.precommit(txSpec, hdr)
+	if err != nil {
+		return nil, err
+	}
+
+	err = s.durablePrecommitWHub.WaitFor(txHdr.ID, nil)
```
byo, line 2129: Potential performance issue - we durably store each precomitted TX separately.

embedded/store/immustore.go
```diff
@@ -1942,17 +2121,47 @@ func (s *ImmuStore) ReplicateTx(exportedTx []byte, waitForIndexing bool) (*TxHea
...
+	if err == watchers.ErrAlreadyClosed {
```
byo, line 2130:
```suggestion
	if errors.Is(err, watchers.ErrAlreadyClosed) {
```

embedded/store/immustore.go
```diff
@@ -2036,7 +2259,18 @@ func (s *ImmuStore) appendableReaderForTx(txID uint64) (*appendable.Reader, erro
 }
 
 func (s *ImmuStore) ReadTx(txID uint64, tx *Tx) error {
-	r, err := s.appendableReaderForTx(txID)
+	s.mutex.Lock()
```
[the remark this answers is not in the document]
jeroiraz: probably it can be removed, as appendable and cache are thread-safe, and the state-related attributes have its own lock... I don't expect an noticeable impact in performance, it will ensure if any read is initiated it will complete and not being closed after partially reading...

pkg/api/schema/schema.proto
jeroiraz: it's the follower UUID

pkg/api/schema/schema.proto
jeroiraz: as discussed offline, that information may be needed to decide which follower may become the new master

pkg/database/database.go
jeroiraz: to be able to open the store with ExternalCommitAllowance enabled or disabled, but not requiring to switch to special cases because initialization transactions must be committed before

pkg/database/database.go
jeroiraz: at this stage is more about safety, probably the locking may be moved to the state holder

embedded/store/immustore.go
```diff
@@ -183,8 +186,9 @@ type ImmuStore struct {
 	blBuffer chan ([sha256.Size]byte)
 	blErr    error
 
-	precommitWHub *watchers.WatchersHub
-	commitWHub    *watchers.WatchersHub
+	precommitWHub        *watchers.WatchersHub
+	durablePrecommitWHub *watchers.WatchersHub
```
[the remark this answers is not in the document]
jeroiraz: I'm not convinced also about the name, it was the best I could think of. Didn't wanted to change existent one, but I'd think it could be done even afterwards as well

embedded/store/immustore.go
```diff
@@ -1425,31 +1476,152 @@ func (s *ImmuStore) performPreCommit(tx *Tx, ts int64, blTxID uint64) error {
...
+	s.commitStateRWMutex.Lock()
+	defer s.commitStateRWMutex.Unlock()
+
+	if txID <= s.commitAllowedUpToTxID {
+		// once a commit is allowed, it cannot be revoked
+		return nil
+	}
+
+	if s.preCommittedTxID < txID {
+		// commit allowances apply only to pre-committed transactions
```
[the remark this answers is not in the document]
jeroiraz: hm, no sure if this could lead to an inconsistent situation

embedded/store/immustore.go
```diff
@@ -1772,8 +1944,15 @@ func (r *slicedReaderAt) ReadAt(bs []byte, off int64) (n int, err error) {
 	return available, nil
 }
 
-func (s *ImmuStore) ExportTx(txID uint64, tx *Tx) ([]byte, error) {
-	err := s.ReadTx(txID, tx)
+func (s *ImmuStore) ExportTx(txID uint64, allowPrecommitted bool, tx *Tx) ([]byte, error) {
+	s.mutex.Lock()
```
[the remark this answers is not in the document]
jeroiraz: yes, I'd prefer at this stage to keep  the lock, but we could re-evaluate it

embedded/store/immustore.go
jeroiraz: actually no, the precommitted transactions are subject to grouped sync as before

embedded/store/immustore.go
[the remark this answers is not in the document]
byo: This would not intruduce inconsistency but could block the replica from performing the commit, it would automatically fix in the future but that requires another commit from the primary with updated information up to what TX the commit can be performed.

embedded/store/immustore.go
byo: yes, but if you're a replica there's only a single replicator that writes transactions to the database - and if I understand correctly, it is doing it sequentially.

pkg/api/schema/schema.proto
byo: Something to keep in mind then - once we introduce a notion of a cluster - such cluster should be seen as a single UUID which means we will have to differentiate between immudb process UUID and immudb cluster UUID.

embedded/store/immustore.go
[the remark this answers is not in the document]
byo: I could see a potential risk in a situation where more than one secondary is involved - it the time to export a TX is large and dominates the whole commit time, adding more replicas will mean that those will fight for the mutex lock.

I think we could leave it for future improvements but also be cautious about this lock if it turns out that there's significant performance drop when second or third replica is added.

pkg/database/database.go
```diff
@@ -1145,18 +1172,188 @@ func (d *db) serializeTx(tx *store.Tx, spec *schema.EntriesSpec, snap *store.Sna
...
+			(req.AllowPreCommitted && committedTxID < preCommittedTxID) {
+			ctx, cancel := context.WithTimeout(context.Background(), d.options.storeOpts.SyncFrequency*4)
+			defer cancel()
+			cancellation = ctx.Done()
+		}
+
+	}
+
+	err = d.WaitForTx(req.Tx, req.AllowPreCommitted, cancellation)
+	if err == watchers.ErrCancellationRequested {
```
byo, line 1344:
```suggestion
	if errors.Is(err, watchers.ErrCancellationRequested) {
```

pkg/replication/replicator.go
```diff
@@ -244,14 +213,88 @@ func (txr *TxReplicator) disconnect() {
...
-func (txr *TxReplicator) fetchTX() ([]byte, error) {
-	exportTxStream, err := txr.client.ExportTx(txr.clientContext, &schema.ExportTxRequest{Tx: txr.nextTx})
+func (txr *TxReplicator) nextTx() error {
+	commitState, err := txr.db.CurrentState()
 	if err != nil {
-		return nil, err
+		return err
+	}
+
+	syncReplicationEnabled := txr.db.IsSyncReplicationEnabled()
```
byo, line 222: The syncReplicationEnabled flag is used for both primary and replicas. This limits possible setups - e.g. there may be a replica that is synchronous in relation to its primary but it can also have replicas that are asynchronous. Are we able to handle such case today?

cmd/immuadmin/command/database.go
```diff
@@ -33,10 +33,12 @@ import (
 func addDbUpdateFlags(c *cobra.Command) {
 	c.Flags().Bool("exclude-commit-time", false,
 		"do not include server-side timestamps in commit checksums, useful when reproducibility is a desired feature")
-	c.Flags().Bool("replication-enabled", false, "set database as a replica")
+	c.Flags().Bool("replication-enabled", false, "set database as a replica") // TODO: flag name should be changed to something like `replication-is-replica`
+	c.Flags().Bool("replication-sync-replication", false, "enable synchronous replication")
+	c.Flags().Uint32("replication-sync-followers", 0, "set a minimum number of followers for ack replication before transactions can be committed")
```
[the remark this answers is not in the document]
jeroiraz: I'd think from perspective of the deployment, the name `replication-sync-followers` is clear because it means how many followers are actually synchronized with the master. There could be more followers, but that specify the number which is guaranteed to be in sync
However I got your point, we may need to get feedback on this as there are multiple possibilities @SimoneLazzaris ?

pkg/database/database.go
```diff
@@ -1145,18 +1172,188 @@ func (d *db) serializeTx(tx *store.Tx, spec *schema.EntriesSpec, snap *store.Sna
...
+
+	// we may clean up followerStates from those who are lagging behind commit
+	for _, st := range d.followerStates {
+		if st.precommittedTxID < mayCommitUpToTxID {
+			mayCommitUpToTxID = st.precommittedTxID
+		}
+		allowances++
+	}
+
+	if allowances >= d.options.syncFollowers {
```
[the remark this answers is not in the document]
jeroiraz: it's a good point, not even due to malicious but on slower ones, as current implementation is waiting for them

pkg/replication/replicator.go
jeroiraz: yes, that scenario should be handled without issues

embedded/store/precommit_buffer.go
```diff
@@ -0,0 +1,152 @@
...
+	e.alh = alh
+	e.txOff = txOff
+	e.txSize = txSize
+
+	b.full = b.rpos == b.wpos
+
+	return nil
+}
+
+func (b *precommitBuffer) recedeWriter(n int) error {
```
arriqaaq, line 97: Could you imply what recedewriter does? does it do some rollback?

cmd/immuadmin/command/database.go
```diff
@@ -33,12 +34,17 @@ import (
...
-	c.Flags().Bool("replication-enabled", false, "set database as a replica")
+	c.Flags().Bool("replication-enabled", false, "set database as a replica") // TODO: flag name should be changed to something like `replication-is-replica`
+	c.Flags().Bool("replication-sync-enabled", false, "enable synchronous replication")
+	c.Flags().Uint32("replication-sync-followers", 0, "set a minimum number of followers for ack replication before transactions can be committed")
 	c.Flags().String("replication-master-database", "", "set master database to be replicated")
 	c.Flags().String("replication-master-address", "", "set master address")
 	c.Flags().Uint32("replication-master-port", 0, "set master port")
 	c.Flags().String("replication-follower-username", "", "set username used for replication")
 	c.Flags().String("replication-follower-password", "", "set password used for replication")
+	c.Flags().Uint32("replication-prefetch-tx-buffer-size", uint32(replication.DefaultPrefetchTxBufferSize), "maximum number of prefeched transactions")
```
arriqaaq, line 45: wdyt of using the already existing pattern for options

1) `replication-master-*` for master settings
2) `replication-follower-*` for follower settings

Few suggestions
`replication-prefetch-tx-buffer-size` --> `replication-follower-buffer-size`
`replication-commit-concurrency` --> `replication-follower-concurrency`
`replication-allow-tx-discarding` --> `replication-follower-discard-precommit`

arriqaaq: This makes it simple to understand which setting is for what when configuring the primary or follower. Now that we have so many flags, this small differentiation will help IMO

jeroiraz: yes, it's becoming complex as the number of flags is increasing. I also though about removing the `commit` portion in `replication-commit-concurrency`, however by then I was not fully sure there would be  a single thing to set, in this case is the number of threads requesting commit, but in the future we might also concurrently fetch transactions. Probably it's not the case and then we can remove `commit` from the flag. And then we can add the `follower` prefix to make it clear as you are proposing

PR #1377, opened 2022-09-14
Fix/helm init volume

SimoneLazzaris:
This PR adds an init container which fixes ownership for the persistent volume immudb is mounting to store its databases.

At least with digital ocean, a persisten volume is mounted as root:root with 0755 permissions, and that overrides permissions and ownership of the mountpoint, so it can't be fixed simply in the Dockerfile
Without this fix immudb helm chart can't be successfully deployed on Digital Ocean kubernetes (and probably other providers).

### DIFF
--- a/helm/templates/statefulset.yaml
+++ b/helm/templates/statefulset.yaml
@@ -72,8 +72,8 @@ spec:
           volumeMounts:
           - mountPath: /var/lib/immudb
             name: immudb-storage
-            {{- if $.Values.volumeSubPath }}
-            subPath: immudb
+            {{- if $.Values.volumeSubPath.enabled }}
+            subPath: {{ $.Values.volumeSubPath.path | quote }}
             {{- end}}
       {{- with .Values.nodeSelector }}
       nodeSelector:
--- a/helm/templates/statefulset.yaml
+++ b/helm/templates/statefulset.yaml
@@ -32,20 +32,12 @@ spec:
       - name: immudb-storage
         persistentVolumeClaim:
           claimName: {{ include "immudb.fullname" . }}
-      initContainers:
-        - command: ["/bin/chown", "-R", "3322:3322", "/data"]
-          image: busybox
-          name: permission-fix
-          volumeMounts:
-          - mountPath: /data
-            name: immudb-storage
-            {{- if $.Values.volumeSubPath }}
-            subPath: immudb
-            {{- end}}
+      securityContext:
+        {{- toYaml .Values.securityContext | nindent 8 }}
       containers:
         - name: {{ .Chart.Name }}
           securityContext:
-            {{- toYaml .Values.securityContext | nindent 12 }}
+            {{- toYaml .Values.podSecurityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           ports:
--- a/helm/templates/statefulset.yaml
+++ b/helm/templates/statefulset.yaml
@@ -12,7 +12,7 @@ spec:
   selector:
     matchLabels:
       {{- include "immudb.selectorLabels" . | nindent 6 }}
-  serviceName: {{ include "immudb.fullname" . }}    
+  serviceName: {{ include "immudb.fullname" . }}
   template:
     metadata:
       {{- with .Values.podAnnotations }}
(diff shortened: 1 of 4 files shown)


helm/templates/statefulset.yaml
```diff
@@ -64,12 +72,13 @@ spec:
...
                 name: {{ include "immudb.fullname" . }}-credentials
                 key: immudb-admin-password
-          {{- end}}      
+          {{- end}}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
           volumeMounts:
           - mountPath: /var/lib/immudb
             name: immudb-storage
+            subPath: immudb
```
SimoneLazzaris, line 81: I've introduced a subPath since otherwise immudb would see a `lost+found` directory and assume it is a database

marcosQuesada (approved):

raducrisan1 (approved):

helm/templates/statefulset.yaml
byo: This is a breaking change - we have to think about how to announce it then if someone is already using the helm chart. Maybe some automatic migration procedure? (although I'm not really in favor of such approach).

helm/templates/statefulset.yaml
```diff
@@ -32,6 +32,14 @@ spec:
       - name: immudb-storage
         persistentVolumeClaim:
           claimName: {{ include "immudb.fullname" . }}
+      initContainers:
+        - command: ["/bin/chown", "-R", "3322:3322", "/data"]
```
byo, line 36: This uses hardcoded uid/gid. In fips we want to use distroless:noroot images which use completely different uid/gid, we should make it parametrizable I believe.

Also kubernetes shoudl automatically ensure the GID is set properly when fsGroup is set in the security context: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods - can't we use that functionality ? It would not require init container that runs as root (in some clusters its not allowed to run any container as a root).

helm/templates/statefulset.yaml
SimoneLazzaris: I can add a compatibility flag maybe? So that the `subPath` clause is only added if something like `.values.volumeSubPath` is set to true. 
We set it to true in `values.yaml` and explain that if you already have some data either you migrate it (and we can also provide a guide to do that) or you set this value to false.

helm/templates/statefulset.yaml
arriqaaq: I agree, we can use securityContext and add the runAsUser/runAsGroup/fsGroup to set this

helm/templates/statefulset.yaml
byo: I was thinking about other solutions here too:
1) Ignore lost+found (and maybe some other folders) in immudb itself
2) Add some init container to check if there's a subvolume but the main folder contains immudb.identifier file that would indicate that previously immudb files were stored in the main folder
3) Don't setup volumeSubPath by default but have a clear documentation why it should be set, maybe also allow specifying that in the values.yaml explicitly to any subfolder, not only `immudb` value?

SimoneLazzaris: See commit e3005ef7 for that

SimoneLazzaris: Solution 2 is complicated, I'd go with 1 or 3. 
I suggest to adopt the third.
We can have a `value.yaml` file like this:
```yaml
subPath:
  enabled: true # or false
  path: /immudb
```

and then use those values in the template. Plus add some documentations on when to use that (tipically, ext4 formatted block-based volumes, like DO or EBS)

helm/templates/statefulset.yaml
SimoneLazzaris: I'm not sure it will do, since the directory is mounted 0755, and we will need 0775. I'm doing some tests on it

SimoneLazzaris: Interesting. If using securityContext.fsGroup, directory is mounted as 2775. So that will work.

SimoneLazzaris: Commit fc9d9d49 should address this; now we use securityContext.fsGroup and it work

helm/templates/statefulset.yaml
SimoneLazzaris: Commit 3eddd21a should address this.

PR #1380, opened 2022-09-16
chore(pkg/errors): add multiple error aggregation

arriqaaq:
This PR adds a helper utility for error aggregation. This is helpful in aggregating multiple errors. It works as follows

```
func doSomething() error{
       err := make([]error,0,1)

        _, err1  := somefunc()
        err = append(err, err1)  // no need to return unless it needs to

        _, err2  := somefunc2()
        err = append(err, err2) // concatenate errors easily
       
        _, err3  := somefunc3()
        err = append(err, err3) // concatenate errors easily

       return NewAggregate(err)
}
```

In a for loop
```
func doSomethingInLoop() error{
       err := make([]error,0,1)

       for i:=range something {
        _, err1  := somefunc()
                err = append(err, err1)  // no need to return unless it needs to
        }
       return NewAggregate(err)
}
```

### DIFF
--- a/pkg/errors/aggregate.go
+++ b/pkg/errors/aggregate.go
@@ -1,0 +1,112 @@
+/*
+Copyright 2022 Codenotary Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


byo (commented):
Please check https://github.com/codenotary/immudb/tree/master/embedded/multierr - looks like it's very similar functionality.

arriqaaq:
I see, thanks for pointing it out @byo

Issue #1385, opened 2022-09-20
[sync-replication] Unable to set up a replicated database


SimoneLazzaris:
**What happened**
I've tried creating a replicated database, but didn't have success.
I end up with an unresponsive database
**What you expected to happen**

**How to reproduce it (as minimally and precisely as possible)**
- master setup:
```sh
./immudb -a 127.0.0.2 --dir data_master -d
export IMMUADMIN_IMMUDB_ADDRESS=127.0.0.2
./immuadmin login immudb
./immuadmin database create bla --replication-sync-followers 1 -a 127.0.0.2
```
- replica setup
```sh
./immudb -a 127.0.0.3 --dir data_slave -d
export IMMUADMIN_IMMUDB_ADDRESS=127.0.0.3
./immuadmin login immudb
./immuadmin database create bla -a 127.0.0.3 --replication-enabled --replication-master-address 127.0.0.2 --replication-master-database bla --replication-follower-username immudb --replication-follower-password immudb
```

- client
```sh
export IMMUCLIENT_IMMUDB_ADDRESS=127.0.0.2
./immuclient login immudb
./immuclient use bla
./immuclient safeset gorilla banana # freezes
```

Tried to kill servers and restart, I had to `kill -9` master instance, since it didn't respond to SIGTERM.
After restart, client still hangs on the `safeset`

**Environment**
```shell
# run "immu* version" and copy/paste the output here
immudb 1.3.2
Commit  : 113f73e4d014ce81b7838b9c7c73151681e4913d
Built at: Tue, 20 Sep 2022 08:53:30 CEST
```

**Additional info (any other context about the problem)**

arriqaaq:
I've been starting the follower server using this command
```
./immudb --replication-enabled=true --replication-follower-password=immudb --replication-follower-username=immudb --replication-master-address=localhost --port=3323 --pgsql-server=false --metrics-server=false
```

I am not sure but here is my guess on why it is blocking. Currently the follower is setup 

```
./immudb -a 127.0.0.3 --dir data_slave -d
```

So the follower does not know who the master is, so because bla database is created with one sync follower, it does not get an ack

jeroiraz:
`--replication-master-port` may be missing, then automatic replication is not enabled. This is not related to new changes, we may need to add more validations on the parameters.
Recently the expected default value for this flag was set. to this situation won't happen here. Nevertheless, further validations and logging will help

emkaminsk:
[...]

PR #1387, opened 2022-09-20
fix(pkg/server): add validation to replication options

arriqaaq:
Addresses #1385


pkg/server/server.go
```diff
@@ -862,8 +864,12 @@ func (s *ImmuServer) CreateDatabaseV2(ctx context.Context, req *schema.CreateDat
 	s.logDBOptions(db.GetName(), dbOpts)
 
 	err = s.startReplicationFor(db, dbOpts)
-	if err != nil && err != ErrReplicatorNotNeeded {
-		return nil, fmt.Errorf("%w: while starting replication", err)
+	if err != nil {
+		if err == ErrReplicatorNotNeeded {
+			s.Logger.Warningf("Replication for database '%s' not required.", db.GetName())
```
jeroiraz, line 869: it shouldn't be a warning. It's possible to use an external replicator.
A replicator may be initialized if the required data to connect with the master database are specified. Validation will return error if only partial information was provided, but otherwise it should just info that no replicator was initialized

arriqaaq: ack

PR #1390, opened 2022-09-26
chore(pkg/server): add logs for replicator

arriqaaq:
Addresses #1385 

- add logs for replicator if it is not required/configured

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -574,6 +574,7 @@ func (s *ImmuServer) replicationInProgressFor(db string) bool {
 
 func (s *ImmuServer) startReplicationFor(db database.DB, dbOpts *dbOptions) error {
 	if !dbOpts.isReplicatorRequired() {
+		s.Logger.Infof("Replication for database '%s' is not required.", db.GetName())
 		return ErrReplicatorNotNeeded
 	}
 


pkg/server/server.go
```diff
@@ -863,8 +863,12 @@ func (s *ImmuServer) CreateDatabaseV2(ctx context.Context, req *schema.CreateDat
 	s.logDBOptions(db.GetName(), dbOpts)
 
 	err = s.startReplicationFor(db, dbOpts)
-	if err != nil && err != ErrReplicatorNotNeeded {
-		return nil, fmt.Errorf("%w: while starting replication", err)
+	if err != nil {
+		if err == ErrReplicatorNotNeeded {
```
jeroiraz, line 867: it's better to compare errors using `errors.Is()` because it will cover the case the error is wrapped

pkg/server/server.go
```diff
@@ -934,9 +938,11 @@ func (s *ImmuServer) LoadDatabase(ctx context.Context, req *schema.LoadDatabaseR
 
 	s.dbList.Put(db)
 
-	if dbOpts.isReplicatorRequired() {
-		err = s.startReplicationFor(db, dbOpts)
-		if err != nil && err != ErrReplicatorNotNeeded {
+	err = s.startReplicationFor(db, dbOpts)
+	if err != nil {
+		if err == ErrReplicatorNotNeeded {
```
jeroiraz, line 943: same as above, probably the logging about not needed can be done inside the `startReplicationFor` method? so it will be only in one place

PR #1399, opened 2022-09-29
Add sync replication test for precommit state

arriqaaq:
It takes time for indexing the precommited transactions, so there is no way to guarantee the sync of states b/w master and followers by doing a Get/GetAt operation. This test checks if the pre commit state in master and followers are the same, by checking the precommit txn id on both master/follower to ensure synchronous replication is working as expected and data is stored on master.

### DIFF
--- a/pkg/integration/replication/synchronous_replication_test.go
+++ b/pkg/integration/replication/synchronous_replication_test.go
@@ -3,9 +3,11 @@ package replication
 import (
 	"context"
 	"fmt"
+	"sync"
 	"testing"
 	"time"
 
+	"github.com/codenotary/immudb/pkg/api/schema"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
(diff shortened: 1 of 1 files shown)


pkg/integration/replication/synchronous_replication_test.go
```diff
@@ -70,6 +72,61 @@ func (suite *SyncTestSuite) TestMasterRestart() {
...
+			ctx, client, cleanup := suite.ClientForReplica(followerID)
+			defer cleanup()
+			for {
+				select {
+				case <-startCh:
+					suite.Run(fmt.Sprintf("test replica sync state %d", followerID), func() {
+						state, err := client.CurrentState(ctx)
+						require.NoError(suite.T(), err)
+						suite.Require().Equal(state.PrecommittedTxId, masterState.PrecommittedTxId)
+						suite.Require().Equal(state.PrecommittedTxHash, masterState.PrecommittedTxHash)
```
byo, line 105: Shouldn't we check against Committed state on master ?

```suggestion
						suite.Require().Equal(state.PrecommittedTxId, masterState.TxId)
						suite.Require().Equal(state.PrecommittedTxHash, masterState.TxHash)
```

Actually at that point on the primary the Precomitted and Comitted should be the same (since we don't have any pending unconfirmed transaction) - we could check that right after getting `CurrentState` from primary.

byo (approved):

pkg/integration/replication/synchronous_replication_test.go
arriqaaq: ack, makes sense

byo (approved):

PR #1400, opened 2022-09-29
Feat sync followers its bart

byo:


pkg/integration/replication/suite.go
```diff
@@ -44,8 +44,9 @@ type baseReplicationTestSuite struct {
 	mu sync.Mutex
 
 	// server settings
-	master    TestServer
-	followers []TestServer
+	master           TestServer
+	followers        []TestServer
+	followersRunning []bool
```
arriqaaq, line 49: IMO, if we pop the follower from the list (treat it like a queue and remove the last item), we won't need this array. Else it is a lot of state to maintain. WDYT?

pkg/integration/replication/synchronous_replication_test.go
```diff
@@ -21,13 +21,9 @@ func TestSyncTestSuite(t *testing.T) {
 	suite.Run(t, &SyncTestSuite{})
 }
 
-// this function executes before the test suite begins execution
-func (suite *SyncTestSuite) SetupSuite() {
-	suite.baseReplicationTestSuite.SetupSuite()
+func (suite *SyncTestSuite) TestSyncFromMasterToAllFollowers() {
 	suite.SetupCluster(2, 2)
```
arriqaaq, line 25: This could be a problem with docker, with so many containers being spawned and closed for a test. If we want many followers, we can create diff test suites right? For ex, for TestMinimumFollowers, we can add a testsuite MinFollowerTestSuite. I think this function would be better in SetupSuite

pkg/integration/replication/suite.go
arriqaaq:
```
// get last item
f:=followers[len(followers)-1]
// pop
followers = followers[:len(followers]
```

something similar

byo: I wanted to keep follower on that list to allow starting it with the same data folder (and in some tests I want to explicitly start again the same instance).

The `followersRunning` added here is only a workaround for the TearDown method which I didn't notice earlier - because we don't allow double shutdowns.

pkg/integration/replication/synchronous_replication_test.go
byo: ACK, I agree that this could affect the performance.

Although that way we don't have a clean-page start for each test (like there could already be some transactions in the database). This can be fixed by creating new DB for each test later.

arriqaaq: in that scenario, this can be added to BeforeTest right?

arriqaaq: I think suite has a BeforeTest and AfterTest which executes before/after each test

byo: Fixed

pkg/integration/replication/suite.go
arriqaaq: ack, makes sense to know the state

arriqaaq (approved):

arriqaaq (approved):
really nice!

PR #1402, opened 2022-09-30
chore(pkg/util): add chain and retry packages

arriqaaq:
- add chain util which helps in chaining functions
- add retry util which helps in retrying functions with delay
- replace chunkedprocess with chain


embedded/appendable/remoteapp/options.go
```diff
@@ -26,20 +26,18 @@ type Options struct {
...
 }
 
 func DefaultOptions() *Options {
 	return &Options{
-		Options:          *multiapp.DefaultOptions(),
-		parallelUploads:  10,
-		retryMinDelay:    time.Second,
-		retryMaxDelay:    2 * time.Minute,
-		retryDelayExp:    2,
-		retryDelayJitter: 0.1,
```
byo, line 42: We're removing this jitter argument - is it replaced by something else?

embedded/appendable/remoteapp/remote_app.go
```diff
@@ -187,58 +179,47 @@ func (r *RemoteStorageAppendable) uploadChunk(chunkID int64, dontRemoveFile bool
 		defer metricsUploadFinished.Inc()
 
 		var newApp appendable.Appendable
-		cp := r.chunkedProcess(ctx)
+		cp := chain.New().WithContext(ctx).WithBackoff(r.backoff)
```
byo, line 182: :+1:

embedded/appendable/remoteapp/remote_app.go
```diff
@@ -187,58 +179,47 @@ func (r *RemoteStorageAppendable) uploadChunk(chunkID int64, dontRemoveFile bool
...
+				return nil
 			}
 
 			metricsUploadRetried.Inc()
-			return true, nil
-		})
-
-		// Open new appendable from the remote storage
-		cp.RetryableStep(func(retries int, delay time.Duration) (bool, error) {
+			return err
```
byo, line 212: This will work incorrectly if `err == nil && !exists` - at that situation we want to retry too since there may be some asynchronous processing on the remote storage and we want to ensure that the new object is already visible.

pkg/util/retry/backoff.go
```diff
@@ -0,0 +1,117 @@
...
+// NewExponentialRetrier returns an instance of ExponentialRetrier
+func NewExponentialRetrier(minTimeout, maxTimeout time.Duration, exponentFactor float64) Retrier {
+	return &exponentialRetrier{
+		factor:     exponentFactor,
+		minTimeout: minTimeout,
+		maxTimeout: maxTimeout,
+	}
+}
+
+// Next returns next time for retrying operation with exponential strategy
```
byo, line 48: Comments need to be updated

embedded/appendable/remoteapp/options.go
arriqaaq: I realised it wasn't needed, but have added it back to keep the jitter logic the same as you'd previously written

embedded/appendable/remoteapp/remote_app.go
arriqaaq: ack

pkg/util/retry/backoff.go
arriqaaq: ack

pkg/util/retry/backoff.go
```diff
@@ -0,0 +1,117 @@
...
+func (e *jitterRetrier) NextInterval(retryCount int) time.Duration {
+	if retryCount <= 0 {
+		return 0 * time.Millisecond
+	}
+
+	//calculate the jitter duration
+	minf := float64(e.minTimeout)
+	durf := minf * math.Pow(e.factor, float64(retryCount))
+	durf = rand.Float64()*(durf-minf) + minf
```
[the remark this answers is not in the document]
arriqaaq: I see your point, replacing it with existing logic

embedded/appendable/remoteapp/options.go
arriqaaq: fixed

embedded/appendable/remoteapp/remote_app.go
arriqaaq: Thank you :)

embedded/appendable/remoteapp/remote_app.go
```diff
@@ -187,58 +182,52 @@ func (r *RemoteStorageAppendable) uploadChunk(chunkID int64, dontRemoveFile bool
...
 
 			metricsUploadRetried.Inc()
-			return true, nil
-		})
 
-		// Open new appendable from the remote storage
-		cp.RetryableStep(func(retries int, delay time.Duration) (bool, error) {
+			// Retry since there may be some asynchronous processing on the remote
+			// storage and we want to ensure that the new object is already visible.
+			if err == nil && !exists {
```
arriqaaq, line 217: @byo does this step fix this comment?

https://github.com/codenotary/immudb/pull/1402/files#r984674900

pkg/util/retry/backoff.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// NextInterval returns the next time interval to retrying the operation
+func (e *exponentialRetrier) NextInterval(retryCount int) time.Duration {
+	if retryCount <= 0 {
+		return 0 * time.Millisecond
+	}
+
+	efac := math.Pow(e.factor, float64(retryCount)) * float64(e.minTimeout)
+	sleep := math.Min(efac, float64(e.maxTimeout)) * (1.0 - rand.Float64()*e.jitterFactor)
```
arriqaaq, line 59: @byo keeping the jitter calculation the same for exponential backoff

embedded/appendable/remoteapp/remote_app.go
byo: Looks like it solves this case. The code may even be simplified (to expose the happy path better):

```go
		}).NextWithRetry(func() error { // Wait for the chunk to become ready
			exists, err := r.rStorage.Exists(ctx, r.remotePath+appName)
			if err != nil {
				metricsUploadRetried.Inc()
				return err
			}

			if !exists {
				// Retry since there may be some asynchronous processing on the remote
				// storage and we want to ensure that the new object is already visible.
				return retry.ErrRetry
			}

			return nil
		}).NextWithRetry(func() error { // Open new appendable from the remote storage
```

The only difference to the current code is the `metricsUploadRetried.Inc()` - but now I think it does not need to be updated when the remote storage is still processing the upload and not yet exposes the uploaded blob. We're interested in other errors though.

embedded/appendable/remoteapp/remote_app.go
```diff
@@ -324,36 +310,32 @@ func (r *RemoteStorageAppendable) downloadChunk(chunkID int64) {
...
-		cp := r.chunkedProcess(ctx)
-		cp.RetryableStep(func(retries int, delay time.Duration) (retryNeeded bool, err error) {
-
+		cp := chain.New().WithContext(ctx).WithBackoff(r.backoff)
+		cp.NextWithRetry(func() (err error) {
 			data, err := r.rStorage.Get(ctx, r.remotePath+r.appendableName(chunkID), 0, -1)
 			if err != nil {
 				metricsDownloadRetried.Inc()
-				return true, nil
+				return nil
```
byo, line 318:
```suggestion
				return err
```

embedded/appendable/remoteapp/remote_app.go
```diff
@@ -324,36 +310,32 @@ func (r *RemoteStorageAppendable) downloadChunk(chunkID int64) {
...
 			}
 			defer data.Close()
 
 			flTmp, err = os.OpenFile(fileNameTmp, os.O_CREATE|os.O_RDWR, r.fileMode)
 			if err != nil {
 				// Couldn't create temporary local file, something is broken on local FS
-				return false, err
+				return err
```
byo, line 325: This is different behavior from the previous implementation.

In case the local temporary file can not be created, we don't even want to retry this. Instead the whole chain should fail with an error without retries.

I think that currently there's no way to break retrying - any error returned will cause retry, correct? And returning nil will continue with the next step.

byo: Maybe we should experiment with wrapping errors?

So in order to retry when error happens e.g.:

```go
if err != nil {
  if error.Is(err, ErrThatCanBeRetried) {
    // Will retry after backoff timeout
    return retry.WithError(err)
  }
  // Will not retry and fail the whole chain
  return err
}

if shouldRetryAnyway {
  // Will retry after backoff timeout
  return retry.WithoutError()
}

// This will not retry and will continue with the following step
return nil
```

The `retry.WithError` and `retry.WithoutError` could then return a dedicated error declared in the `retry` package, it could also be extended in the future with things like whether we want to skip the backoff sleep or set the maximum number of retries allowed etc.:

```go
return retry.WithError(err).SkipBackoffSleep()

return retry.WithoutError().MaxRetries(10)
```

^ that I would keep for future updates to avoid adding code that we don't currently use.

embedded/appendable/remoteapp/remote_app.go
arriqaaq: Ack, thanks for the pointer, I'll fix this

embedded/appendable/remoteapp/remote_app.go
arriqaaq: Actually the `Next` step will return error and block other steps without retrying, I've kept the functionality as close as possible to chunkedProcess, but I'll have a look at this

embedded/appendable/remoteapp/remote_app.go
arriqaaq: resolved

embedded/appendable/remoteapp/remote_app.go
arriqaaq: I've just added a different step to the chain function, and added a retriable function, where we can specify which error to retry on, eg:

```go
		cp := chain.New().WithContext(ctx).WithBackoff(r.backoff).WithRetriable(func(err error) bool {
			switch {
			case err == retry.ErrNoRetry:
				return false
			default:
				return true
			}
		})
		cp.NextWithRetry(func() (err error) {
			data, err := r.rStorage.Get(ctx, r.remotePath+r.appendableName(chunkID), 0, -1)
			if err != nil {
				metricsDownloadRetried.Inc()
				return err
			}
			defer data.Close()

			flTmp, err = os.OpenFile(fileNameTmp, os.O_CREATE|os.O_RDWR, r.fileMode)
			if err != nil {
				// Couldn't create temporary local file, something is broken on local FS
				return retry.ErrNoRetry
			}

			_, err = io.Copy(flTmp, data)
			if err != nil {
				metricsDownloadRetried.Inc()
				return nil
			}
			return nil
		})
```

pkg/util/retry/retry.go
```diff
@@ -0,0 +1,126 @@
...
+
+var (
+	// ErrRetryAgain is an error used to retry on a fail condition
+	ErrRetryAgain = errors.New("retry again")
+	// ErrNoRetry is an error used to not retry on error
+	ErrNoRetry = errors.New("do not retry on error")
+)
+
+// Retriable returns true if the error has to be retried
+type Retriable func(error) bool
```
byo, line 47: I would go for a single method for determination if something is retryable or not.

My suggestion here is:

* if the returned value is `ErrRetryAgain` - this should be treated as now - just retry the step
* introduce something like `errNoRetry` struct (and a method `ErrNoRetry` to create its instance) that wraps another error - that way in retryable step one can return an error that will break the chain and will not be retried
* optional: introduce something like `errStop` struct that will stop the chain but the result of chain execution will not be an error - this one is not currently needed thus I'm not sure if it's worth implementing it now (and it will make the chain implementation a bit more complex) - but the main idea is that it should be relatively easy to add if needed

pkg/util/retry/retry.go
```diff
@@ -0,0 +1,126 @@
...
+				return nil
+			case retriable(err): // if a retriable error, continue
+				lastErr = err
+			default: // if err is not handled by retriable, return without retrying
+				return nil
+			}
+			if backoff.Steps == 1 {
+				break LOOP
+			}
+			time.Sleep(backoff.Step())
```
byo, line 100: This must also be executed in parallel to the context. Otherwise we risk that it will be forced to wait for the full delay before things get cancelled. This may for example cause serious issues when shutting down PODs on kubernetes - where the's a sort kill followed by hard kill. Increasing the time needed for clean shutdown may result in hard termination.

Good snippet for such sleep would be:

```go
timer := time.NewTimer(backoff.Step())
select {
  case <-timer.C:
  case <-backoff.ctx.Done():
}
timer.Stop()
```

PR #1408, opened 2022-10-04
Test sync replication bart

byo:


embedded/ahtree/ahtree.go
```diff
@@ -37,7 +38,7 @@ var ErrInvalidOptions = fmt.Errorf("%w: invalid options", ErrIllegalArguments)
 var ErrorPathIsNotADirectory = errors.New("ahtree: path is not a directory")
 var ErrorCorruptedData = errors.New("ahtree: data log is corrupted")
 var ErrorCorruptedDigests = errors.New("ahtree: hash log is corrupted")
-var ErrAlreadyClosed = errors.New("ahtree: already closed")
```
arriqaaq, line 40: We may want to split this PR into two, one for handling the error code changes, and the other for the test suite changes

pkg/integration/replication/suite.go
```diff
@@ -214,17 +281,29 @@ func (suite *baseReplicationTestSuite) WaitForCommittedTx(
 	}
 }
 
-func (suite *baseReplicationTestSuite) SetupCluster(replicas int, syncFollowers int) {
-	suite.StartMaster(syncFollowers)
+func (suite *baseReplicationTestSuite) SetupCluster(syncReplicas, syncAcks, asyncReplicas int) {
```
arriqaaq, line 284: wdyt of adding syncReplicas and asyncReplicas in the baseReplicationTestSuite as params? Since we are calling SetupCluster always inside SetupSuite, it will make it simpler

pkg/integration/replication/suite.go
```diff
@@ -214,17 +281,29 @@ func (suite *baseReplicationTestSuite) WaitForCommittedTx(
...
+	suite.masterDBName = masterDBName
+
+	suite.StartMaster(syncAcks)
 
 	wg := sync.WaitGroup{}
-	for i := 0; i < replicas; i++ {
+
+	for i := 0; i < syncReplicas; i++ {
```
arriqaaq, line 291: Can't we have one bool parameter inside the suite, and run it in either sync or async mode? I don't see a test scenario where both sync and async need to exist, unless we plan to test those complex topologies?

arriqaaq: it makes it bit hacky, like for sync scenarios, one has to always imply 0 async, takes away the simplicity from the suite IMO. WDYT?

```
suite.SetupCluster(2, 2, 0)
```

byo:
This was only an assisting PR, the issue was found and resolved in another PR

PR #1409, opened 2022-10-04
fix(pkg/replication): solve issues when follower diverged from master

jeroiraz:

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -1316,33 +1316,31 @@ func (d *db) ExportTxByID(req *schema.ExportTxRequest) (txbs []byte, mayCommitUp
 		}
 	}
 
-	// it might be the case master will commit some txs
-	// current timeout it's not a special value but at least a relative one
-	// note: master might also be waiting ack from any follower (even this follower may do progress)
-
-	// TODO: under some circustances, follower might not be able to do further progress until master
-	// has made changes, such wait doesn't need to have a timeout, reducing networking and CPU utilization
-	ctx, cancel := context.WithTimeout(context.Background(), d.options.storeOpts.SyncFrequency*4)
-	defer cancel()
-	cancellation := ctx.Done()
-
 	// the follower requested a non-existent or non-ready transaction
-	// follower commits are prioritized over passive waiting for the requested transaction
-	if (req.AllowPreCommitted && req.Tx > preCommittedTxID) ||
-		(!req.AllowPreCommitted && req.Tx > committedTxID) {
-
+	if (req.AllowPreCommitted && req.Tx > preCommittedTxID) || (!req.AllowPreCommitted && req.Tx > committedTxID) {
+		// follower commits are prioritized over passive waiting for the requested transaction
 		// follower committed state lies behind master committed state
+		// TODO: evaluate if it's better to remove this special case and just wait for the requested tx (faster replication)
 		if req.FollowerState != nil && req.FollowerState.CommittedTxID < committedTxID {
 			return nil, mayCommitUpToTxID, mayCommitUpToAlh, nil
 		}
-	}
 
-	err = d.WaitForTx(req.Tx, req.AllowPreCommitted, cancellation)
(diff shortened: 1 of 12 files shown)


pkg/integration/replication/synchronous_replication_test.go
```diff
@@ -125,21 +122,21 @@ func (suite *SyncTestSuite) TestPrecommitStateSync() {
...
-				case <-startCh:
-					suite.Run(fmt.Sprintf("test replica sync state %d", followerID), func() {
-						state, err := client.CurrentState(ctx)
-						require.NoError(suite.T(), err)
-						suite.Require().Equal(state.PrecommittedTxId, masterState.TxId)
-						suite.Require().Equal(state.PrecommittedTxHash, masterState.TxHash)
-					})
-					return
-				}
+			for range startCh {
```
arriqaaq, line 125: This would not run at all because I don't think range checks for nil

https://go.dev/play/p/pJ1sU5I9x6n

pkg/integration/replication/synchronous_replication_test.go
```diff
@@ -125,21 +122,21 @@ func (suite *SyncTestSuite) TestPrecommitStateSync() {
...
+				suite.Run(fmt.Sprintf("test replica sync state %d", followerID), func() {
+					state, err := client.CurrentState(ctx)
+					require.NoError(suite.T(), err)
+					suite.Require().Equal(state.PrecommittedTxId, masterState.TxId)
+					suite.Require().Equal(state.PrecommittedTxHash, masterState.TxHash)
+				})
+				return
```
byo:
```suggestion
			<-startCh
			suite.Run(fmt.Sprintf("test replica sync state %d", followerID), func() {
				state, err := client.CurrentState(ctx)
				require.NoError(suite.T(), err)
				suite.Require().Equal(state.PrecommittedTxId, masterState.TxId)
				suite.Require().Equal(state.PrecommittedTxHash, masterState.TxHash)
			})
```

pkg/integration/replication/synchronous_replication_test.go
```diff
@@ -102,59 +116,49 @@ func (suite *SyncTestSuite) TestMasterRestart() {
...
-
+func (suite *SyncTestSuitePrecommitStateSync) TestPrecommitStateSync() {
 	ctx, client, cleanup := suite.ClientForMaster()
 	defer cleanup()
 
-	// Create goroutines for client waiting to query the state
-	// of the followers. This is initialised before to avoid
-	// spending time initialising the follower client for faster
-	// state access
-	var wg sync.WaitGroup
```
byo, line 121: This should be restored

pkg/integration/replication/suite.go
```diff
@@ -305,6 +308,20 @@ func (suite *baseReplicationTestSuite) SetupCluster(syncReplicas, syncAcks, asyn
 	}
 
 	wg.Wait()
+
+	uuids := make(map[string]struct{}, 1+suite.GetFollowersCount())
+
+	uuids[suite.master.UUID(suite.T()).String()] = struct{}{}
+
+	for _, f := range suite.followers {
+		uuid := f.UUID(suite.T()).String()
```
byo, line 317: This disables ability to run tests with docker runner - every cluster setup will panic - maybe we should not be so strict here?

pkg/database/database.go
```diff
@@ -1318,7 +1316,12 @@ func (d *db) ExportTxByID(req *schema.ExportTxRequest) (txbs []byte, mayCommitUp
 		}
 	}
 
-	var cancellation <-chan struct{}
+	// it might be the case master will commit some txs
+	// current timeout it's not a special value but at least a relative one
+	// note: master might also be waiting ack from any follower (even this follower may do progress)
+	ctx, cancel := context.WithTimeout(context.Background(), d.options.storeOpts.SyncFrequency*4)
```
byo, line 1322: Please add a TODO comment here just in case ;)

byo (dismissed):
Few smaller comments left, other than that LGTM

PR #1412, opened 2022-10-05
fix(Makefile): remove webconsole tag from immuclient/immuadmin builds

arriqaaq:
This calls subsequent make commands, which is a downside

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,7 @@ DOCKER ?= docker
 PROTOC ?= protoc
 STRIP = strip
 
+
 V_COMMIT := $(shell git rev-parse HEAD)
 #V_BUILT_BY := "$(shell echo "`git config user.name`<`git config user.email`>")"
 V_BUILT_BY := $(shell git config user.email)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -228,9 +232,9 @@ dist/binaries:
     		for os_arch in ${TARGETS}; do \
     			goos=`echo $$os_arch|sed 's|/.*||'`; \
     			goarch=`echo $$os_arch|sed 's|^.*/||'`; \
-    		    GOOS=$$goos GOARCH=$$goarch $(GO) build -tags webconsole -v -ldflags '${V_LDFLAGS_COMMON}' -o ./dist/$$service-v${VERSION}-$$goos-$$goarch ./cmd/$$service/$$service.go ; \
+				GOOS=$$goos GOARCH=$$goarch V_BUILD_NAME=./dist/$$service-v${VERSION}-$$goos-$$goarch make $$service ; \
```
byo, line 235: How about some computed variable names? https://www.gnu.org/software/make/manual/html_node/Computed-Names.html

That way we could have some variable like `V_BUILD_FLAGS_immudb` and access it with `$(V_BUILD_FLAGS_$$service)`?

arriqaaq: Thanks for the suggestion! I will try this out

PR #1416, opened 2022-10-06
fix heartbeater keepAlive context, add errorHandler to notify upper layers on keepAlive error

marcosQuesada:
- PR created in top of v.1.3.2 release
- Fix Heartbeater KeepAlive context, original context has 60 seconds timeout scope, use context.Background as context base
- Add ErrorHandler callback to allow upper layers react on Heartbeater KeepAlive failure

### DIFF
--- a/pkg/client/heartbeater/heartbeater.go
+++ b/pkg/client/heartbeater/heartbeater.go
@@ -42,14 +42,14 @@ type HeartBeater interface {
 	Stop()
 }
 
-func NewHeartBeater(sessionID string, sc schema.ImmuServiceClient, keepAliveInterval time.Duration, h ErrorHandler) *heartBeater {
+func NewHeartBeater(sessionID string, sc schema.ImmuServiceClient, keepAliveInterval time.Duration, errhandler ErrorHandler) *heartBeater {
 	return &heartBeater{
 		sessionID:     sessionID,
 		logger:        logger.NewSimpleLogger("immuclient", stdos.Stdout),
 		serviceClient: sc,
 		done:          make(chan struct{}),
 		t:             time.NewTicker(keepAliveInterval),
-		errorHandler:  h,
+		errorHandler:  errhandler,
 	}
 }
 
@@ -65,10 +65,9 @@ func (hb *heartBeater) KeepAlive(ctx context.Context) {
 				err := hb.keepAliveRequest(ctx)
 				if err != nil {
 					hb.logger.Errorf("an error occurred on keep alive %s at %s: %v\n", hb.sessionID, t.String(), err)
-					if hb.errorHandler == nil {
-						continue
+					if hb.errorHandler != nil {
+						hb.errorHandler(hb.sessionID, err)
 					}
-					hb.errorHandler(hb.sessionID, err)
 				}
 			}
 		}
--- a/pkg/client/heartbeater/heartbeater.go
+++ b/pkg/client/heartbeater/heartbeater.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/codenotary/immudb/pkg/api/schema"
+	"github.com/codenotary/immudb/pkg/client"
 	"github.com/codenotary/immudb/pkg/logger"
 	"github.com/golang/protobuf/ptypes/empty"
 )
@@ -32,6 +33,7 @@ type heartBeater struct {
 	serviceClient schema.ImmuServiceClient
 	done          chan struct{}
 	t             *time.Ticker
+	errorHandler  client.ErrorHandler
 }
 
 type HeartBeater interface {
@@ -39,13 +41,14 @@ type HeartBeater interface {
 	Stop()
 }
 
-func NewHeartBeater(sessionID string, sc schema.ImmuServiceClient, keepAliveInterval time.Duration) *heartBeater {
+func NewHeartBeater(sessionID string, sc schema.ImmuServiceClient, keepAliveInterval time.Duration, h client.ErrorHandler) *heartBeater {
 	return &heartBeater{
 		sessionID:     sessionID,
 		logger:        logger.NewSimpleLogger("immuclient", stdos.Stdout),
 		serviceClient: sc,
 		done:          make(chan struct{}),
 		t:             time.NewTicker(keepAliveInterval),
+		errorHandler:  h,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


pkg/client/session.go
```diff
@@ -105,3 +109,11 @@ func (c *immuClient) CloseSession(ctx context.Context) error {
 func (c *immuClient) GetSessionID() string {
 	return c.SessionID
 }
+
+func (c *immuClient) keepAliveErrorHandler(sessionID string, err error) {
+	if err == sessions.ErrSessionNotFound || err == ErrNotConnected || strings.Contains(err.Error(), "session not found") {
+		c.Logger.Errorf("heartbeater KeepAlive error, sessionID %s error %v", sessionID, err)
+
+		panic(fmt.Errorf("heartbeater KeepAlive error, sessionID %s error %v", sessionID, err))
```
ldej, line 117: This is just for us to test this situation right? Because I don't think we want to have a panic in the client.

marcosQuesada: the intention is to grab info about the immudb connection state

pkg/client/session.go
```diff
@@ -80,7 +82,8 @@ func (c *immuClient) OpenSession(ctx context.Context, user []byte, pass []byte,
 	c.Options.DialOptions = dialOptions
 	c.SessionID = resp.GetSessionID()
 
-	c.HeartBeater = heartbeater.NewHeartBeater(c.SessionID, c.ServiceClient, c.Options.HeartBeatFrequency)
+	c.WithErrorHandler(c.keepAliveErrorHandler)
```
jeroiraz, line 85: isn't this line a NOP in the end? i.e setting the same value

jeroiraz, line 85: isn't it overwriting the error handler that could be externally specified?

pkg/client/session.go
jeroiraz: it shouldn't panic... in case it's desired such error handler could be externally specified

pkg/client/heartbeater/heartbeater.go
```diff
@@ -61,6 +65,10 @@ func (hb *heartBeater) KeepAlive(ctx context.Context) {
 				err := hb.keepAliveRequest(ctx)
 				if err != nil {
 					hb.logger.Errorf("an error occurred on keep alive %s at %s: %v\n", hb.sessionID, t.String(), err)
+					if hb.errorHandler == nil {
```
jeroiraz, line 68: it might be better
```
if hb.errorHandler == nil {
    hb.errorHandler(hb.sessionID, err)
}
```
because then if needed further lines could be added latter on without a dependency on the error handler being set

arriqaaq: you mean `if hb.errorHandler != nil` right?

pkg/client/session.go
jeroiraz: the error handler can be already specified `func (c *immuClient) WithErrorHandler(handler heartbeater.ErrorHandler) *immuClient {`, but I think here is overwritten with `c.WithErrorHandler(c.keepAliveErrorHandler)`. Furthermore, `keepAliveErrorHandler` is panicking if there is an error and it shouldn't

pkg/client/session.go
```diff
@@ -123,3 +125,9 @@ func (c *immuClient) CloseSession(ctx context.Context) error {
 func (c *immuClient) GetSessionID() string {
 	return c.SessionID
 }
+
+func (c *immuClient) keepAliveErrorHandler(sessionID string, err error) {
```
jeroiraz, line 129: dead code?

arriqaaq: removed

pkg/client/session.go
arriqaaq: resolved

pkg/client/types.go
```diff
@@ -75,3 +76,8 @@ func (c *immuClient) WithOptions(options *Options) *immuClient {
 	c.Options = options
 	return c
 }
+
+func (c *immuClient) WithErrorHandler(handler heartbeater.ErrorHandler) *immuClient {
```
jeroiraz, line 80: error handler should not be in package heartbeater, in the future more client errors can be handled and not just the ones related to heartbeats

arriqaaq: This is error handler for heartbeater exclusively, if defined on client level, leads to cyclic dependency so

jeroiraz: `func (c *immuClient) WithErrorHandler(handler ErrorHandler) `  client api should not expose internals. I think either the error handler is defined in the client package or `func (c *immuClient) WithHeartBeater(...)` is directly set as a component. But I'd think the first alternative is better as it gives the chance to handle more errors and not just those related to heartbeats...

jeroiraz: func (c *immuClient) WithErrorHandler(handler ErrorHandler)  client api should not expose internals. I think either the error handler is defined in the client package or func (c *immuClient) WithHeartBeater(...) is directly set as a component. But I'd think the first alternative is better as it gives the chance to handle more errors and not just those related to heartbeats...
regarding cyclic dependencies, probably it's not worth to have a hearbeater package..

pkg/client/session.go
```diff
@@ -128,7 +128,7 @@ func (c *immuClient) GetSessionID() string {
 }
 
 func (c *immuClient) keepAliveErrorHandler(sessionID string, err error) {
-	if err != sessions.ErrSessionNotFound && err != ErrNotConnected && !strings.Contains(err.Error(), "session not found") {
+	if err == sessions.ErrSessionNotFound || err == ErrNotConnected || strings.Contains(err.Error(), "session not found") {
```
jeroiraz: we should compare errors with errors.Is(...)

pkg/client/heartbeater/heartbeater.go
```diff
@@ -22,26 +22,27 @@ import (
...
+	errorHandler  ErrorHandler
 }
 
 type HeartBeater interface {
 	KeepAlive(ctx context.Context)
 	Stop()
 }
 
-func NewHeartBeater(sessionID string, sc schema.ImmuServiceClient, keepAliveInterval time.Duration, h client.ErrorHandler) *heartBeater {
+func NewHeartBeater(sessionID string, sc schema.ImmuServiceClient, keepAliveInterval time.Duration, h ErrorHandler) *heartBeater {
```
jeroiraz: errHandler param name

jeroiraz (dismissed):

jeroiraz:
it would be good to have a unit test to validate an error is forwarded to the provisioned error handler

jeroiraz (approved):

PR #1418, opened 2022-10-06
Fix follower diverged prefetch

jeroiraz:


pkg/database/database.go
```diff
@@ -1331,15 +1329,15 @@ func (d *db) ExportTxByID(req *schema.ExportTxRequest) (txbs []byte, mayCommitUp
 		}
 
 		// it might be the case master will commit some txs
-		// current timeout it's not a special value but at least a relative one
-		// note: master might also be waiting ack from other followers
+		// current timeout it's not a special value but at least a r8elative one
```
byo, line 1332:
```suggestion
		// current timeout it's not a special value but at least a relative one
```

pkg/integration/replication/synchronous_replication_test.go
```diff
@@ -102,59 +116,49 @@ func (suite *SyncTestSuite) TestMasterRestart() {
...
+func (suite *SyncTestSuitePrecommitStateSync) TestPrecommitStateSync() {
 	ctx, client, cleanup := suite.ClientForMaster()
 	defer cleanup()
 
-	// Create goroutines for client waiting to query the state
-	// of the followers. This is initialised before to avoid
-	// spending time initialising the follower client for faster
-	// state access
-	var wg sync.WaitGroup
-	for i := 0; i < suite.GetFollowersCount(); i++ {
```
byo, line 122: This should be restored - it's about reading replicas state as soon as possible right after the commit is confirmed.

PR #1426, opened 2022-10-12
chore: Rename sync-followers to sync-acks

byo:


arriqaaq (dismissed):

pkg/server/db_options.go
```diff
@@ -675,15 +675,15 @@ func (opts *dbOptions) Validate() error {
...
 			return fmt.Errorf(
 				"%w: invalid replication options for master database '%s'. It is necessary to have at least one sync follower",
 				ErrIllegalArguments, opts.Database)
 		}
 
-		if !opts.SyncReplication && opts.SyncFollowers > 0 {
+		if !opts.SyncReplication && opts.SyncAcks > 0 {
 			return fmt.Errorf(
-				"%w: invalid replication options for master database '%s'. Sync followers are not expected",
+				"%w: invalid replication options for master database '%s'. Sync acks is not expected",
```
jeroiraz, line 686:
```suggestion
				"%w: invalid replication options for master database '%s'. SyncAcks is not expected",
```

jeroiraz (approved):
LGTM

I'd think we could also rename `follower-username` and `follower-password`,  having master and replica seems to be simpler now. That change may be done after the release and as part of a bigger change when adding mtls support for replica authentication...

jeroiraz:
There is already a PR with mtls support for replication  #1352 by @zauberhaus

PR #1433, opened 2022-10-24
feat(ci): fix message and input

marcosanchotene:
- Add missing line breaks to Mattermost message
- Extract Go version from environment variable to pass it to reusable workflow

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -172,15 +172,28 @@ jobs:
           -total-entries-written 300000 \
           -total-entries-read 10000
 
+# This job is needed because currently it's not possible to pass an environment variable
+# to the called workflow on job performance-tests.
+# Reference: https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations
+  go-version:
+    name: Extract Go version
+    runs-on: ubuntu-latest
+    outputs:
+      go-version: ${{ steps.extraction.outputs.go_version }}
+    steps:
+      - id: extraction
+        run: echo "go_version=$GO_VERSION" >> $GITHUB_OUTPUT
+
   performance-tests:
     name: Performance tests
     needs:
       - gosec
       - old-go
+      - go-version
     uses: ./.github/workflows/performance.yml
     secrets: inherit
     with:
-      go-version: ${{ env.GO_VERSION }}
+      go-version: ${{ needs.go-version.outputs.go-version }}
 
   notarize-binaries:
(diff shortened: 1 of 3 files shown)


.github/workflows/push.yml
```diff
@@ -172,15 +172,25 @@ jobs:
...
     name: Performance tests
     needs:
       - gosec
       - old-go
+      - go-version
     uses: ./.github/workflows/performance.yml
     secrets: inherit
     with:
-      go-version: ${{ env.GO_VERSION }}
+      go-version: ${{ needs.go-version.outputs.go-version }}
```
byo, line 193: Shouldn't this be:

```suggestion
      go-version: ${{ needs.go-version.outputs.go_version }}
```
?

.github/workflows/push.yml
```diff
@@ -172,15 +172,25 @@ jobs:
           -total-entries-written 300000 \
           -total-entries-read 10000
 
+  go-version:
```
byo, line 175: Can you explain a bit more why this is needed?

.github/workflows/push.yml
marcosanchotene: No. The second `go-version` on line 193 refers to the `output` property of the file on line 179, not to the variable on line 182. I know it's a bit confusing, but I wanted to have an underscore on the variable name.

.github/workflows/push.yml
marcosanchotene: It's a workaround to pass the `$GO_VERSION` environment variable to the called workflow, as this is currently not supported. Please read [this](https://github.com/orgs/community/discussions/26671#discussioncomment-3252793) for reference.

byo: Ok, got it :+1:.

Can you please add a bit of a comment for this step with the link to mentioned issue?
Looking at that code itself may be confusing for others in the future.

marcosanchotene: Done, but I added a [link to the official documentation](https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations).

byo: Hi @marcosanchotene - I don't see the comment anywhere, can you check if the correct commit was pushed?

marcosanchotene: Yes, I made a mistake here. It's there now. Sorry!

byo (approved):

byo (approved):

PR #1448, opened 2022-11-08
Fix/replication naming convention

byo:


pkg/api/schema/schema.proto
```diff
@@ -1016,19 +1016,19 @@ message ReplicationNullableSettings {
   NullableBool replica = 1;
 
   // Name of the database to replicate
-  NullableString masterDatabase = 2;
+  NullableString primaryDatabase = 2;
```
jeroiraz, line 1019: hm, we may need to revise sdks as they may be exposing this types to the final app...

jeroiraz (approved):
LGTM, only one backward compatibility concern related to proto spec changes

arriqaaq (dismissed):

pkg/api/schema/schema.proto
byo: This will require updates on the client code if the sdk version was updated, but this does not break wire protocol compatibility.

I did test with various scenarios - e.g. connected with newer immuadmin to older immudb server or a cluster with mixed old and new immudb versions and there were no issues.

PR #1449, opened 2022-11-08
chore(cmd/immudb): Allow resetting sysadmin password

byo:


pkg/server/user.go
```diff
@@ -547,7 +547,7 @@ func (s *ImmuServer) getValidatedUser(username []byte, password []byte) (*auth.U
 }
 
 // getUser returns userdata (username,hashed password, permission, active) from username
-func (s *ImmuServer) getUser(username []byte, includeDeactivated bool) (*auth.User, error) {
```
jeroiraz, line 550: was not used to reactivate a user?

cmd/immudb/command/init.go
```diff
@@ -58,6 +58,7 @@ func (cl *Commandline) setupFlags(cmd *cobra.Command, options *server.Options) {
 	cmd.Flags().String("clientcas", "", "clients certificates list. Aka certificate authority")
 	cmd.Flags().Bool("devmode", options.DevMode, "enable dev mode: accept remote connections without auth")
 	cmd.Flags().String("admin-password", options.AdminPassword, "admin password (default is 'immudb') as plain-text or base64 encoded (must be prefixed with 'enc:' if it is encoded)")
+	cmd.Flags().Bool("force-admin-password", false, "if true, reset the admin password is to the one passed through `admin-password` option if it was changed")
```
jeroiraz, line 61: what about always overwriting it if `admin-password` is specified? otherwise it will just use current one (or default if the user needs to be created). So we don't need to add an extra flag

byo: That would be a pretty nasty backwards incompatibility. There could be some services out there where the admin password is set on the first run but the option is still there in subsequent runs (e.g. to avoid having a dedicated first-run operation).

A separate option means we can maintain backwards compatibility.

pkg/server/user.go
byo: All places in the code call this function with the `true` value. There's only a call in a unit test that uses `false` here.

I decided to remove it though for another reason. The code is actually completely broken since the condition in line 568 should be `if !usr.Active` (Which I found while testing sysadmin password reset). So since it's not used, broken and incorrectly tested, removing it seemed like a logical thing to do.

jeroiraz (dismissed):

Issue #1436, opened 2022-10-26
Test improvements: prefer `require` to `assert`


byo:
**What would you like to be added or enhanced**

In our test suite we should prefer `require` to `assert`. `require` will stop the test if the precondition fails which is what we expect in most of the cases.

This change can not be applied automatically though since there may be places where continuing the test if the assertion fails would still be beneficial (e.g. to get better failure context).

Also the usage of `require` on goroutines spawned by the test should be careful. This will stop the goroutine but the test itself will continue to run which may be unexpected.

**Why is this needed**

We should fail tests early which is what `require` provides.

Nowikens:
Hi, I would like to take this issue. I should mention that this would be my first open source issue, and I'm fairly new to golang. Is there a good file to start with?

jeroiraz:
> Hi, I would like to take this issue. I should mention that this would be my first open source issue, and I'm fairly new to golang. Is there a good file to start with?

Hi @Nowikens, it would be great.

Under embedded package all test files rely on require but you may find unit tests under pkg/server or pkg/client still using assert.

E.g. embedded/htree/htree_test.go

Issue #1435, opened 2022-10-26
Tests improvements: Better check for errors


byo:
**What would you like to be added or enhanced**

In many tests we're using `require.Error(t, err)` which only checks if the error happened without checking if the error is actually expected.

We should add more detailed error check, either by using `require.ErrorIs` or `require.ErrorContains` assertions.

**Why is this needed**

When expecting an error it is necessary to ensure the returned error is what was expected. It may turn out that in some cases the error checked by the test is not what it should be.

Nowikens:
I would Like to take on this issue.

Nowikens:
I'm still working on it, but I noticed that many tested functions are returning `ErrNotLoggedIn` when failing in test. Is this expected behaviour? It seems like tested functions are failing on calling `getDBFromCtx` inside them.

FYI I am running and debugging tests via vscode's test runner.

<details>
  <summary>Some of test functions where I encountered this situation:</summary>

- TestExportTxEdgeCases
- testServerSetGetError
- testServerCurrentRootError
- testServerSetGetBatchError
- testServerByIndexError
- testServerBySafeIndexError
- testServerHistoryError
- testServerZAddError
- testServerScanError
- testServerSafeReferenceError
- TestServerErrors
- TestSQLInteraction
</details>

PR #1456, opened 2022-11-16
Replaced assert with require in pkg/

Nowikens:
Replaced assert with require as mentioned in [this issue](https://github.com/codenotary/immudb/issues/1436).
As this is my first contribution to your project and to open source at all I'm making this pull request only in scope of `pkg/`.

I intentionally omitted replacement in `pkg/server/sessions/options_test.go` because I would argue that it's beneficial to see all errors related to configuration at once instead of only the first that is raised. I did not revert existing `require` to `assert` tho.

I'm open for suggestions, and if pull is accepted I'm happy to contribute further.

### DIFF
--- a/pkg/server/errors_test.go
+++ b/pkg/server/errors_test.go
@@ -23,20 +23,20 @@ import (
 
 	"github.com/codenotary/immudb/embedded/store"
 	immuerrors "github.com/codenotary/immudb/pkg/errors"
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestMapServerError(t *testing.T) {
 	err := mapServerError(store.ErrIllegalState)
-	assert.Equal(t, ErrIllegalState, err)
+	require.Equal(t, ErrIllegalState, err)
 
 	err = mapServerError(store.ErrIllegalArguments)
-	assert.Equal(t, ErrIllegalArguments, err)
+	require.Equal(t, ErrIllegalArguments, err)
 
 	someError := errors.New("some error")
 	err = mapServerError(someError)
-	assert.Equal(t, someError, err)
+	require.Equal(t, someError, err)
 
 	err = mapServerError(fmt.Errorf("%w: test", store.ErrPreconditionFailed))
-	assert.Equal(t, immuerrors.CodIntegrityConstraintViolation, err.(immuerrors.Error).Code())
+	require.Equal(t, immuerrors.CodIntegrityConstraintViolation, err.(immuerrors.Error).Code())
 }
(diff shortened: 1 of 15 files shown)


pkg/server/errors_test.go
```diff
@@ -23,20 +23,20 @@ import (
...
 	"github.com/codenotary/immudb/embedded/store"
 	immuerrors "github.com/codenotary/immudb/pkg/errors"
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestMapServerError(t *testing.T) {
 	err := mapServerError(store.ErrIllegalState)
-	assert.Equal(t, ErrIllegalState, err)
+	require.Equal(t, ErrIllegalState, err)
```
jeroiraz, line 31: it may be better to improve error comparison with `require.ErrorIs(t, err, ErrIllegalState)`

pkg/server/errors_test.go
```diff
@@ -23,20 +23,20 @@ import (
...
 
 	err = mapServerError(store.ErrIllegalArguments)
-	assert.Equal(t, ErrIllegalArguments, err)
+	require.Equal(t, ErrIllegalArguments, err)
```
jeroiraz, line 34: it may be better to improve error comparison with `require.ErrorIs(t, err, ErrIllegalArguments)`

pkg/server/state_signer_test.go
```diff
@@ -22,27 +22,27 @@ import (
...
+	require.NoError(t, err)
+	require.IsType(t, &schema.ImmutableState{}, state)
 }
 
 func TestStateSigner_Err(t *testing.T) {
 	s, _ := signer.NewSigner("./../../test/signer/ec3.key")
 	stSigner := NewStateSigner(s)
 	err := stSigner.Sign(nil)
-	assert.Error(t, store.ErrIllegalArguments, err)
+	require.Error(t, store.ErrIllegalArguments, err)
```
jeroiraz, line 47: it may be better to improve error comparison with `require.ErrorIs(t, err, store.ErrIllegalArguments)`

pkg/server/errors_test.go
Nowikens: I see that there's [this issue](https://github.com/codenotary/immudb/issues/1435) that relates to your remarks. I can take on this issue too, but should I make such changes on this branch?

jeroiraz: Sure, in another PR would be fine. Thanks a lot

Nowikens: And is this PR considered ok, or should I just wait for review? I don't really know what should happen next, so sorry if that's a dumb question.

PR #1467, opened 2022-12-19
mvcc

jeroiraz:
This PR implements multi-version concurrency control in  order to support simultaneous read-write transactions

Read conflicts may still be raised if data read by the transaction being committed has been invalidated by another recently committed one

PR #1468, opened 2022-12-21
MVCC read-set with boundaries

jeroiraz:
This PR sets a configurable maximum size for the read-set used to validate transaction conflicts

Note: This PR may be reviewed after PR #1467

### DIFF
--- a/embedded/store/immustore.go
+++ b/embedded/store/immustore.go
@@ -59,6 +59,7 @@ var ErrorMaxValueLenExceeded = errors.New("max value length exceeded")
 var ErrPreconditionFailed = errors.New("precondition failed")
 var ErrDuplicatedKey = errors.New("duplicated key")
 var ErrMaxActiveTransactionsLimitExceeded = errors.New("max active transactions limit exceeded")
+var ErrMVCCReadSetLimitExceeded = errors.New("MVCC read-set limit exceeded")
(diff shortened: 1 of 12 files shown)


embedded/store/immustore.go
```diff
@@ -158,6 +159,7 @@ type ImmuStore struct {
 	synced                bool
 	syncFrequency         time.Duration
 	maxActiveTransactions int
+	mvccReadSetLimit      int
```
tomekkolo, line 162: would be great to see comments explaining those fields :P

embedded/store/options.go
```diff
@@ -73,6 +74,7 @@ type Options struct {
 	CompactionDisabled bool
 
 	MaxActiveTransactions int
+	MVCCReadSetLimit      int
```
tomekkolo, line 77: please add comment what this limit means

embedded/store/ongoing_tx.go
```diff
@@ -273,7 +274,12 @@ func (tx *OngoingTx) GetWithFilters(key []byte, filters ...FilterFn) (ValueRef,
 			filters: filters,
 		}
 
+		if tx.readsetSize == tx.st.mvccReadSetLimit {
```
arriqaaq, line 277: you could use a method on the tx here, something like `isMVCCLimitReached` or something similar, as this check is repeated a lot

arriqaaq (dismissed):
lgtm, minor comment on refactor

PR #1470, opened 2022-12-23
Optimization: snapshot reuse

jeroiraz:
This PR is aimed to improve performance by reusing snapshots.

Transaction creation can be parametrized with conditions under which snapshots can be reused.

MVCC validations are in place, it means, if a read conflict is detected, it will be raised and the transaction may need to re-started. 

Note: This PR may be reviewed after PR #1468

PR #1471, opened 2022-12-23
read-only transactions

jeroiraz:
This PR introduces support for read-only transactions.

The benefit of this type of transactions is that MVCC validations are disabled for this type of transactions. It means, read-only transactions should be faster and lighter than read-write ones.

Note: This PR may be reviewed after PR #1470

### DIFF
--- a/embedded/store/ongoing_tx.go
+++ b/embedded/store/ongoing_tx.go
@@ -29,7 +29,7 @@ import (
 type OngoingTx struct {
 	st       *ImmuStore
 	snap     *Snapshot
-	readOnly bool
+	readOnly bool // MVCC validations are not needed for read-only transactions
 
 	entries      []*EntrySpec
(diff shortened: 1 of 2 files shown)


embedded/store/ongoing_tx.go
```diff
@@ -323,7 +339,7 @@ func (tx *OngoingTx) GetWithPrefixAndFilters(prefix, neq []byte, filters ...Filt
 	}
 
 	key, valRef, err = tx.snap.GetWithPrefixAndFilters(prefix, neq, filters...)
-	if errors.Is(err, ErrKeyNotFound) {
+	if !tx.readOnly && errors.Is(err, ErrKeyNotFound) {
```
arriqaaq, line 342: minor comment, but if we can combine this and line 360 under the same condition for tx.readOnly, would be very readable

arriqaaq (approved):
lgtm, minor comment

PR #1472, opened 2022-12-26
Remove useless code.

axaysagathiya:
Signed-off-by: Axay sagathiya <axaysagathiya@gmail.com>

- Returned error when verifiedGet operation fails.
- Removed unneeded fmt.Sprintf().

### DIFF
--- a/pkg/client/streams.go
+++ b/pkg/client/streams.go
@@ -301,7 +301,7 @@ func (c *immuClient) _streamVerifiedGet(ctx context.Context, req *schema.Verifia
 		return nil, err
 	}
 
-	gs, err := c.streamVerifiableGet(ctx, req)
+	gs, _ := c.streamVerifiableGet(ctx, req)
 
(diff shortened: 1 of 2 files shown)


pkg/client/streams.go
```diff
@@ -301,7 +301,7 @@ func (c *immuClient) _streamVerifiedGet(ctx context.Context, req *schema.Verifia
 		return nil, err
 	}
 
-	gs, err := c.streamVerifiableGet(ctx, req)
+	gs, _ := c.streamVerifiableGet(ctx, req)
```
jeroiraz, line 304: actually, the error should be returned instead

jeroiraz:
thanks @axaysagathiya

jeroiraz (approved):

jeroiraz:
@axaysagathiya could you ammend the commit message so to include the prefix `fix(pkg/client):`  before the description? e.g. fix(pkg/client): return error when verifiedGet operation fails

jeroiraz (approved):

PR #1474, opened 2022-12-28
tbtree optimizations

jeroiraz:
This PR introduces optimizations in bulk insertions by following an optimistic concurrent approach.
KV pairs to be inserted are first distributed among corresponding nodes and insertion takes place at the same time, finally the new list of nodes is built up.

### DIFF
--- a/embedded/tbtree/tbtree.go
+++ b/embedded/tbtree/tbtree.go
@@ -27,7 +27,6 @@ import (
 	"math"
 	"os"
 	"path/filepath"
-	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -1562,7 +1561,8 @@ func (t *TBtree) BulkInsert(kvs []*KV) error {
 		return ErrIllegalArguments
 	}
 
-	sortedKVs := make([]*KV, len(kvs))
+	// validated immutable copy of input kv pairs
+	immutableKVs := make([]*KV, len(kvs))
 
 	for i, kv := range kvs {
 		if kv == nil || kv.K == nil || kv.V == nil {
@@ -1583,16 +1583,11 @@ func (t *TBtree) BulkInsert(kvs []*KV) error {
 		v := make([]byte, len(kv.V))
 		copy(v, kv.V)
 
-		sortedKVs[i] = &KV{
+		immutableKVs[i] = &KV{
 			K: k,
 			V: v,
 		}
 	}
-
-	// sort entries to increase cache hits
-	sort.Slice(sortedKVs, func(i, j int) bool {
-		return bytes.Compare(sortedKVs[i].K, sortedKVs[j].K) < 0
-	})
 
 	t.rwmutex.Lock()
 
@@ -1616,7 +1611,7 @@ func (t *TBtree) BulkInsert(kvs []*KV) error {
 
 	ts := t.root.ts() + 1
 
-	nodes, depth, err := t.root.insertAt(sortedKVs, ts)
+	nodes, depth, err := t.root.insertAt(immutableKVs, ts)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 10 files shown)


embedded/tbtree/tbtree.go
```diff
@@ -1758,63 +1758,102 @@ func (t *TBtree) snapshotClosed(snapshot *Snapshot) error {
...
+		}
 	}
 
-	copy(newNode.nodes, n.nodes[:insertAt])
-	copy(newNode.nodes[insertAt:], cs)
-	copy(newNode.nodes[insertAt+len(cs):], n.nodes[insertAt+1:])
+	n.nodes = ns
 
-	nodes, err = newNode.split()
+	nodes, _ = n.split()
```
arriqaaq, line 1854: are we ignoring error here intentionally?

embedded/tbtree/tbtree.go
```diff
@@ -641,7 +640,7 @@ func OpenWith(path string, nLog, hLog, cLog appendable.Appendable, opts *Options
 	}
 
 	if validatedCLogEntry == nil {
-		t.root = &leafNode{t: t, mut: true}
+		t.root = &leafNode{t: t, mut: false}
```
arriqaaq, line 643: why is this changed to false?

embedded/tbtree/tbtree.go
jeroiraz: inner nodes does not return error during split, but I'll add an special error handling

embedded/tbtree/tbtree.go
jeroiraz: there is just a small difference, if the node start mutated it means changes will be done in the same instance, otherwise a copy will be made. It's better to always do the changes in a copy so any rollback can be done, even when starting with a fresh btree

embedded/tbtree/tbtree.go
```diff
@@ -1760,63 +1766,105 @@ func (t *TBtree) snapshotClosed(snapshot *Snapshot) error {
...
+		childIndex := n.indexOf(kv.K)
+		kvsPerChild[childIndex] = append(kvsPerChild[childIndex], kv)
 	}
 
-	n._ts = ts
+	var wg sync.WaitGroup
+	wg.Add(len(kvsPerChild))
 
-	ns := make([]node, len(n.nodes)+len(cs)-1)
+	nodesPerChild := make(map[int][]node)
```
arriqaaq, line 1799: it would be nice if you make this in a struct too, with lock/unlock methods which will make it easier to read/test

embedded/tbtree/tbtree.go
```diff
@@ -1760,63 +1766,105 @@ func (t *TBtree) snapshotClosed(snapshot *Snapshot) error {
...
 
-	n.nodes = ns
+			c := n.nodes[childIndex]
 
-	nodes, err = n.split()
+			cs, cdepth, cerr := c.insertAt(childKVs, ts)
 
-	return nodes, depth + 1, err
-}
+			nodesMutex.Lock()
```
arriqaaq, line 1811: shouldn't the lock be acquired after the error?

embedded/tbtree/tbtree.go
```diff
@@ -1760,63 +1766,105 @@ func (t *TBtree) snapshotClosed(snapshot *Snapshot) error {
...
+			defer nodesMutex.Unlock()
 
-func (n *innerNode) copyOnInsertAt(key []byte, value []byte, ts uint64) (nodes []node, depth int, err error) {
-	insertAt := n.indexOf(key)
+			if err != nil {
```
arriqaaq, line 1814: shouldn't the check be on cerr? Or am I missing something?

jeroiraz: yep

arriqaaq (approved):

PR #1476, opened 2023-01-02
tbtree: multi-timed bulk insertions

jeroiraz:
This PR introduces support for multi-timed bulk insertions. Each entry may have associated its own logical timestamp.

Previously, all the entries were associated to the same timestamp i.e. root's current timestamp increased by one.

This functionality will be used to group multiple transactions for bulk insertion in the btree. However those changes will be implemented as part of a different PR.

### DIFF
--- a/embedded/tbtree/tbtree.go
+++ b/embedded/tbtree/tbtree.go
@@ -1576,6 +1576,11 @@ func (t *TBtree) Insert(key []byte, value []byte) error {
 	return t.bulkInsert([]*KVT{{K: key, V: value}})
 }
 
+// BulkInsert inserts multiple entries atomically.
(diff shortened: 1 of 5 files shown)


embedded/tbtree/tbtree.go
```diff
@@ -1549,71 +1549,100 @@ func (t *TBtree) IncreaseTs(ts uint64) error {
...
+		immutableKVTs[i] = &KVT{
+			K: k,
+			V: v,
+			T: t,
 		}
-	}()
 
-	if t.closed {
-		return ErrAlreadyClosed
+		if t > newTs {
```
arriqaaq, line 1640: Can we just add test cases for this to understand the flow better for ts?

arriqaaq (approved):

PR #1478, opened 2023-01-04
grouped indexing

jeroiraz:


embedded/store/indexer.go
```diff
@@ -77,30 +85,71 @@ var (
...
+		WithMaxNodeSize(opts.IndexOpts.MaxNodeSize).
+		WithMaxKeySize(opts.MaxKeyLen).
+		WithMaxValueSize(lszSize + offsetSize + sha256.Size + sszSize + maxTxMetadataLen + sszSize + maxKVMetadataLen). // indexed values
+		WithNodesLogMaxOpenedFiles(opts.IndexOpts.NodesLogMaxOpenedFiles).
+		WithHistoryLogMaxOpenedFiles(opts.IndexOpts.HistoryLogMaxOpenedFiles).
+		WithCommitLogMaxOpenedFiles(opts.IndexOpts.CommitLogMaxOpenedFiles).
+		WithRenewSnapRootAfter(opts.IndexOpts.RenewSnapRootAfter).
+		WithCompactionThld(opts.IndexOpts.CompactionThld).
+		WithDelayDuringCompaction(opts.IndexOpts.DelayDuringCompaction)
+
```
arriqaaq, line 113: I think the part to validate indexopts is missing? 
This line I mean err = indexOpts.Validate()

jeroiraz: actually, the opts are validated by the btree itself

embedded/store/indexer.go
```diff
@@ -371,78 +420,100 @@ func (idx *indexer) doIndexing() {
...
+			idx._kvs[indexableEntries].V = b[:o]
+			idx._kvs[indexableEntries].T = txID + uint64(i)
 
-		indexableEntries++
+			indexableEntries++
+		}
+
+		bulkSize++
+
+		if bulkSize < idx.maxBulkSize {
```
arriqaaq, line 493: Wouldn't this condition always be true?

embedded/store/indexer.go
```diff
@@ -371,78 +420,100 @@ func (idx *indexer) doIndexing() {
...
+			break
+		}
+		if err != nil {
+			return err
+		}
 	}
 
+	var err error
+
 	if indexableEntries == 0 {
```
arriqaaq, line 507: Can you just add a comment on why ts is being increased when indexableEntries ==0? :)

arriqaaq (commented):
minor comments, lgtm

embedded/store/indexer.go
jeroiraz: it shouldn't hold in the last iteration

embedded/store/indexer.go
jeroiraz: when a transaction didn't have any indexable entry, the logical time in the tree was increased. But that is actually not required anymore, given the transaction id is specified per entry now

arriqaaq (approved):

PR #1486, opened 2023-01-11
Feat/truncate vlog

arriqaaq:


embedded/sql/engine.go
```diff
@@ -493,3 +495,265 @@ func normalizeParams(params map[string]interface{}) (map[string]interface{}, err
 
 	return nparams, nil
 }
+
+// CopyCatalog returns a new transaction with a copy of the current catalog.
+func (e *Engine) CopyCatalog(ctx context.Context) (*store.OngoingTx, error) {
+	e.mutex.RLock()
+	defer e.mutex.RUnlock()
+
+	sqltx, err := e.NewTx(context.Background(), DefaultTxOptions())
```
jeroiraz, line 504:
```suggestion
	sqltx, err := e.NewTx(ctx, DefaultTxOptions())
```

embedded/sql/engine.go
```diff
@@ -493,3 +495,265 @@ func normalizeParams(params map[string]interface{}) (map[string]interface{}, err
 
 	return nparams, nil
 }
+
+// CopyCatalog returns a new transaction with a copy of the current catalog.
+func (e *Engine) CopyCatalog(ctx context.Context) (*store.OngoingTx, error) {
```
jeroiraz, line 500: already discussed about it, it seems rare to return a new transaction as part of `CopyCatalog`, it may be better to receive it as input

pkg/database/database.go
```diff
@@ -1617,6 +1639,13 @@ func (d *db) Close() (err error) {
 
 	d.sqlInit.Wait() // Wait for SQL Engine initialization to conclude
 
+	if d.isDataRetentionEnabled() {
+		d.quitOnce.Do(func() {
```
jeroiraz, line 1643: quitOnce might not be needed

pkg/database/truncator.go
```diff
@@ -0,0 +1,159 @@
...
+type Truncator interface {
+	// Plan returns the transaction upto which the log can be truncated.
+	// When resulting transaction before specified time does not exists
+	//  * No transaction header is returned.
+	//  * Returns nil TxHeader, and an error.
+	Plan(time.Time) (*store.TxHeader, error)
+
+	// Truncate runs truncation against the relevant appendable logs. Must
+	// be called after result of Plan().
+	Truncate(*store.TxHeader) error
```
jeroiraz, line 45: minor: receiving a TxHeader as parameter might not be required and just a txID would be enough e.g. if an API to truncate up to a specific transaction is provided, there won't be a need of reading the header

pkg/database/truncator.go
```diff
@@ -0,0 +1,159 @@
...
+// Truncator provides truncation against an underlying storage
+// of appendable data.
+type Truncator interface {
+	// Plan returns the transaction upto which the log can be truncated.
+	// When resulting transaction before specified time does not exists
+	//  * No transaction header is returned.
+	//  * Returns nil TxHeader, and an error.
+	Plan(time.Time) (*store.TxHeader, error)
+
+	// Truncate runs truncation against the relevant appendable logs. Must
```
jeroiraz, line 43: there shouldn't be a strong dependency on `Plan`, calling truncation up to a transaction should be possible without calling `Plan`

pkg/database/truncator.go
```diff
@@ -0,0 +1,159 @@
...
+	metrics *truncatorMetrics
+}
+
+// Plan returns the transaction upto which the value log can be truncated.
+// When resulting transaction before specified time does not exists
+//  * No transaction header is returned.
+//  * Returns nil TxHeader, and an error.
+// The retentionPeriod time is truncated to the day.
+func (v *vlogTruncator) Plan(retentionPeriod time.Time) (*store.TxHeader, error) {
+	hdr, err := v.db.st.FirstTxSince(retentionPeriod)
```
jeroiraz, line 67: `retentionPeriod` doesn't seem to be a retention period as it's used in this line

pkg/database/truncator.go
```diff
@@ -0,0 +1,159 @@
...
+type vlogTruncator struct {
+	db      *db
+	metrics *truncatorMetrics
+}
+
+// Plan returns the transaction upto which the value log can be truncated.
+// When resulting transaction before specified time does not exists
+//  * No transaction header is returned.
+//  * Returns nil TxHeader, and an error.
+// The retentionPeriod time is truncated to the day.
```
jeroiraz, line 65: wondering why retentionPeriod needs to be truncated

pkg/database/database.go
arriqaaq: This is required in case close is called twice, was failing in one test case

pkg/database/database.go
```diff
@@ -1686,3 +1715,93 @@ func logErr(log logger.Logger, formattedMessage string, err error) error {
...
+//		tn-1:vx  tn:vx   tn+1:vx
+//
+func (d *db) truncate(ts time.Time) error {
+	for _, c := range d.truncators {
+		// Plan determines the transaction header before time period ts. If a
+		// transaction is not found, or if an error occurs fetching the transaction,
+		// then truncation does not run for the specified appendable.
+		hdr, err := c.Plan(ts)
+		if err != nil {
+			switch err {
```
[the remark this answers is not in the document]
arriqaaq: I've been able to catch specific errors, as these are returned without being wrapped, the rest are not caught as such.

pkg/database/truncator.go
arriqaaq: In terms of the truncator, plan is required to figure out the transaction header up to which truncation can be done, so it seems apt for this interface

pkg/database/truncator.go
arriqaaq: This is just to keep uniformity on when truncation is done for the entire day, because if we say delete data for n days, time in hours/seconds shouldn't matter else half a day data could be truncated on that day depending on the timestamps

pkg/database/database.go
jeroiraz: aham, probably it's better to have a flag in the database level to indicate the database is already closed. So to avoid any action when calling close more than once

pkg/database/database.go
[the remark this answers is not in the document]
jeroiraz: understand but we shouldn't rely on the errors not being wrapped, as that may change at any moment e.g. attaching contextual info to the error

embedded/sql/engine.go
arriqaaq: resolved

pkg/database/database.go
arriqaaq: resolved

pkg/database/database.go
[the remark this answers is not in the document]
arriqaaq: resolved

embedded/sql/engine.go
arriqaaq: resolved

pkg/database/truncator.go
arriqaaq: resolved

embedded/store/immustore.go
```diff
@@ -2737,3 +2828,163 @@ func minUint64(a, b uint64) uint64 {
 	}
 	return a
 }
+
+// readTxOffsetAt reads the value-log offset of a specific entry (index) in a transaction (txID)
+func (s *ImmuStore) readTxOffsetAt(txID uint64, allowPrecommitted bool, index int) (*TxEntry, error) {
```
jeroiraz, line 2833: method name and index parameter are not self-explanatory

embedded/store/immustore.go
```diff
@@ -2737,3 +2828,163 @@ func minUint64(a, b uint64) uint64 {
...
+		return s.readTxOffsetAt(id, false, 1)
+	}
+
+	back := func(id uint64) error {
+		first, err := readFirstEntryOffset(id)
+		if err != nil {
+			return err
+		}
+
+		// Iterate over all past transactions and store the minimum offset for each value log file.
```
jeroiraz, line 2915: comment here mention iteration but this method is not doing an iteration

embedded/store/immustore.go
```diff
@@ -2737,3 +2828,163 @@ func minUint64(a, b uint64) uint64 {
...
+		v, off := decodeOffset(first.VOff())
```
jeroiraz, line 2916: variable names such as first and v may become more clear if they include further detail e.g. firstTxEntry and vLogID

embedded/store/immustore.go
```diff
@@ -2737,3 +2828,163 @@ func minUint64(a, b uint64) uint64 {
...
+
+	var err error
+	// tombstones maintain the minimum offset for each value log file that can be safely deleted.
+	tombstones := make(map[byte]int64)
+
+	readFirstEntryOffset := func(id uint64) (*TxEntry, error) {
+		return s.readTxOffsetAt(id, false, 1)
+	}
+
+	back := func(id uint64) error {
```
jeroiraz, line 2909: id correspond to txID, if that's the case, it's better to use that naming as it's used in the rest of the code and it's faster to read

embedded/store/immustore.go
```diff
@@ -2737,3 +2828,163 @@ func minUint64(a, b uint64) uint64 {
...
+		first, err := readFirstEntryOffset(id)
+		if err != nil {
+			return err
+		}
+
+		// Iterate over all past transactions and store the minimum offset for each value log file.
+		v, off := decodeOffset(first.VOff())
+		if _, ok := tombstones[v]; !ok {
```
jeroiraz, line 2917: in which cases this would happen? another transaction in the same vLog? if that's the case, the one with lower offset should be the one tracked right? is this relying on calling the method in an specific order i.e. with ascending tx ids?

embedded/store/immustore.go
```diff
@@ -2737,3 +2828,163 @@ func minUint64(a, b uint64) uint64 {
...
+			tombstones[v] = off
+		}
+		return nil
+	}
+
+	front := func(id uint64) error {
+		first, err := readFirstEntryOffset(id)
```
jeroiraz, line 2924: similar naming comments as above

embedded/store/immustore.go
```diff
@@ -2737,3 +2828,163 @@ func minUint64(a, b uint64) uint64 {
...
+		if err != nil {
+			return err
+		}
+
+		// Check if any future transaction offset lies before past transaction(s)
```
jeroiraz, line 2929: comment here does not reflect what the code is doing within the scope of this method, probably it's how it's being used?

embedded/store/immustore.go
```diff
@@ -2737,3 +2828,163 @@ func minUint64(a, b uint64) uint64 {
...
+		return nil
+	}
+
+	// Walk back to get offsets across vlogs which can be deleted safely.
+	// This way, we can calculate the minimum offset for each value log file.
+	{
+		var i uint64 = minTxID
+		for i > 0 && len(tombstones) != s.MaxIOConcurrency() {
+			err = back(i)
+			i--
```
jeroiraz, line 2946: shouldn't i be decremented if there is no error?

embedded/store/immustore.go
```diff
@@ -2737,3 +2828,163 @@ func minUint64(a, b uint64) uint64 {
...
+			if err != nil { // if txn not found, then continue as previous txn could have been deleted
```
jeroiraz, line 2947: I think if tx is not found it won't continue

embedded/store/immustore.go
```diff
@@ -2737,3 +2828,163 @@ func minUint64(a, b uint64) uint64 {
...
+	{
+		// Check for transactions upto the last committed transaction to avoid deletion of values for any future transaction.
+		maxTxID := s.LastCommittedTxID()
+		s.logger.Infof("running truncation check between transaction '%d' and '%d'", minTxID, maxTxID)
+
+		// TODO: add more integration tests
+		// Iterate over all future transactions to check if any offset lies before past transaction(s) offset.
+		for j := minTxID; j <= maxTxID; j++ {
+			err = front(j)
+			if err != nil {
```
jeroiraz, line 2964: it's not clear to me why `front` will return an error according to the description in the comment

pkg/api/schema/schema.proto
```diff
@@ -1439,6 +1454,19 @@ message RetryInfo {
   int32 retry_delay = 1;
 }
 
+message TruncateDatabaseRequest {
+  // Database name
+  string database = 1;
+
+  // Retention Period of data
+  NullableMilliseconds retentionPeriod = 2;
```
jeroiraz, line 1462: if the truncator process is not enabled and retentionPeriod is not specified in the request, which value is used? or it returns an error?

pkg/api/schema/schema.proto
```diff
@@ -1439,6 +1454,19 @@ message RetryInfo {
...
+}
+
+message TruncateDatabaseResponse {
+  // Database name
+  string database = 1;
```
jeroiraz, line 1467: ideally we should return the older txID still present in the database after truncation

pkg/database/dboptions.go
```diff
@@ -36,14 +41,21 @@ type Options struct {
...
 // DefaultOption Initialise Db Optionts to default values
 func DefaultOption() *Options {
 	return &Options{
-		dbRootPath:     DefaultDbRootPath,
-		storeOpts:      store.DefaultOptions(),
-		readTxPoolSize: DefaultReadTxPoolSize,
+		dbRootPath:          DefaultDbRootPath,
+		storeOpts:           store.DefaultOptions(),
+		readTxPoolSize:      DefaultReadTxPoolSize,
+		TruncationFrequency: DefaultTruncationFrequency,
```
jeroiraz, line 58: others attributes are private, does it require to be public?

pkg/database/truncator.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// setting the metadata to record the transaction upto which the log was truncated
+	tx.WithMetadata(store.NewTxMetadata().WithTruncatedTxID(hdr.ID))
+
+	// commit catalogue as a new transaction
+	return tx.Commit(context.Background())
+}
+
+// Truncate runs truncation against the relevant appendable logs upto the specified transaction offset.
+func (v *vlogTruncator) Truncate(hdr *store.TxHeader) error {
```
jeroiraz, line 116: minor, TruncateUptoTx naming as in the embedded store is more descriptive

jeroiraz: header struct doesn't seem to be needed in this method but just the txID

pkg/server/server.go
```diff
@@ -1586,3 +1627,66 @@ func (s *ImmuServer) mandatoryAuth() bool {
...
+		return nil, fmt.Errorf("could not get loggedin user data")
+	}
+
+	//if the requesting user has admin permission on this database
+	if (!user.IsSysAdmin) &&
+		(!user.HasPermission(req.Database, auth.PermissionAdmin)) {
+		return nil, fmt.Errorf("the database '%s' does not exist or you do not have admin permission on this database", req.Database)
+	}
+
+	if req.RetentionPeriod == nil {
```
jeroiraz, line 1665: if it's required, then in proto definition is may not be a nullable but a plain value

pkg/server/server.go
```diff
@@ -1586,3 +1627,66 @@ func (s *ImmuServer) mandatoryAuth() bool {
...
+	defer s.dbListMutex.Unlock()
+
+	db, err := s.dbList.GetByName(req.Database)
+	if err != nil {
+		return nil, err
+	}
+
+	rp := time.Duration(req.RetentionPeriod.Value) * time.Millisecond
+	truncator := truncator.NewTruncator(db, s.Logger)
+	err = truncator.Truncate(rp)
```
jeroiraz, line 1684: is this safe even when truncator process is running, right?

pkg/server/server_test.go
```diff
@@ -2032,3 +2032,88 @@ func TestServerMaintenanceMode(t *testing.T) {
...
+
+	s.Initialize()
+
+	r := &schema.LoginRequest{
+		User:     []byte(auth.SysAdminUsername),
+		Password: []byte(auth.SysAdminPassword),
+	}
+
+	ctx := context.Background()
+	lr, err := s.Login(ctx, r)
```
jeroiraz, line 2051: new code should rely on session-based authentication as login/logout is deprecated and we should gradually remove all the code using it

pkg/server/truncator.go
```diff
@@ -0,0 +1,85 @@
...
+
+package server
+
+import (
+	"github.com/codenotary/immudb/pkg/database"
+	"github.com/codenotary/immudb/pkg/truncator"
+)
+
+func (s *ImmuServer) truncationInProgressFor(db string) bool {
+	_, ok := s.truncators[db]
```
jeroiraz, line 25: just to check: it doesn't need to be protected from concurrency, right?

pkg/server/truncator.go
```diff
@@ -0,0 +1,85 @@
...
+	s.truncatorMutex.Lock()
+	defer s.truncatorMutex.Unlock()
+
+	t, ok := s.truncators[db]
+	if !ok {
+		return ErrTruncatorNotInProgress
+	}
+
+	err := t.Stop()
+	if err == truncator.ErrTruncatorAlreadyStopped {
```
jeroiraz, line 63: stopTruncatorFor may remove the truncator from the map but stopTruncation doesn't, I'm wondering why such difference

embedded/store/immustore.go
arriqaaq: actually the error is being returned, so should not matter, but yeah can be moved post error check

embedded/store/immustore.go
arriqaaq: If there is any error in reading the transaction, there is some corruption in data etc, I am returning the same error when the value is read from the log

arriqaaq: And the error is returned to avoid deleting any record accidentally if that transaction can't be read. Could be signs of a corrupted data record

pkg/api/schema/schema.proto
arriqaaq: This is used in the Truncate command in the immuadmin. If retention period is not specified, an error is returned

pkg/api/schema/schema.proto
arriqaaq: I can try adding that later, will keep this as lower priority

pkg/database/truncator.go
arriqaaq: fixed

embedded/store/immustore.go
arriqaaq: fixed

embedded/store/immustore.go
arriqaaq: fixed

embedded/store/immustore.go
arriqaaq: fixed

embedded/store/immustore.go
arriqaaq: yes, fixed the comment

pkg/database/dboptions.go
arriqaaq: yes, because this is being used by the truncator to know the trunc frequency and retention period

pkg/database/truncator.go
arriqaaq: fixed

pkg/server/server.go
arriqaaq: yes

pkg/server/truncator.go
arriqaaq: yes, it should basically delete it

embedded/store/immustore.go
jeroiraz: ah ok, so it's a validation that all transactions greater than minTxID won't be truncated

pkg/server/server_test.go
arriqaaq: fixed

embedded/store/ongoing_tx.go
```diff
@@ -69,6 +69,13 @@ type EntrySpec struct {
 	Key      []byte
 	Metadata *KVMetadata
 	Value    []byte
+	// hashValue is the hash of the value
+	// if the actual value is truncated. This is
+	// used during replication.
+	hashValue []byte
+	// isValueTruncated is true if the value is
+	// truncated. This is used during replication.
+	isValueTruncated bool
```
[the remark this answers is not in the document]
arriqaaq: could you please give an example? I was checking the code, and was not able to understand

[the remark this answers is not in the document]
arriqaaq: fixed

pkg/server/server.go
arriqaaq: fixed

pkg/database/truncator_test.go
```diff
@@ -92,7 +92,7 @@ func Test_vlogCompactor_WithMultipleIO(t *testing.T) {
 	hdr, err := db.st.ReadTxHeader(deletePointTx, false)
 	require.NoError(t, err)
 	c := NewVlogTruncator(db)
-	require.NoError(t, c.Truncate(hdr.ID))
+	require.NoError(t, c.Truncate(context.TODO(), hdr.ID))
```
jeroiraz, line 95: `context.Background()` may be used instead of `context.TODO()`, here and in other places

jeroiraz (approved):
while there are still things to evaluate e.g. tx serialization and test coverage, I think we can proceed merging the PR into master and continue from it

Note: made a minor comment regarding context in the test cases

embedded/store/immustore.go
```diff
@@ -2566,35 +2566,31 @@ func (s *ImmuStore) ReadValue(entry *TxEntry) ([]byte, error) {
...
 func (s *ImmuStore) readValueAt(b []byte, off int64, hvalue [sha256.Size]byte) (n int, err error) {
-	mustReadFromVLog := true
-
 	if s.vLogCache != nil {
 		val, err := s.vLogCache.Get(off)
 		if err == nil {
 			// the requested value was found in the value cache
 			copy(b, val.([]byte))
-			mustReadFromVLog = false
+			return len(b), nil
```
jeroiraz, line 2574: the flag was intentional so to do the digest comparison, otherwise the validation against the expected hvalue won't be done

embedded/store/immustore.go
```diff
@@ -2566,35 +2566,31 @@ func (s *ImmuStore) ReadValue(entry *TxEntry) ([]byte, error) {
...
 	}
 
-	if mustReadFromVLog {
-		vLogID, offset := decodeOffset(off)
+	vLogID, offset := decodeOffset(off)
 
-		if vLogID > 0 {
-			vLog := s.fetchVLog(vLogID)
-			defer s.releaseVLog(vLogID)
+	if vLogID > 0 {
```
jeroiraz, line 2584: I think we should return error is `vLogID === 0`

embedded/store/immustore.go
arriqaaq: can we do the validation only when len(b)>0?

embedded/store/immustore_test.go
```diff
@@ -4253,7 +4253,8 @@ func TestImmudbStoreTruncateUptoTx_WithMultipleIOConcurrency(t *testing.T) {
 	opts := DefaultOptions().
 		WithFileSize(6).
 		WithMaxConcurrency(100).
-		WithMaxIOConcurrency(5)
+		WithMaxIOConcurrency(5).
+		WithVLogCacheSize(0)
```
jeroiraz, line 4257: better to disable the vlogCache by default, only enable at specific test cases

embedded/store/immustore.go
arriqaaq: makes sense, what error should it be?

embedded/store/immustore.go
jeroiraz: validation should be done in all the cases

embedded/store/immustore.go
jeroiraz: I think we could return the existing error `ErrOffsetOutOfRange`

embedded/store/immustore.go
arriqaaq: added validation for cached value

embedded/store/immustore.go
arriqaaq: Closing this in favor of issue created for handling this condition

embedded/store/immustore_test.go
arriqaaq: done

embedded/store/options.go
```diff
@@ -207,7 +207,6 @@ func DefaultOptions() *Options {
 		MaxIOConcurrency: DefaultMaxIOConcurrency,
 
 		TxLogCacheSize: DefaultTxLogCacheSize,
-		VLogCacheSize:  DefaultVLogCacheSize,
```
jeroiraz, line 207: explicit initialization should be keep `VLogCacheSize:  DefaultVLogCacheSize,`

jeroiraz (approved):

PR #1487, opened 2023-01-12
context propagation

jeroiraz:


embedded/sql/row_reader.go
```diff
@@ -380,7 +381,11 @@ func (r *rawRowReader) reduceTxRange() (err error) {
 	return nil
 }
 
-func (r *rawRowReader) Read() (row *Row, err error) {
+func (r *rawRowReader) Read(ctx context.Context) (row *Row, err error) {
+	if ctx.Err() != nil {
```
arriqaaq, line 385: should this not be in a defer? Context could be cancelled during the whole read process

arriqaaq: Actually both checks are required, at the start and before returning

embedded/sql/row_reader.go
```diff
@@ -297,24 +298,24 @@ func (r *rawRowReader) ScanSpecs() *ScanSpecs {
 	return r.scanSpecs
 }
 
-func (r *rawRowReader) Columns() ([]ColDescriptor, error) {
+func (r *rawRowReader) Columns(ctx context.Context) ([]ColDescriptor, error) {
```
arriqaaq, line 301: is context required here if it is not being used or propagated?

embedded/sql/row_reader.go
```diff
@@ -297,24 +298,24 @@ func (r *rawRowReader) ScanSpecs() *ScanSpecs {
...
 	ret := make([]ColDescriptor, len(r.colsByPos))
 	for i := range r.colsByPos {
 		ret[i] = r.colsByPos[i]
 	}
 	return ret, nil
 }
 
-func (r *rawRowReader) colsBySelector() (map[string]ColDescriptor, error) {
+func (r *rawRowReader) colsBySelector(ctx context.Context) (map[string]ColDescriptor, error) {
```
arriqaaq, line 309: same question as above

embedded/sql/row_reader.go
arriqaaq: discussed with you, closing this as the context is part of the interface hence included

embedded/sql/row_reader.go
arriqaaq: discussed with you, closing this as the context is part of the interface hence included

embedded/sql/row_reader.go
jeroiraz: has discussed offline but leaving a comment for future reference, context might be closed at any point in time and it shouldn't make a difference in the result but it may avoid unnecessary delays whenever possible

embedded/sql/row_reader.go
jeroiraz: this method is declared in an interface, some concrete implementations are using it. Ideally I'd remove the context on this particular method but it may involve some further changes we may do in the future

embedded/sql/row_reader.go
jeroiraz: replied above

arriqaaq (commented):
LGTM, just one comment, we might want to add tests where context has a timeout to check if the functionality works as expected, wdyt?

arriqaaq (approved):

PR #1490, opened 2023-01-16
Feat/add truncate command

arriqaaq:


jeroiraz:
> [Coverage Status](https://coveralls.io/builds/55961338)
> 
> Coverage: 88.918% (-0.2%) from 89.135% when pulling **[a65e000](https://github.com/codenotary/immudb/commit/a65e000c482796e24c7c9e72d3874d0be63fc4bf) on feat/add-truncate-command** into **[3e91ef1](https://github.com/codenotary/immudb/commit/3e91ef115f09d93dc53084fb8d6434035dc5d200) on feat/truncate-vlog**.

@arriqaaq,  please review why coverage is decreasing that much

cmd/immuadmin/command/database.go
```diff
@@ -317,6 +317,45 @@ func (cl *commandline) database(cmd *cobra.Command) {
 		Args: cobra.ExactArgs(0),
 	}
 
+	truncateCmd := &cobra.Command{
+		Use:               "truncate",
+		Short:             "Truncate database (unrecoverable operation)",
+		Example:           "delete --yes-i-know-what-i-am-doing {database_name}",
```
jeroiraz, line 323:
```suggestion
		Example:           "truncate --yes-i-know-what-i-am-doing {database_name}",
```

jeroiraz: probably the flag `retention-period` should be mandatory

pkg/client/client.go
```diff
@@ -505,6 +505,12 @@ type ImmuClient interface {
 	//
 	// Note: Currently such transaction can only be used for SQL operations.
 	NewTx(ctx context.Context) (Tx, error)
+
+	// TruncateDatabase truncates a database.
+	// This truncates the locally stored value log files used by the database.
+	//
+	// This call requires SysAdmin permission level or admin permission to the database.
+	TruncateDatabase(ctx context.Context, r *schema.TruncateDatabaseRequest) (*schema.TruncateDatabaseResponse, error)
```
jeroiraz, line 513: we should avoid schema structs as much as possible from the client APIs, in this case it would be enough with the context and the retention period as parameters and an error as response, unless more info is returned e.g. up to which transaction ID truncation was done

pkg/server/server.go
```diff
@@ -1586,3 +1586,65 @@ func (s *ImmuServer) mandatoryAuth() bool {
...
+		(!user.HasPermission(req.Database, auth.PermissionAdmin)) {
+		return nil, fmt.Errorf("the database '%s' does not exist or you do not have admin permission on this database", req.Database)
+	}
+
+	if req.RetentionPeriod == nil {
+		return nil, fmt.Errorf("retention period can not be empty for database '%s'", req.Database)
+	}
+	if req.RetentionPeriod.Value < 0 || (req.RetentionPeriod.Value > 0 && req.RetentionPeriod.Value < (store.MinimumRetentionPeriod.Milliseconds())) {
+		return nil, fmt.Errorf(
+			"%w: invalid retention period for database '%s'. RetentionPeriod should at least 1 day",
```
jeroiraz, line 1629: log message should not contain hardcoded values i.e. `RetentionPeriod should at least 1 day`

pkg/database/database.go
```diff
@@ -137,6 +137,8 @@ type DB interface {
 
 	IsClosed() bool
 	Close() error
+
+	Truncate(rp time.Duration) error
```
jeroiraz, line 141: returning the transaction id up to which it was truncated may be of interest

cmd/immuadmin/command/database.go
arriqaaq: fixed it

pkg/server/server.go
arriqaaq: fixed

pkg/client/client.go
arriqaaq: noted, removed it

pkg/database/database.go
arriqaaq: I'll address this in a different PR as this requires changes on the pkg/database level

PR #1491, opened 2023-01-16
chore(pkg/api/schema): add txmetadata conversion

arriqaaq:


pkg/server/db_options.go
```diff
@@ -711,6 +728,18 @@ func (opts *dbOptions) Validate() error {
 		)
 	}
 
+	if opts.RetentionPeriod < 0 || (opts.RetentionPeriod > 0 && opts.RetentionPeriod < Milliseconds(store.MinimumRetentionPeriod.Milliseconds())) {
+		return fmt.Errorf(
+			"%w: invalid retention period for database '%s'. RetentionPeriod should at least 1 day",
+			ErrIllegalArguments, opts.Database)
+	}
+
+	if opts.TruncationFrequency < 0 || (opts.TruncationFrequency > 0 && opts.TruncationFrequency < Milliseconds(store.MinimumTruncationFrequency.Milliseconds())) {
```
jeroiraz, line 737: It looks like the previous validation was duplicated

pkg/server/db_options.go
```diff
@@ -711,6 +728,18 @@ func (opts *dbOptions) Validate() error {
 		)
 	}
 
+	if opts.RetentionPeriod < 0 || (opts.RetentionPeriod > 0 && opts.RetentionPeriod < Milliseconds(store.MinimumRetentionPeriod.Milliseconds())) {
+		return fmt.Errorf(
+			"%w: invalid retention period for database '%s'. RetentionPeriod should at least 1 day",
```
jeroiraz, line 733: it would be better use the constant value denoting the minimum value in the error message

pkg/server/db_options.go
```diff
@@ -711,6 +728,18 @@ func (opts *dbOptions) Validate() error {
...
+			ErrIllegalArguments, opts.Database)
+	}
+
+	if opts.TruncationFrequency < 0 || (opts.TruncationFrequency > 0 && opts.TruncationFrequency < Milliseconds(store.MinimumTruncationFrequency.Milliseconds())) {
+		return fmt.Errorf(
+			"%w: invalid truncation frequency for database '%s'. TruncationFrequency should at least 1 hour",
```
jeroiraz, line 739: it would be better use the constant value denoting the minimum value in the error message

cmd/immuadmin/command/database.go
```diff
@@ -52,6 +52,8 @@ func addDbUpdateFlags(c *cobra.Command) {
 	c.Flags().Uint32("write-buffer-size", store.DefaultWriteBufferSize, "set the size of in-memory buffers for file abstractions")
 	c.Flags().Uint32("read-tx-pool-size", database.DefaultReadTxPoolSize, "set transaction read pool size (used for reading transaction objects)")
 	c.Flags().Bool("autoload", true, "enable database autoloading")
+	c.Flags().Duration("retention-period", 0, "duration of time to retain data in storage")
```
jeroiraz, line 55: a default constant might also be used when declaring `retention-period` flag?

pkg/server/db_options.go
arriqaaq: ack, noted

pkg/server/db_options.go
arriqaaq: noted

cmd/immuadmin/command/database.go
arriqaaq: actually I am using the condition, > 0 to determine if to start the truncation process. Because truncation by default is not enabled. DO you think we need to add another flag, like `--enable-truncation` and then take in these params?

jeroiraz: no, I think it's ok

pkg/api/schema/schema.proto
```diff
@@ -365,7 +365,13 @@ message TxHeader {
 }
 
 // TxMetadata contains metadata set to whole transaction
-message TxMetadata {}
+message TxMetadata {
+  // True if this metadata has a truncated transaction
+  bool hasTruncatedTxID = 1;
```
jeroiraz, line 370: probably `hasTruncatedTxID` may not be needed

arriqaaq: yes, makes sense, removed

PR #1492, opened 2023-01-18
store: value cache

jeroiraz:


embedded/store/immustore_test.go
```diff
@@ -4225,3 +4225,26 @@ func TestImmudbStoreWithClosedContext(t *testing.T) {
 		require.ErrorIs(t, err, context.Canceled)
 	})
 }
+
+func TestImmudbStoreWithoutVLogCache(t *testing.T) {
```
arriqaaq, line 4229: may be we can call this test `TestImmudbStoreWithVLogCache` and add additional check if the value is stored in cache and is returned from there, and is being set properly, just to add test for the set part in the vlogcache

arriqaaq (approved):
LGTM, minor comment on a test

PR #1493, opened 2023-01-19
improve replication

jeroiraz:


embedded/sql/engine_test.go
```diff
@@ -5903,42 +5903,43 @@ func setupCommonTestWithOptions(t *testing.T, sopts *store.Options) (*Engine, *s
 	require.NoError(t, err)
 	t.Cleanup(func() { closeStore(t, st) })
 
-	ctx := context.TODO()
 	engine, err := NewEngine(st, DefaultOptions().WithPrefix(sqlPrefix))
 	require.NoError(t, err)
 
-	_, _, err = engine.Exec(ctx, nil, "CREATE DATABASE db1;", nil)
+	_, _, err = engine.Exec(context.Background(), nil, "CREATE DATABASE db1;", nil)
```
arriqaaq, line 5909: do we need a new context every call? Why can't we declare a variable like before and use it throughout?

embedded/sql/engine_test.go
```diff
@@ -6083,15 +6087,15 @@ func TestCopyCatalogToTx(t *testing.T) {
...
 		require.ErrorIs(t, err, store.ErrKeyAlreadyExists)
 
 		// should fail due non-available index
-		_, err = engine.Query(ctx, nil, "SELECT * FROM table1 ORDER BY amount DESC", nil)
+		_, err = engine.Query(context.Background(), nil, "SELECT * FROM table1 ORDER BY amount DESC", nil)
 		require.ErrorIs(t, err, ErrNoAvailableIndex)
 
 		// should use primary index by default
-		r, err := engine.Query(ctx, nil, "SELECT * FROM table1", nil)
+		r, err := engine.Query(context.Background(), nil, "SELECT * FROM table1", nil)
```
arriqaaq, line 6098: I feel we could have made these changes in a different PR, just to not make this big

embedded/store/immustore.go
```diff
@@ -851,7 +851,7 @@ func (s *ImmuStore) syncBinaryLinking() error {
 	}
 	defer s.releaseAllocTx(tx)
 
-	txReader, err := s.newTxReader(s.aht.Size()+1, false, true, tx)
+	txReader, err := s.newTxReader(s.aht.Size()+1, false, true, false, tx)
```
arriqaaq, line 854: I feel we should either declare these true/false as named conditional variables, so that readability and understanding is simple because it is getting complicated reading it
```
DisablePrecommittedTx = false
EnablePrecommittedTx = true

s.newTxReader(s.aht.Size()+1, DisablePrecommittedTx, DisableIntegrityCheck, DisableY, tx)
```

embedded/store/immustore.go
```diff
@@ -2321,7 +2323,7 @@ func (s *ImmuStore) FirstTxSince(ts time.Time) (*TxHeader, error) {
 	for left < right {
 		middle := left + (right-left)/2
 
-		header, err := s.ReadTxHeader(middle, false)
+		header, err := s.ReadTxHeader(middle, false, false)
```
arriqaaq, line 2326: same comment as above, hard to understand with multiple boolean flags on what is happening

embedded/store/immustore.go
```diff
@@ -2550,30 +2552,31 @@ func (s *ImmuStore) ReadValue(entry *TxEntry) ([]byte, error) {
...
+
 	b := make([]byte, entry.vLen)
 
-	bval, err := s.readValueAt(b, entry.vOff, entry.hVal)
+	_, err := s.readValueAt(b, entry.vOff, entry.hVal, false)
 	if err != nil {
 		return nil, err
 	}
 
-	// if bval == 0, the value is nil
```
arriqaaq, line 2560: should not nil be returned if data is nil? We are still returning an array of n bytes which are 0s when data is empty

embedded/store/tx.go
```diff
@@ -389,8 +389,8 @@ func (tx *Tx) Proof(key []byte) (*htree.InclusionProof, error) {
 	return tx.htree.InclusionProof(kindex)
 }
 
-func (tx *Tx) readFrom(r *appendable.Reader) error {
-	tdr := &txDataReader{r: r}
+func (tx *Tx) readFrom(r *appendable.Reader, skipIntegrityCheck bool) error {
+	tdr := &txDataReader{r: r, skipIntegrityCheck: skipIntegrityCheck}
```
arriqaaq, line 393: we can have a factory for this as we have at multiple places

embedded/sql/engine_test.go
jeroiraz: yes, it's not needed, I did it so to have total control of the control on each call, emulating the most typical manner in which is used. It could be as before as well

embedded/store/immustore.go
jeroiraz: I'm totally agree. I think the way to go would be to have optional parameters as it's done in the sdk

embedded/store/immustore.go
jeroiraz: yes, even the previous parameter is not clear. Described a possible approach in the previous comment

embedded/store/immustore.go
jeroiraz: it should be equivalent, when committing a transaction, either nil or empty values are treated the same way. So it should be equivalent when returning the value. Checking the length of the returned value unifies them.

Nevertheless, I though it was done in this PR otherwise it's on the bidirectional streaming one, I have this change already
```
if entry.vLen == 0 {
		return nil, nil
	}
```

So `nil` is returned even before allocating the buffer or reading data.

jeroiraz: just checked, that change is already in this PR at line 2555

pkg/replication/options.go
```diff
@@ -119,6 +125,18 @@ func (o *Options) WithAllowTxDiscarding(allowTxDiscarding bool) *Options {
 	return o
 }
 
+// WithSkipIntegrityCheck disable integrity checks when reading data during replication
+func (o *Options) WithSkipIntegrityCheck(skipIntegrityCheck bool) *Options {
```
arriqaaq, line 129: This should be just WithIntegrityCheck imo, because it can be set to either true or false, adding skip just makes it conforming to skip

arriqaaq (approved):
Lot of changes, we can merge but we need to create an issue for renaming the booleans to named conditional variables to understand what is happening.

PR #1500, opened 2023-01-31
Extend immudb performance tests with immudb replication sync/async

marcosanchotene:
- [x] async
- [x] sync

This is a summary of a sample result of the changed performance tests:

```sh
Write TX/s - no replicas
TX: 18810, KV: 18810, TX/s: 1880.98, KV/S: 1880.98, CPUTime: 3.86, VMM: 2.29G, RSS: 321.01M, Writes (bytes/calls): 14.30M/88987, Reads (bytes/calls): 90.11k/190709

Write KV/s - no replicas
TX: 3438, KV: 3438000, TX/s: 341.78, KV/S: 341777.54, CPUTime: 38.33, VMM: 3.92G, RSS: 1.61G, Writes (bytes/calls): 910.06M/106995, Reads (bytes/calls): 29.41M/120928

Write TX/s - one async replica
TX: 9240, KV: 9240, TX/s: 923.91, KV/S: 923.91, CPUTime: 6.27, VMM: 5.18G, RSS: 1.83G, Writes (bytes/calls): 63.80M/136470, Reads (bytes/calls): 1.86M/285687

Write KV/s - one async replica
TX: 756, KV: 756000, TX/s: 75.60, KV/S: 75596.01, CPUTime: 12.50, VMM: 6.52G, RSS: 3.35G, Writes (bytes/calls): 194.29M/30254, Reads (bytes/calls): 1.13M/111999

Write TX/s - one sync replica
TX: 1296, KV: 1296, TX/s: 129.60, KV/S: 129.60, CPUTime: 19.18, VMM: 9.34G, RSS: 5.35G, Writes (bytes/calls): 522.53M/169316, Reads (bytes/calls): 57.84M/539576

Write KV/s - one sync replica
TX: 268, KV: 268000, TX/s: 26.80, KV/S: 26798.07, CPUTime: 16.37, VMM: 10.26G, RSS: 5.64G, Writes (bytes/calls): 214.32M/110988, Reads (bytes/calls): 16.40M/482183
```

### DIFF
--- a/test/performance-test-suite/pkg/benchmarks/writetxs/benchmark.go
+++ b/test/performance-test-suite/pkg/benchmarks/writetxs/benchmark.go
@@ -47,7 +47,7 @@ type Config struct {
 	KeySize    int
 	ValueSize  int
 	AsyncWrite bool
-	Replica    bool
+	Replica    string
 }
 
 type benchmark struct {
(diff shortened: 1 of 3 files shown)


test/performance-test-suite/pkg/benchmarks/writetxs/benchmark.go
```diff
@@ -130,6 +132,36 @@ func (b *benchmark) Warmup() error {
...
+			WithLogFormat(logger.LogFormatJSON)
+
+		if b.cfg.Replica == "async" {
+			options2.
+				WithReplicationOptions(replicaOptions.WithIsReplica(true))
+		}
+
+		if b.cfg.Replica == "sync" {
+			options2.
+				WithReplicationOptions(replicaOptions.WithIsReplica(true).WithSyncReplication(true))
```
jeroiraz, line 153: primary server also needs to be configured for sync replication

e.g. `WithReplicationOptions(replicaOptions.WithIsReplica(false).WithSyncReplication(true).WithSyncAcks(1))`

jeroiraz (commented):
just to note that the time it takes to generate the client data to be sent is included into the total time

test/performance-test-suite/pkg/runner/benchmarks.go
```diff
@@ -24,21 +24,63 @@ import (
...
+			Replica:    "async",
+		}),
+
+		writetxs.NewBenchmark(writetxs.Config{
+			Name:       "Write TX/s - one sync replica",
+			Workers:    30,
+			BatchSize:  1,
+			KeySize:    32,
+			ValueSize:  128,
+			AsyncWrite: true,
```
tomekkolo, line 72: @jeroiraz wdyt, having those tests with asyncwrite=true would be more reasonable ?

jeroiraz: performance changes a lot when using sync vs async writes, ideally both cases could be covered. If not, I'd suggest to use sync writes i.e. `AsyncWrite: false,` so to be sure we are not downgrading performance with newer changes

jeroiraz (approved):
LGTM, if the goal of this performance tests are to prevent performance drops, I'd suggest to use sync writes or even cover both modes of operation. at least in some cases

jeroiraz (approved):

test/performance-test-suite/pkg/runner/benchmarks.go
marcosanchotene: I've created https://github.com/vchain-us/codenotary-cloud/issues/2401 to add these scenarios when we can.

PR #1503, opened 2023-02-02
Chore improve replication - bidirectional streaming

jeroiraz:


pkg/replication/replicator.go
```diff
@@ -375,25 +392,32 @@ func (txr *TxReplicator) fetchNextTx() error {
 	}
 
 	if syncReplicationEnabled {
-		md := exportTxStream.Trailer()
+		bMayCommitUpToTxID, ok := emd["may-commit-up-to-txid-bin"]
```
arriqaaq, line 395: This metadata returned can be a type and you can add a validate method to do this validation

pkg/server/stream_replication.go
```diff
@@ -26,6 +26,24 @@ import (
 )
 
 func (s *ImmuServer) ExportTx(req *schema.ExportTxRequest, txsServer schema.ImmuService_ExportTxServer) error {
+	return s.exportTx(req, txsServer, true)
+}
+
+func (s *ImmuServer) StreamExportTx(stream schema.ImmuService_StreamExportTxServer) error {
```
arriqaaq, line 32: Can we add a comment somewhere that this is for bidirectional streaming?

arriqaaq (approved):
LGTM, some refactoring is required on the metadata, post that we can merge this

pkg/replication/replicator.go
jeroiraz: metadata is a generic placeholder, in the context of this PR it's limited to implement bidirectional streaming. Having a dynamic checking might be beneficial in the future as the metadata is used extensively, but I'd suggest to keep it simpler in the scope of this PR

arriqaaq: I don't think adding a small function to validate that would make it complex, even if we keep it as a map, because it is a lot of validation happening atm

PR #1505, opened 2023-02-06
use gosec action

jeroiraz:


arriqaaq (dismissed):
Master could be flaky, but should not be a blocker as such

jeroiraz:
> Master could be flaky, but should not be a blocker as such



> Master could be flaky, but should not be a blocker as such



> Master could be flaky, but should not be a blocker as such

it was not removed from master branch

PR #1507, opened 2023-02-06
fix(embedded/store): integrity checks covering empty values

jeroiraz:

### DIFF
--- a/embedded/store/immustore.go
+++ b/embedded/store/immustore.go
@@ -2571,13 +2571,17 @@ func (s *ImmuStore) readValueAt(b []byte, off int64, hvalue [sha256.Size]byte, s
 		val, err := s.vLogCache.Get(off)
 		if err == nil {
 			// the requested value was found in the value cache
-			copy(b, val.([]byte))
+			bval := val.([]byte)
+			n := len(bval)
 
-			if !skipIntegrityCheck && hvalue != sha256.Sum256(b) {
-				return len(b), ErrCorruptedData
+			copy(b, bval)
+
+			// length is compared to cover the corner case where empty value is stored in the cache
+			if !skipIntegrityCheck && (len(b) != n || hvalue != sha256.Sum256(b[:n])) {
+				return n, ErrCorruptedData
 			}
 
-			return len(b), nil
+			return n, nil
 		} else {
 			if !errors.Is(err, cache.ErrKeyNotFound) {
 				return 0, err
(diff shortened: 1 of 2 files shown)


embedded/store/immustore.go
```diff
@@ -2571,13 +2571,17 @@ func (s *ImmuStore) readValueAt(b []byte, off int64, hvalue [sha256.Size]byte, s
...
 			// the requested value was found in the value cache
-			copy(b, val.([]byte))
+			bval := val.([]byte)
+			n := len(bval)
 
-			if !skipIntegrityCheck && hvalue != sha256.Sum256(b) {
-				return len(b), ErrCorruptedData
+			copy(b, bval)
+
+			// length is compared to cover the corner case where empty value is stored in the cache
```
arriqaaq, line 2579: just curious, when can an empty value be stored in the cache? Because the vlogcache is being set when appendData is called. Shouldn't we put a check there too to avoid this scenario?

embedded/store/immustore.go
```diff
@@ -2571,13 +2571,17 @@ func (s *ImmuStore) readValueAt(b []byte, off int64, hvalue [sha256.Size]byte, s
 		val, err := s.vLogCache.Get(off)
 		if err == nil {
 			// the requested value was found in the value cache
-			copy(b, val.([]byte))
+			bval := val.([]byte)
+			n := len(bval)
```
arriqaaq, line 2575: we may not want to redeclare this variable as it is already defined in the return parameter

embedded/store/immustore.go
```diff
@@ -2571,13 +2571,17 @@ func (s *ImmuStore) readValueAt(b []byte, off int64, hvalue [sha256.Size]byte, s
...
+			// length is compared to cover the corner case where empty value is stored in the cache
+			if !skipIntegrityCheck && (len(b) != n || hvalue != sha256.Sum256(b[:n])) {
+				return n, ErrCorruptedData
 			}
 
-			return len(b), nil
+			return n, nil
 		} else {
 			if !errors.Is(err, cache.ErrKeyNotFound) {
 				return 0, err
```
arriqaaq, line 2587: return n, err?

jeroiraz: err is nil and so, having it explicit simplifies reading the code

arriqaaq: it is not really consistent with the changes being made

embedded/store/immustore.go
jeroiraz: empty values are not stored in the cache, it potentially could and it shouldn't cause an issue. However, due to we associate such cache to the vlog, I'm placing the cache after knowing an offset was set

embedded/store/immustore.go
jeroiraz: don't see like that..  

I figured out that truncated transaction in the replicas should receive the value length as it's receiving the digest, otherwise it becomes inconsistent with the data hold in master after truncation, making impossible to differentiate empty value with non-empty ones during integrity checks

PR #1512, opened 2023-02-13
fix(pkg/database): skip eof error during scan

arriqaaq:
eof error, which is returned when data is truncated, should not block reading scanning all keys


Razikus (approved):
LGTM, SonarCloud Quality Gate reports issue, but its fake - new test case covers 2 new cases of EOF

PR #1513, opened 2023-02-14
Test/e2e replication

SimoneLazzaris:
This PR adds an end to end test to check replication and truncation. 
Tests are performed inside a docker container to help isolate the environmennt


test/e2e/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM golang:1.20
```
jeroiraz, line 1: minor but we may want to use the same golang version used to build official docker images?

test/e2e/replication/replic.sh
```diff
@@ -0,0 +1,139 @@
...
+txid() {
+ADDR=$1
+$IMMUCLIENT login -a $ADDR --username immudb --password immudb > /dev/null 2>/dev/null
+$IMMUCLIENT status -a $ADDR --username immudb --password immudb --database repl | awk '/^txID/{print $2}'
+}
+
+TX1=$(txid $PRIMARY_ADDR)
+TX2=$(txid $REPLICA_ADDR)
+echo "Replication in progress ($TX1 / $TX2)"
+while [ "$TX1" != "$TX2" ]
```
jeroiraz, line 125: with sync replication enabled, this condition may always hold, isn't it?

test/e2e/truncation/main.go
```diff
@@ -0,0 +1,135 @@
...
+	k := []byte(fmt.Sprintf("key:%d", time.Now().Unix()))
+	v := randStr(256)
+	_, err := ic.Set(ctx, k, v)
+	if err != nil {
+		log.Fatal("Error inserting data: ", err)
+	}
+}
+
+func flushData(ic immuclient.ImmuClient, ctx context.Context) {
+	err := ic.TruncateDatabase(ctx, config.DBName, config.Retention)
```
jeroiraz, line 114: in the future, it may be good to add additional validations to ensure truncation actually occurred e.g. disk usage was reduced or reading expired data does return an error

SimoneLazzaris: if you look carefully, there *already is* such a test: the `run.sh` [make the test fails](https://github.com/codenotary/immudb/blob/69965270800acb093a143de941b7a02f5c938553/test/e2e/truncation/run.sh#L133) if the older file in the data folder is older than the retention period + purge frequency.

test/e2e/Dockerfile
SimoneLazzaris: Since we are using an older version for building the "production" image, I like using the newer for the tests. We could do the other way around: the point is to be as compatible as possible

test/e2e/replication/replic.sh
jeroiraz: but it won't mean the immediate termination of the task without waiting for all the insertions to be done?

SimoneLazzaris: No, that check is done after the insertion on the primary has ended. It is just to be sure that the replica has caught up. And as you said, in case of synchronous replication, that is (or should be) always true

jeroiraz (approved):

PR #1516, opened 2023-02-20
feat(embedded/object): add object store to embedded pkg

arriqaaq:
This PR adds the object layer on the embedded store


embedded/object/document.go
```diff
@@ -0,0 +1,114 @@
+package object
```
jeroiraz, line 1: probably package name should be `document`?

embedded/object/document.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/tidwall/gjson"
```
jeroiraz, line 9: we should evaluate the use of built-in json support

embedded/object/engine_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+func closeStore(t *testing.T, st *store.ImmuStore) {
+	err := st.Close()
+	if !t.Failed() {
+		// Do not pollute error output if test has already failed
+		require.NoError(t, err)
+	}
+}
+
+func TestCreateCollection(t *testing.T) {
```
jeroiraz, line 20: test doesn't seem to be validating functionality

embedded/object/schema.go
```diff
@@ -0,0 +1,23 @@
+package object
```
jeroiraz, line 1: I think this file is not used

embedded/object/schema.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"github.com/codenotary/immudb/embedded/sql"
+)
+
+var ObjectPrefix = []byte{3}
```
jeroiraz, line 7: the sql engine already includes support for prefix isolation

document engine could just instantiate the engine with a different prefix. Similarly to what is done here but there should not be a need for fine-grained prefixes for tables, columns or indexes.

object prefix may be set in the database package together with the kv, sorted set and sql prefixes already declared.

I'd still create an object engine which can be a wrapper of the sql engine, so at the database package an object engine is instantiated without the knowledge of its internal implementation

embedded/sql/catalog.go
```diff
@@ -29,6 +29,32 @@ import (
 type Catalog struct {
 	dbsByID   map[uint32]*Database
 	dbsByName map[string]*Database
+
+	// prefixes
```
jeroiraz, line 33: related comments in `schema.go` file

embedded/sql/engine.go
```diff
@@ -102,6 +102,12 @@ type Engine struct {
 	distinctLimit int
 	autocommit    bool
 
+	// prefixes
```
jeroiraz, line 105: related comments in `schema.go` file

embedded/sql/sql_tx.go
```diff
@@ -24,6 +24,18 @@ import (
 	"github.com/codenotary/immudb/embedded/store"
 )
 
+func NewSQLTx(engine *Engine, opts *TxOptions, tx *store.OngoingTx, catalog *Catalog, currentDB *Database) *SQLTx {
```
jeroiraz, line 27: probably not being used

engine has a method to create transactions, is this method created with a different logic?

pkg/database/interface.go
```diff
@@ -0,0 +1,115 @@
...
+	IsClosed() bool
+	Close() error
+
+	ObjectDatabase
+}
+
+// ObjectDatabase is the interface for object database
+type ObjectDatabase interface {
+	// GetCollection returns the collection schema
+	GetCollection(ctx context.Context, req *schemav2.CollectionGetRequest) (*schemav2.CollectionInformation, error)
```
jeroiraz, line 107: to be changed in the proto spec, but return type should be CollectionGetResponse, as it will allow further incremental changes in the future and provide a consistent naming among all the methods

pkg/database/interface.go
```diff
@@ -0,0 +1,115 @@
...
+	// CreateCollection creates a new collection
+	CreateCollection(ctx context.Context, req *schemav2.CollectionCreateRequest) error
+
+	// GetDocument returns the document
+	GetDocument(ctx context.Context, req *schemav2.DocumentSearchRequest) (*schemav2.DocumentSearchResponse, error)
+	// CreateDocument creates a new document
+	CreateDocument(ctx context.Context, req *schemav2.DocumentInsertRequest) (string, error)
```
jeroiraz, line 114: return type should be DocumentInsertResponse instead of string

pkg/database/interface.go
```diff
@@ -0,0 +1,115 @@
...
+	"crypto/sha256"
+	"time"
+
+	"github.com/codenotary/immudb/embedded/sql"
+	"github.com/codenotary/immudb/embedded/store"
+	"github.com/codenotary/immudb/pkg/api/schema"
+	schemav2 "github.com/codenotary/immudb/pkg/api/schemav2"
+)
+
+type DB interface {
```
jeroiraz, line 14: good to have interface definition on its own file

pkg/database/object_database.go
```diff
@@ -0,0 +1,356 @@
...
+		// TODO: add support for max length
+		// TODO: add support for auto increment
+		collectionSpec = append(collectionSpec, sql.NewColSpec(name, schType, 0, false, false))
+	}
+
+	// add support for blob
+	// TODO: add support for max length for blob storage
+	collectionSpec = append(collectionSpec, sql.NewColSpec(defaultObjectBLOBField, sql.BLOBType, 10000, false, false))
+
+	_, _, err := d.objectEngine.ExecPreparedStmts(
```
jeroiraz, line 144: concrete implementation should be inside the objectEngine

pkg/database/object_database.go
```diff
@@ -0,0 +1,356 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return table.ColsByName(), nil
+}
+
+// CreateDocument creates a new document
+func (d *db) CreateDocument(ctx context.Context, req *schemav2.DocumentInsertRequest) (string, error) {
+	tx, err := d.objectEngine.NewTx(ctx, sql.DefaultTxOptions().WithReadOnly(true))
```
jeroiraz, line 176: hm it should not work with a readOnly tx, if it's working then there may be an issue in ongoingTx or engine.NewTx method...

jeroiraz: concrete implementation may be inside the engine

pkg/database/object_database.go
```diff
@@ -0,0 +1,356 @@
...
+			vtype, err := valueTypeToFieldValue(v)
+			if err != nil {
+				return nil, err
+			}
+			document.Fields[cols[i].Name] = vtype
+		}
+		resp.Results = append(resp.Results, document)
+
+		if l == d.maxResultSize {
+			return nil, fmt.Errorf("%w: found at least %d rows (the maximum limit). "+
```
jeroiraz, line 311: `rows` -> `documents`

embedded/object/document.go
arriqaaq: I don't have any strong opinions on this, changing it to document

embedded/object/engine_test.go
arriqaaq: Yes, I am adding more to it

embedded/object/schema.go
arriqaaq: I see your point, I was in two minds about the same about using the prefixes or they have to be different at granular level. Thanks for the clarification, I'll just remove the granularities.

I agree, it will be better to add functionality to it separately

embedded/sql/catalog.go
arriqaaq: resolved

embedded/sql/engine.go
arriqaaq: resolved

pkg/database/database.go
```diff
@@ -150,6 +63,7 @@ type db struct {
 	st *store.ImmuStore
 
 	sqlEngine     *sql.Engine
+	objectEngine  *sql.Engine
```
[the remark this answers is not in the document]
arriqaaq: made it to documentEngine *document.Engine

pkg/util/flatten.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func DefaultOption() *Option {
+	return &Option{
+		Delimiter: ".",
+	}
+}
+
+// Flatten takes a data map and returns a flattened map.
+func Flatten(data map[string]interface{}, opts *Option) (m map[string]interface{}) {
```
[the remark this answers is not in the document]
arriqaaq: yes, removing it, was used in a previous iteration to flatten json and add individual keys

pkg/database/object_database.go
arriqaaq: done

embedded/sql/sql_tx.go
arriqaaq: not required, was dead code from previous implementation

pkg/database/interface.go
arriqaaq: moved

pkg/database/interface.go
arriqaaq: It returns a VerifiableTx for now, so did not know what to use here, will change it post discussion with adam

pkg/database/object_database.go
arriqaaq: so this transaction is used just to read the schema, it is not being used for commit

embedded/object/engine_test.go
arriqaaq: added functional tests, will make a separate PR to add more, have a list created

pkg/database/interface.go
arriqaaq: done

arriqaaq: fixed

pkg/database/interface.go
arriqaaq: fixed, but needs work in a separate PR to return proof

pkg/database/object_database.go
arriqaaq: moved

pkg/database/database.go
```diff
@@ -147,11 +150,13 @@ type replicaState struct {
 
 // IDB database instance
 type db struct {
-	st *store.ImmuStore
+	st          *store.ImmuStore
+	objectStore *store.ImmuStore
```
arriqaaq, line 154: @jeroiraz should we have a separate store for objects, or should there be a single store? If we use a single store, the starting tx id becomes 3, instead of 2 because there is one more commit for document store

pkg/database/database.go
```diff
@@ -349,9 +390,20 @@ func NewDB(dbName string, multidbHandler sql.MultiDBHandler, op *Options, log lo
 		if err != nil {
 			return nil, logErr(dbi.Logger, "Unable to open database: %s", err)
 		}
+
+		// _, _, err = dbi.documentEngine.ExecPreparedStmts(context.Background(), nil, []sql.SQLStmt{&sql.CreateDatabaseStmt{DB: dbInstanceName}}, nil)
```
arriqaaq, line 394: @jeroiraz would we need to init a new database for document engine?

arriqaaq: resolved with prefix bug resolution

PR #1517, opened 2023-02-20
chore: upgrade go.mod pkgs for dependabot security alerts

arriqaaq:


arriqaaq (commented):
This just upgrades the crypto library for fixing some of the dependabot alerts

PR #1526, opened 2023-02-28
Upgraded x/net and x/text

SimoneLazzaris:
This PR upgrades vulnerable packages `golang.org/x/net` and `golang.org/x/text` , and some others.


SimoneLazzaris:
We cannot upgrade packages if we want to keep compatibility with 1.15

PR #1537, opened 2023-03-02
Authentication via strings, all object storage methods authenticated, gateway authentication

Razikus:
This PR adds implementation on LoginV2 method, tests, and option to include session ID inside header to authenticate incoming request


pkg/server/objects_operations.go
```diff
@@ -2,38 +2,78 @@ package server
...
 import (
 	"context"
+	"time"
 
 	"github.com/codenotary/immudb/pkg/api/schema"
 	"github.com/codenotary/immudb/pkg/api/schemav2"
 	"github.com/golang/protobuf/ptypes/empty"
 )
 
 func (s *ImmuServer) LoginV2(ctx context.Context, loginReq *schemav2.LoginRequest) (*schemav2.LoginResponseV2, error) {
```
jeroiraz, line 12: what about OpenSession rather than login?

pkg/server/objects_operations.go
```diff
@@ -2,38 +2,78 @@ package server
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	expirationTimestamp := int32(0)
+	if s.Options.SessionsOptions.MaxSessionInactivityTime > 0 {
+		expirationTimestamp = int32(time.Now().Add(s.Options.SessionsOptions.MaxSessionInactivityTime).Unix())
+	}
+	return &schemav2.LoginResponseV2{
+		Token:               session.SessionID,
```
jeroiraz, line 28: what about being explicit with the value included into the response? i.e. SesssionUUID or SessionID

pkg/server/objects_operations.go
```diff
@@ -2,38 +2,78 @@ package server
...
+		Username:     username,
+		Password:     password,
+		DatabaseName: loginReq.Database,
+	})
+	if err != nil {
+		return nil, err
+	}
+	expirationTimestamp := int32(0)
+	if s.Options.SessionsOptions.MaxSessionInactivityTime > 0 {
+		expirationTimestamp = int32(time.Now().Add(s.Options.SessionsOptions.MaxSessionInactivityTime).Unix())
```
jeroiraz, line 25: expirationTime is not exactly the same as inactivity time

Razikus: What should we here return in order to inform user explicitly what he should do?
I mean - expirationTimestamp is a hint, that in x minutes he should refresh token / open new session

pkg/server/objects_operations.go
Razikus: Could be, but again (like in the next comment) its again a path to learn user about new concepts which he doesnt know in REST

pkg/server/objects_operations.go
jeroiraz: the inactivity time is setting for how long the server should keep the session alive despite the client has not sent any request for a while (flag --session-timeout).
Expiration time refers to the longest age of a session, even if some activity is continuously done (flag: --max-session-age-time duration)

Probably it's worth sending both values

pkg/server/objects_operations.go
Razikus: Resolved in 
https://github.com/codenotary/immudb/pull/1546

pkg/server/objects_operations.go
Razikus: Resolved in https://github.com/codenotary/immudb/pull/1546

PR #1541, opened 2023-03-06
fix(pkg/client): handle broken keep-alive connection error

arriqaaq:
Handling broken keep-alive connection error

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -744,7 +744,11 @@ func (c *immuClient) Disconnect() error {
 
 // IsConnected checks whether the client is connected to the server.
 func (c *immuClient) IsConnected() bool {
-	return c.clientConn != nil && c.ServiceClient != nil
+	isConnected := c.clientConn != nil && c.ServiceClient != nil
+	if c.HeartBeater != nil {
+		isConnected = isConnected && c.HeartBeater.IsAlive()
+	}
+	return isConnected
 }
 
 // WaitForHealthCheck waits for up to Options.HealthCheckRetries seconds to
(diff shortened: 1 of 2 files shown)


arriqaaq:
Hi @marcosQuesada 

I am trying to understand this error better, of connection drops when keep-alive fails. Drafting a PR based on my assumptions. Please check and let me know your thoughts

marcosQuesada:
Hi @arriqaaq !

This was the original PR: https://github.com/codenotary/immudb/pull/1416
The original idea was to let the application react on immudb conn drop, in that case as workaround, App was panicking and thus restarted initializing again the immudb connection.

When we talk about this the agreement was to implement a connection pool, not just a single connection, as Postgres does in the backgrounds, iterating connections in the pool would lead to new connections against immudb.

Let me know if you need more info about it

arriqaaq:
Thank you @marcosQuesada for the details, I now recall this issue. I see you already have worked on a PR for this. The session management/connection pool feature was put in the backlog due to the recent requirements, but I'll discuss and update on this

arriqaaq:
@marcosQuesada does this help in any way?

https://github.com/codenotary/immudb/pull/1421

jeroiraz:
> Hi @arriqaaq !
> 
> This was the original PR: #1416 The original idea was to let the application react on immudb conn drop, in that case as workaround, App was panicking and thus restarted initializing again the immudb connection.
> 
> When we talk about this the agreement was to implement a connection pool, not just a single connection, as Postgres does in the backgrounds, iterating connections in the pool would lead to new connections against immudb.
> 
> Let me know if you need more info about it

I think cited PR is good. I made some comments there

arriqaaq:
Thanks @jeroiraz ! Closing this in favor of cited PR by @marcosQuesada

PR #1546, opened 2023-03-08
Adjusting Documents/Objects API

Razikus:


Makefile
```diff
@@ -72,23 +86,23 @@ webconsole/default:
 
 .PHONY: immuclient
 immuclient:
-	$(GO) build -v -ldflags '$(V_LDFLAGS_COMMON)' ./cmd/immuclient
+	$(GO) build $(IMMUDB_BUILD_TAGS) -v -ldflags '$(V_LDFLAGS_COMMON)' ./cmd/immuclient
```
jeroiraz, line 89: are build tags relevant for immuclient now?

Makefile
```diff
@@ -72,23 +86,23 @@ webconsole/default:
...
 
 .PHONY: immuadmin
 immuadmin:
-	$(GO) build -v -ldflags '$(V_LDFLAGS_COMMON)' ./cmd/immuadmin
+	$(GO) build $(IMMUDB_BUILD_TAGS) -v -ldflags '$(V_LDFLAGS_COMMON)' ./cmd/immuadmin
```
jeroiraz, line 93: are build tags relevant for immuadmin now?

Makefile
```diff
@@ -72,23 +86,23 @@ webconsole/default:
...
 
 .PHONY: immutest
 immutest:
-	$(GO) build -v -ldflags '$(V_LDFLAGS_COMMON)' ./cmd/immutest
+	$(GO) build $(IMMUDB_BUILD_TAGS) -v -ldflags '$(V_LDFLAGS_COMMON)' ./cmd/immutest
 
 .PHONY: immuclient-static
 immuclient-static:
-	CGO_ENABLED=0 $(GO) build -a -ldflags '$(V_LDFLAGS_STATIC)' ./cmd/immuclient
+	CGO_ENABLED=0 $(GO) build $(IMMUDB_BUILD_TAGS) -a -ldflags '$(V_LDFLAGS_STATIC)' ./cmd/immuclient
```
jeroiraz, line 105: related to above comments

Makefile
```diff
@@ -97,7 +111,7 @@ immuclient-fips:
 
 .PHONY: immuadmin-static
 immuadmin-static:
-	CGO_ENABLED=0 $(GO) build -a -ldflags '$(V_LDFLAGS_STATIC)' ./cmd/immuadmin
+	CGO_ENABLED=0 $(GO) build $(IMMUDB_BUILD_TAGS) -a -ldflags '$(V_LDFLAGS_STATIC)' ./cmd/immuadmin
```
jeroiraz, line 114: related to above comments

Makefile
```diff
@@ -115,7 +129,7 @@ immudb-fips: webconsole
 
 .PHONY: immutest-static
 immutest-static:
-	CGO_ENABLED=0 $(GO) build -a -ldflags '$(V_LDFLAGS_STATIC)' ./cmd/immutest
+	CGO_ENABLED=0 $(GO) build $(IMMUDB_BUILD_TAGS) -a -ldflags '$(V_LDFLAGS_STATIC)' ./cmd/immutest
```
jeroiraz, line 132: related to above comments

Makefile
```diff
@@ -175,9 +189,100 @@ build/codegen:
 	  --doc_out=pkg/api/schema --doc_opt=markdown,docs.md \
 	  --plugin=protoc-gen-doc=$(PWD)/scripts/protoc-gen-doc
 
+.PHONY: build/codegendocuments
+build/codegendocuments:
+	$(PROTOC) -I pkg/api/documentsschema/ pkg/api/documentsschema/documentsschema.proto \
```
jeroiraz, line 194: probably `pkg/api/documentschema` is slightly more readable...

cmd/immudb/command/init.go
```diff
@@ -84,6 +84,8 @@ func (cl *Commandline) setupFlags(cmd *cobra.Command, options *server.Options) {
 	cmd.Flags().Duration("session-timeout", 2*time.Minute, "session timeout is a duration after which an inactive session is forcibly closed by the server")
 	cmd.Flags().Duration("sessions-guard-check-interval", 1*time.Minute, "sessions guard check interval")
 	cmd.Flags().MarkHidden("sessions-guard-check-interval")
+	cmd.Flags().Bool("reflection", options.ReflectionServerEnabled, "GRPC reflection server enabled")
```
jeroiraz, line 87: `reflection` -> `grpc-reflection` ?

pkg/api/authorizationschema/authorizationschema.proto
```diff
@@ -0,0 +1,41 @@
...
+message OpenSessionResponseV2 {
+  string token = 1;
+  string serverUUID = 2;
+  int32 expirationTimestamp = 3;
+  int32 inactivityTimestamp = 4;
+
+}
+
+service AuthorizationService {
+  rpc OpenSessionV2(OpenSessionRequestV2) returns (OpenSessionResponseV2) {
```
jeroiraz, line 32: I think it's ok. However splitting authentication with session handling could be analyzed. As users find confusing to open a session against `defaultdb` in order to create a database the very first time... That can be done in another PR is needed

pkg/api/authorizationschema/authorizationschema.swagger.json
```diff
@@ -0,0 +1,120 @@
+{
```
jeroiraz, line 1: in another PR, generated code might be even removed from version control

pkg/api/documentsschema/documentsschema.proto
```diff
@@ -0,0 +1,252 @@
+syntax = "proto3";
+
+package immudb.documentsschema;
```
jeroiraz, line 3: minor (not required) but it's rare to me `documentsschema`, probably `documentschema` or under `pkg/api/document` or similar thing could be simpler

pkg/api/documentsschema/documentsschema.proto
```diff
@@ -0,0 +1,252 @@
...
+  security: {
+    security_requirement: {
+      key: "bearer"
+    }
+  }
+};
+
+message DocumentInsertRequest {
+  string collection = 1;
+  repeated google.protobuf.Struct document = 2;
```
jeroiraz, line 38: just to be aware of: it won't allow multi-collection insertions

pkg/api/documentsschema/documentsschema.proto
```diff
@@ -0,0 +1,252 @@
...
+  GT = 1;
+  GTE = 2;
+  LT = 3;
+  LTE = 4;
+  LIKE = 5;
+}
+
+message IndexValue {
+  oneof kind {
+    google.protobuf.NullValue null_value = 1;
```
jeroiraz, line 57: not sure I understand the need of `null_value`

pkg/api/documentsschema/documentsschema.proto
```diff
@@ -0,0 +1,252 @@
...
+  repeated DocumentQuery query = 2;
+  uint32 page = 3;
+  uint32 perPage = 4;
+}
+
+message DocumentSearchResponse {
+  repeated google.protobuf.Struct results = 1;
+  uint32 page = 2;
+  uint32 perPage = 3;
+  uint32 entriesLeft = 4;
```
jeroiraz, line 82: to be aware of: determining `entriesLeft` may be hard to achieve...

jeroiraz: is it worth to include page and perPage in the response?

pkg/api/documentsschema/documentsschema.proto
```diff
@@ -0,0 +1,252 @@
...
+message DocumentAuditResponse {
+  repeated DocumentAudit results = 1;
+  uint32 page = 2;
+  uint32 perPage = 3;
+  uint32 entriesLeft = 4;
+}
+
+message DocumentAudit {
+  google.protobuf.Struct value = 1;
+  uint64 transactionID = 2;
```
jeroiraz, line 115: probably it's better keep transactionIDs out of document APIs if simplicity is one of the main objectives. Working with document uuid or primary key might be simpler

pkg/api/documentsschema/documentsschema.proto
```diff
@@ -0,0 +1,252 @@
...
+enum IndexType {
+  DOUBLE = 0;
+  INTEGER = 1;
+  STRING = 2;
+}
+
+message CollectionCreateRequest {
+  string name = 1;
+  map<string, IndexOption> primaryKeys = 2;
+  map<string, IndexOption>  indexKeys = 3;
```
jeroiraz, line 132: to be aware of: it won't be possible to define compound indexes

pkg/api/documentsschema/documentsschema.proto
```diff
@@ -0,0 +1,252 @@
...
+message CollectionListResponse {
+  repeated CollectionInformation collections = 1;
+}
+
+message CollectionDeleteRequest {
+  string name = 1;
+}
+
+message CollectionDeleteResponse {
+  google.protobuf.Empty response = 1;
```
jeroiraz, line 165: empty message should be just ok

pkg/server/authorization_operations.go
```diff
@@ -0,0 +1,42 @@
+package server
```
jeroiraz, line 1: copyright notice missing

Makefile
Razikus: The problem is following:
Statik can only handle multiple files in this way:
https://github.com/rakyll/statik/issues/56

So basically we can do it only at init time

Because of that I had to add imports like
```_ "github.com/codenotary/immudb/webconsole/webconsoleembedded"```

when immudb is builded with let say:
SWAGGER=true WEBCONSOLE=default
It will generate webconsoleembedded content, but only for:
```
//go:build webconsole
// +build webconsole
```
And later on when building immuclient without these tags - it will not include file

that leads into a next problem
```
package github.com/codenotary/immudb/cmd/immuclient
        imports github.com/codenotary/immudb/cmd/immuclient/command
        imports github.com/codenotary/immudb/cmd/immuclient/audit
        imports github.com/codenotary/immudb/pkg/server
        imports github.com/codenotary/immudb/webconsole
        imports github.com/codenotary/immudb/webconsole/webconsoleembedded: build constraints exclude all Go files in /home/razikus/immudb/webconsole/webconsoleembedded
make: *** [Makefile:89: immuclient] Error 1
```

And thats true because its built with different tags, and statik is not generated by default

it will allow to build immuclient only if before immudb was built without tags

Some solutions:
- add tags to makefile (done)
- generate webconsole and swagger defaults always when building other things (basically adding env go xxx to all targets)
- change go 1.15 to go 1.16 to allow go embed https://pkg.go.dev/embed (this will remove this unnecessary import as this support multiple packages out of the box)
- bundle webconsole and swagger together (would rather avoid)

File: Makefile [hunk 18]
```diff
@@ -72,23 +86,23 @@ webconsole/default:
...
 
 .PHONY: immudb
-immudb: webconsole
+immudb: webconsole swagger
 	$(GO) build $(IMMUDB_BUILD_TAGS) -v -ldflags '$(V_LDFLAGS_COMMON)' ./cmd/immudb
 
 .PHONY: immutest
 immutest:
-	$(GO) build -v -ldflags '$(V_LDFLAGS_COMMON)' ./cmd/immutest
+	$(GO) build $(IMMUDB_BUILD_TAGS) -v -ldflags '$(V_LDFLAGS_COMMON)' ./cmd/immutest
```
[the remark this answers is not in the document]
Razikus: comment left on related

Makefile
Razikus: comment left on related

Makefile
Razikus: comment left on related

Makefile
Razikus: comment left on related

Makefile
Razikus: comment left on related

pkg/api/documentsschema/documentsschema.proto
Razikus: This endpoint is audit-related, so transaction ID is needed here to later verify on that

cmd/immudb/command/init.go
Razikus: resolved

pkg/api/documentsschema/documentsschema.proto
Razikus: Resolved

pkg/api/documentsschema/documentsschema.proto
Razikus: There is no way to distinguish between null and 0 in GRPC
If we want to allow null values for index fields - then we will need that, otherwise not

pkg/api/documentsschema/documentsschema.proto
Razikus: Yes and no again

in terms of simplicity yes, because when you are doing for example N requests at once for example with multiprocessing pool or multithreading pool - the order of returned response is not guaranted
so when perpage and page is not returned with response it will force user to add it these fields himself to distinguish fields

or for example with
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all
Promise all will just return value of responses
Again - force user to either include values into reponse, or wrap reponse in something

Not big deal here, we can remove, but i dont see minuses from that

pkg/api/documentsschema/documentsschema.proto
Razikus: Resolved

pkg/api/documentsschema/documentsschema.proto
Razikus: To discuss, currently not important i believe

pkg/api/documentsschema/documentsschema.proto
Razikus: I think we will not allow that

pkg/api/authorizationschema/authorizationschema.swagger.json
Razikus: hm, why not

pkg/api/authorizationschema/authorizationschema.proto
Razikus: On the other hand - I find confusing opening session and then selectin database
maybe we can do this parameter optional, or clearly state this in documentation

Makefile
Razikus: Resolved to documentschema

Makefile
jeroiraz: actually that's revealing a dependency on webconsole that may not be there...

pkg/api/authorizationschema/authorizationschema.proto
jeroiraz: the idea is to have authentication (without database selection), then you can open a session against a database or create a database without having a session... that will avoid the need of opening a session against defaultdb to be able to create another database

pkg/api/authorizationschema/authorizationschema.swagger.json
jeroiraz: it can be left, it's not an issue

pkg/api/documentsschema/documentsschema.proto
jeroiraz: so it's to be explicit on the null value then? is it required for searching for a document with a null value in a particular attribute?

jeroiraz: in fact, it's actually equivalent to what's defined for SQLValue

pkg/api/documentschema/documentschema.proto
```diff
@@ -0,0 +1,267 @@
...
+    double number_value = 2;
+    string string_value = 3;
+    bool bool_value = 4;
+  }
+}
+
+message DocumentQuery {
+  string field = 1;
+  QueryOperator operator = 2;
+  google.protobuf.Value value = 3;
```
jeroiraz, line 83: I'm wondering why value in DocumentQuery has a different type than in DocumentProofRequest and  DocumentAuditRequest.

Not sure if it's possible to fetch an object at a specific revision

Razikus: Actually could be the same type, i was wondering about explicitily making that user needs to provide primary keys, but probably more "natural" would be for him to provide document

Razikus:
As discussed - merge to feat/objects, some issues will be resolved in next stage

PR #1551, opened 2023-03-11
feat(embedded/sql): float support

jeroiraz:
This PR contains the work done in PR #1469

Several conflicts has been solved with current master branch and it's hold in this branch to validate conflicts has been properly fixed and to easily do any further improvement


embedded/sql/engine.go
```diff
@@ -502,246 +501,3 @@ func (e *Engine) CopyCatalogToTx(ctx context.Context, tx *store.OngoingTx) error
 
 	return nil
 }
-
-// addSchemaToTx adds the schema to the ongoing transaction.
```
arriqaaq, line 506: somehow this code seems to be removed?

arriqaaq: please ignore, found it in catalog.go

embedded/sql/type_conversion.go
```diff
@@ -0,0 +1,165 @@
...
+
+		if src == VarcharType {
+			return func(val TypedValue) (TypedValue, error) {
+				if val.RawValue() == nil {
+					return &NullValue{t: TimestampType}, nil
+				}
+
+				str := val.RawValue().(string)
+				for _, layout := range []string{
+					"2006-01-02 15:04:05.999999",
```
arriqaaq, line 62: these can be kept constants? Else this array would get initialised every call

arriqaaq (dismissed):
LGTM, added few comments, post fixing that we can merge

embedded/sql/type_conversion.go
jeroiraz: will add constant for code readability. Not sure if it will make a difference in allocations, what do you think about it @Razikus  ;)

embedded/sql/catalog.go
```diff
@@ -984,101 +992,173 @@ func EncodeID(id uint32) []byte {
...
+	KeyValPrefixNotNull    byte = 0x80
+	KeyValPrefixUpperBound byte = 0xFF
+)
+
+func EncodeValueAsKey(val TypedValue, colType SQLValueType, maxLen int) ([]byte, error) {
+	return EncodeRawValueAsKey(val.RawValue(), colType, maxLen)
+}
+
+// EncodeRawValueAsKey encodes a value in a b-tree meaningful way.
+func EncodeRawValueAsKey(val interface{}, colType SQLValueType, maxLen int) ([]byte, error) {
```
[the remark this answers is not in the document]
jeroiraz: it's actually validated by multiple tests at sql level, when checking returned data matches inserted one...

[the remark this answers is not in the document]
jeroiraz: will add one directly validating encoding decoding cycle

embedded/sql/type_conversion.go
Razikus: in terms of performance probably it will 0-affect due to compilator optimizations

PR #1557, opened 2023-03-16
chore(pkg/database): reload document engine on replica with sql engine

arriqaaq:


pkg/database/sql.go
```diff
@@ -36,6 +36,14 @@ func (d *db) reloadSQLCatalog(ctx context.Context) error {
 	if err == sql.ErrDatabaseDoesNotExist {
 		return ErrSQLNotReady
 	}
+	if err != nil {
+		return err
+	}
+
+	err = d.documentEngine.SetCurrentDatabase(ctx, dbInstanceName)
```
jeroiraz, line 43: it won't be required when merging the single-db engine

PR #1562, opened 2023-03-17
chore(embedded/document): add document id generation

arriqaaq:


embedded/document/document_id.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+/*
+	DocumentID is a 16-byte identifier that is automatically generated
+	by the server upon document insertion. The 16-byte DocumentID is composed of:
+
+		1) A 4-byte timestamp value, representing the time the document was created,
+		measured in seconds since the Unix epoch.
+		2) A 4-byte random value, generated using a secure random number generator.
+		3) A 8-byte value, initialized to the previous transaction id.
```
jeroiraz, line 20: using 8-bytes for timestamp might be better

it may be better to include txid after timestamp

embedded/document/document_id.go
```diff
@@ -0,0 +1,90 @@
...
+
+	The timestamp portion of the DocumentID allows documents to be sorted by creation time, which
+	can be useful for certain types of queries. The random value and tx id portions of the DocumentID
+	ensure that the identifier is unique across all documents in a collection, even if multiple documents
+	are inserted in the same second.
+
+	The 5-byte random value provides a total of 32 bits of randomness, which gives a total of 2^32
```
jeroiraz, line 27: 4-byte random value

embedded/document/document_id.go
jeroiraz: it might also be good to suffix with a monotically increasing number, such number may be restarted when immudb starts or when the txid fragment changes. In such case, not sure of the need of the random piece...

arriqaaq: fixed, added monotically increasing number. Time stamp I feel can be 4 bytes

embedded/document/document_id.go
arriqaaq: fixed

PR #1565, opened 2023-03-17
chore(pkg/database): reload document engine on replica with sql engine

arriqaaq:


arriqaaq:
closing as db initialisaiton is not required now

PR #1607, opened 2023-04-07
chore: handle table creation with deleted table name

arriqaaq:


embedded/sql/catalog.go
```diff
@@ -32,9 +32,16 @@ import (
 type Catalog struct {
 	prefix []byte
 
-	tables       []*Table
-	tablesByID   map[uint32]*Table
-	tablesByName map[string]*Table
+	tables        []*Table
+	tablesByID    map[uint32]*Table
+	tablesByName  map[string]*Table
+	deletedTables map[string]*Table
```
arriqaaq, line 38: this should fail, needs work to be done

arriqaaq: the id for new table is calculated on length of tables array, which is problematic

PR #1608, opened 2023-04-09
fix(embedded/sql): crash when RowReader.Read() returns error

### LINKED ISSUE #1508: Immudb crash when callins SQLExec with `nil` parameter
**What happened**

Immudb server crashed when executing an (invalid) SQLExec query

Invalid or inconsistent queries should rejected in the client SDK, and definitively shouldn't crash the server

Create this table on immudb, `defaultdb` database: 
```sql
CREATE TABLE IF NOT EXISTS t (id VARCHAR[256], value INTEGER, PRIMARY KEY id)
```
Then execute this go code:
```go
package main

import (
	"context"
	immuclient "github.com/codenotary/immudb/pkg/client"
	"log"
)

func main() {
	opts := immuclient.DefaultOptions().WithAddress("127.0.0.1").WithPort(3322)
	client := immuclient.NewClient().WithOptions(opts)
	ctx := context.Background()
	err := client.OpenSession(ctx, []byte("immudb"), []byte("immudb"), "defaultdb")
	if err != nil {
		log.Fatalln("Failed to connect. Reason:", err)
	}
	_, err = client.SQLExec(ctx, "update t set value=@val where id=@id", nil)
	if err != nil {
		log.Fatalln("Error", err)
	}
}
```
**Environment**
```shell
# run "immu* version" and copy/paste the output here
$ ./immudb-v1.4.1-linux-amd64-static version
immudb 1.4.1
Commit  : 99e4f1ea5970177be5287da2529849cd5ee64696
Built at: Mon, 21 Nov 2022 14:50:26 CET
Static  : true
```

knielsen:
One way an error can be returned is when an SQL parameter in a WHERE statement is missing.

fixes #1508

### DIFF
--- a/embedded/sql/engine_test.go
+++ b/embedded/sql/engine_test.go
@@ -1654,6 +1654,27 @@ func TestUpdate(t *testing.T) {
 	})
 }
 
+func TestErrorDuringUpdate(t *testing.T) {
+	engine := setupCommonTest(t)
+
+	_, _, err := engine.Exec(
+		context.Background(), nil,
+		`
+		create table mytable(id varchar[256], value integer, primary key id);
+		insert into mytable(id, value) values('aa',12), ('ab',13);
+	`, nil)
+	require.NoError(t, err)
+
+	_, _, err = engine.Exec(context.Background(), nil, "update mytable set value=@val where id=@id", nil)
+	require.ErrorIs(t, err, ErrMissingParameter)
+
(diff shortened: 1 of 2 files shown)


jeroiraz (dismissed):

knielsen:
Hm, I added a test case, didn't notice your quick review before I forced-pushed it.

 - Kristian.

jeroiraz (approved):

PR #1609, opened 2023-04-10
Fix/table id generation

arriqaaq:


embedded/sql/catalog.go
```diff
@@ -251,10 +253,19 @@ func (catlg *Catalog) newTable(name string, colsSpec []*ColSpec) (table *Table,
...
 	}
 
-	id := len(catlg.tables) + 1
+	// Generate a new ID for the table by incrementing the 'tableCount' variable of the 'catalog' instance.
+	id := (catlg.tableCount + 1)
+
+	// This code is attempting to check if a table with the given id already exists in the Catalog.
+	// If the function returns nil for err, it means that the table already exists and the function
+	// should return an error indicating that the table cannot be created again.
+	_, err = catlg.GetTableByID(id)
```
jeroiraz, line 262: this should never happen actually, if it does it'll be a bug, right?

embedded/sql/catalog_test.go
```diff
@@ -130,3 +132,138 @@ func TestEncodeRawValueAsKey(t *testing.T) {
...
+	})
+
+	t.Run("table count should not increase on adding existing table", func(t *testing.T) {
+		tx, err := engine.NewTx(context.Background(), DefaultTxOptions())
+		require.NoError(t, err)
+		defer tx.Cancel()
+		catlog := tx.catalog
+
+		for _, v := range []string{"table1", "table2", "table3"} {
+			_, err := catlog.newTable(v, []*ColSpec{{colName: "id", colType: IntegerType}})
```
jeroiraz, line 177: it may be better to create from a sql statement as the actual creation is triggered from there.

Also it would be useful to have a transaction creating a table but being cancelled instead of committed, which may not increase the counter.

Additionally, check the case of two concurrent transactions creating tables

embedded/sql/catalog_test.go
```diff
@@ -130,3 +132,138 @@ func TestEncodeRawValueAsKey(t *testing.T) {
...
+				CREATE TABLE `+v+` (
+					id INTEGER AUTO_INCREMENT,
+					PRIMARY KEY(id)
+				)`, nil)
+			require.NoError(t, err)
+			totalTablesCount++
+		}
+	})
+
+	t.Run("table count should decrease on dropping table", func(t *testing.T) {
```
jeroiraz, line 210: should not decrease?

embedded/sql/catalog_test.go
```diff
@@ -130,3 +132,138 @@ func TestEncodeRawValueAsKey(t *testing.T) {
...
+		for _, v := range activeTables {
+			require.True(t, catlog.ExistTable(v))
+		}
+
+		for _, v := range deleteTables {
+			require.False(t, catlog.ExistTable(v))
+		}
+	})
+
+	t.Run("adding new table should not increase table count", func(t *testing.T) {
```
jeroiraz, line 242: should increase?...

embedded/sql/catalog.go
arriqaaq: It was a bug in the previous way the tables were being generated, hence I'd added it. At least it helps catch it

embedded/sql/catalog_test.go
arriqaaq: ack, changed

embedded/sql/catalog_test.go
arriqaaq: ack, changed

embedded/sql/catalog_test.go
arriqaaq: Mentioned the issue here

https://github.com/orgs/codenotary/projects/10?pane=issue&itemId=25297372

PR #1617, opened 2023-04-13
[WIP] chore: return sql reader on document search

arriqaaq:


embedded/document/engine.go
```diff
@@ -444,55 +444,27 @@ func (e *Engine) upsertDocument(ctx context.Context, collectionName string, doc
 	return docID, txID, rev, nil
 }
 
-func (e *Engine) GetDocuments(ctx context.Context, collectionName string, queries []*Query, pageNum int, itemsPerPage int) ([]*structpb.Struct, error) {
+func (e *Engine) GetDocuments(ctx context.Context, collectionName string, queries []*Query) (sql.RowReader, error) {
```
jeroiraz, line 447: sql row reader should not be exposed but a documentReader can wrap it

jeroiraz: I'd think limit and offset still applies ...

embedded/document/engine.go
```diff
@@ -745,3 +717,32 @@ func (e *Engine) readMetadataAndValue(key []byte, atTx uint64, skipIntegrityChec
 
 	return entry.Metadata(), v, nil
 }
+
+// ReadStructMessagesFromReader reads a number of messages from a reader and returns them as a slice of Struct messages.
+func ReadStructMessagesFromReader(ctx context.Context, reader sql.RowReader, count int) ([]*structpb.Struct, error) {
```
jeroiraz, line 722: this would most of the steps of the Read() method in the documentReader...

pkg/server/documents_operations.go
```diff
@@ -165,3 +164,79 @@ func (s *ImmuServer) CollectionUpdate(ctx context.Context, req *documentschema.C
...
+		delErr := sess.DeletePaginatedReader(queryName)
+		s.Logger.Errorf("error deleting paginated reader: %v", delErr)
+	}
+
+	return &documentschema.DocumentSearchResponse{
+		Results: results,
+	}, err
+}
+
+func generateQueryName(req *documentschema.DocumentSearchRequest) string {
```
jeroiraz, line 233: the idea of generating an identifier from the request is interesting but it might lead to unexpected results e.g. same query requested in the session at different ranges (offsets), that query will be considered to be the same by the server.
Probably it's safer to have a unique id for the reader that it has to be provided in the next requests

embedded/document/engine.go
arriqaaq: I have added offset, but limit can' think of the use case for it

pkg/server/documents_operations.go
arriqaaq: makes sense, I was assuming the requests would be unique, but this method sounds better, added it

embedded/document/engine.go
```diff
@@ -414,55 +419,27 @@ func (e *Engine) upsertDocument(ctx context.Context, collectionName string, doc
...
-		err = json.Unmarshal(docBytes, doc)
-		if err != nil {
-			return nil, err
-		}
-
-		results = append(results, doc)
-	}
-
-	return results, nil
+	return NewDocumentReader(r), nil
```
jeroiraz, line 442: minor `NewDocumentReader` may be private

embedded/document/engine.go
```diff
@@ -852,3 +829,44 @@ func (e *Engine) BulkInsertDocuments(ctx context.Context, collectionName string,
...
+func NewDocumentReader(reader sql.RowReader) DocumentReader {
+	return &documentReader{reader: reader}
+}
+
+type documentReader struct {
+	reader sql.RowReader
+}
+
+// ReadStructMessagesFromReader reads a number of messages from a reader and returns them as a slice of Struct messages.
+func (d *documentReader) Read(ctx context.Context, count int) ([]*structpb.Struct, error) {
```
jeroiraz, line 842: `ErrNoMoreDocuments` may be returned if less than specified documents can be read but still returning the documents that has been read. So it would behave the same way as when reading from an input stream with io.EOF

Alternatively the Read method be as in sql layer `Read(ctx context.Context` and return the next document

embedded/document/engine_test.go
```diff
@@ -321,9 +317,12 @@ func TestQueryDocument(t *testing.T) {
 			},
 		}
 
-		doc, err := engine.GetDocuments(context.Background(), collectionName, expressions, 1, 20)
+		reader, err := engine.GetDocuments(ctx, collectionName, expressions, 0)
```
jeroiraz, line 320: reading with count<1 may return illegal argument?

embedded/document/engine_test.go
```diff
@@ -321,9 +317,12 @@ func TestQueryDocument(t *testing.T) {
...
 		require.NoError(t, err)
-		require.Equal(t, 9, len(doc))
+		defer reader.Close()
+		docs, err := reader.Read(ctx, 20)
+		require.ErrorIs(t, err, sql.ErrNoMoreRows)
```
jeroiraz, line 324: `sql.ErrNoMoreRows` should be catch and `ErrNoMoreDocuments` may be returned instead

embedded/document/engine.go
```diff
@@ -852,3 +829,44 @@ func (e *Engine) BulkInsertDocuments(ctx context.Context, collectionName string,
...
+	var err error
+	results := make([]*structpb.Struct, 0)
+
+	for l := 0; l < count; l++ {
```
jeroiraz, line 846: reading with count<1 may return illegal argument?

embedded/document/engine_test.go
```diff
@@ -337,9 +336,12 @@ func TestQueryDocument(t *testing.T) {
 			},
 		}
 
-		doc, err := engine.GetDocuments(context.Background(), collectionName, expressions, 1, 20)
+		reader, err := engine.GetDocuments(ctx, collectionName, expressions, 0)
 		require.NoError(t, err)
-		require.Equal(t, 10, len(doc))
+		defer reader.Close()
+		docs, err := reader.Read(ctx, 20)
+		require.ErrorIs(t, err, sql.ErrNoMoreRows)
```
jeroiraz, line 343: `sql.ErrNoMoreRows` should be catch and `ErrNoMoreDocuments` may be returned instead

pkg/database/document_database.go
```diff
@@ -123,6 +123,32 @@ func (d *db) getCollection(ctx context.Context, collectionName string) (*schemav
...
+	if req.Page < 1 || req.PerPage < 1 {
+		return nil, fmt.Errorf("invalid offset or limit")
+	}
+
+	offset := (req.Page - 1) * req.PerPage
+	if offset < 0 {
+		return nil, fmt.Errorf("invalid offset")
+	}
+
+	reader, err := d.documentEngine.GetDocuments(ctx, req.Collection, queries, int64(offset))
```
jeroiraz, line 145: rare to have perPage as limit but not used in the reader

pkg/server/sessions/session.go
```diff
@@ -39,6 +52,7 @@ type Session struct {
 	creationTime     time.Time
 	lastActivityTime time.Time
 	transactions     map[string]transactions.Transaction
+	paginatedReaders map[string]*PaginatedReader // map from query names to sql.RowReader objects
```
jeroiraz, line 55: document.DocumentReader

pkg/server/sessions/session.go
```diff
@@ -21,16 +21,29 @@ import (
...
 	"github.com/codenotary/immudb/pkg/logger"
 	"github.com/codenotary/immudb/pkg/server/sessions/internal/transactions"
 	"google.golang.org/grpc/metadata"
 )
 
+var (
+	ErrPaginatedReaderNotFound = errors.New("paginated reader not found")
+)
+
+type PaginatedReader struct {
```
jeroiraz, line 40: `PaginatedDocumentReader` may be better as we may so a similar thing for sql...

pkg/server/sessions/session.go
jeroiraz, line 55: `paginatedDocumentReaders` as we might do the same for sql

embedded/document/engine_test.go
arriqaaq: done

embedded/document/engine.go
arriqaaq: ack

embedded/document/engine.go
arriqaaq: added ErrNoMoreDocuments, keeping read method as it is for now to avoid loops while reading from the reader

embedded/document/engine_test.go
arriqaaq: ack

PR #1623, opened 2023-04-18
chore: TruncateDatabase endpoint should use the same ongoing Truncator if present

arriqaaq:


pkg/server/server.go
```diff
@@ -1679,8 +1679,16 @@ func (s *ImmuServer) TruncateDatabase(ctx context.Context, req *schema.TruncateD
...
 
 	rp := time.Duration(req.RetentionPeriod) * time.Millisecond
-	truncator := truncator.NewTruncator(db, rp, 0, s.Logger)
-	err = truncator.Truncate(ctx, rp)
+
+	// check if truncator already exists for the database
+	var t *truncator.Truncator
+
+	t, err = s.getTruncatorFor(db.GetName())
+	if err == ErrTruncatorDoesNotExist {
```
jeroiraz, line 1687: it's better to compare errors using `errors.Is()` e.g. `errors.Is(err, ErrTruncatorDoesNotExist)`, as it tolerates error wrapping

jeroiraz (approved):

PR #1625, opened 2023-04-20
[wip] chore: add bulk insert api

arriqaaq:


embedded/document/engine.go
```diff
@@ -757,3 +717,138 @@ func (e *Engine) readMetadataAndValue(key []byte, atTx uint64, skipIntegrityChec
...
+				values = append(values, &sql.NullValue{})
+			}
+		}
+	}
+
+	return sql.NewRowSpec(values), nil
+
+}
+
+func (e *Engine) BulkInsertDocuments(ctx context.Context, collectionName string, docs []*structpb.Struct) (docIDs []DocumentID, txID uint64, err error) {
```
jeroiraz, line 793: I'd expect `BulkInsertDocuments` or  `BulkUpdateDocuments` to use a generalized implementation of upsert method, and define single-document insertion or update in terms of bulk operations, then there is no duplicated code

pkg/api/documentschema/documentschema.proto
```diff
@@ -297,4 +308,13 @@ service DocumentService {
     };
   }
 
+  rpc DocumentInsertMany(DocumentBulkInsertRequest) returns (DocumentBulkInsertResponse) {
```
jeroiraz, line 311: if method is named `DocumentInsertMany` it may be better to name messages `DocumentInsertManyRequest` and `DocumentInsertManyResponse`

pkg/server/documents_operations.go
```diff
@@ -165,3 +165,15 @@ func (s *ImmuServer) CollectionUpdate(ctx context.Context, req *documentschema.C
 	}
 	return resp, nil
 }
+
+func (s *ImmuServer) DocumentInsertMany(ctx context.Context, req *documentschema.DocumentBulkInsertRequest) (*documentschema.DocumentBulkInsertResponse, error) {
+	db, err := s.getDBFromCtx(ctx, "DocumentInsertMany")
+	if err != nil {
+		return nil, err
+	}
+	resp, err := db.BulkInsertDocuments(ctx, req)
```
jeroiraz, line 174: it might be simpler to use same naming in both layers? `DocumentInsertMany` in database package as well?

jeroiraz: `return db.BulkInsertDocuments(ctx, req)`

pkg/database/document_database.go
```diff
@@ -326,3 +328,16 @@ func (d *db) DocumentProof(ctx context.Context, req *schemav2.DocumentProofReque
 		},
 	}, nil
 }
+
+// BulkInsertDocuments inserts multiple documents
+func (d *db) BulkInsertDocuments(ctx context.Context, req *schemav2.DocumentBulkInsertRequest) (*schemav2.DocumentBulkInsertResponse, error) {
+	// docID, txID, err := d.documentEngine.InsertDocument(ctx, req.Collection, req.Document)
```
jeroiraz, line 334: to be implemented

PR #1640, opened 2023-04-26
pkg/api improvements

jeroiraz:


embedded/document/engine.go
```diff
@@ -24,122 +24,142 @@ import (
...
 	ErrDocumentIDMismatch     = errors.New("document id mismatch")
 	ErrNoMoreDocuments        = errors.New("no more documents")
+	ErrFieldAlreadyExists     = errors.New("field already exists")
+	ErrFieldDoesNotExist      = errors.New("field does not exist")
+	ErrReservedFieldName      = errors.New("reserved field name")
+	ErrLimitedIndexCreation   = errors.New("index creation is only supported on empty collections")
 )
 
-// Schema to ValueType mapping
+func mayTranslateError(err error) error {
```
arriqaaq, line 52: let's call it translateSQLError, or wrapSQLError or anything as such. Use of the word "may" makes it a doubtful clause when reading the code and is not explicit

arriqaaq (approved):

embedded/document/engine_test.go
```diff
@@ -45,6 +41,7 @@ func makeEngine(t *testing.T) *Engine {
 	return engine
 }
 
+/*
```
arriqaaq, line 44: In case this is commented out by mistake

jeroiraz: thanks, need to re-enable them yet

embedded/document/engine.go
arriqaaq: so translateErr is fine

embedded/document/engine.go
```diff
@@ -151,539 +52,780 @@ func NewEngine(store *store.ImmuStore, opts *sql.Options) (*Engine, error) {
...
+		nil,
+	)
+
+	r, err := e.sqlEngine.QueryPreparedStmt(ctx, sqlTx, queryStmt, nil)
+	if err != nil {
+		return 0, nil, 0, mayTranslateError(err)
+	}
+	defer r.Close()
+
+	row, err := r.Read(ctx)
```
arriqaaq, line 625: you may want to check the length of the result, if it is one or not

Makefile
```diff
@@ -153,122 +153,64 @@ test-client:
...
 	  -I$(GOPATH)/pkg/mod \
 	  -I$(GOPATH)/pkg/mod/github.com/grpc-ecosystem/grpc-gateway@$(GRPC_GATEWAY_VERSION)/third_party/googleapis \
 	  -I$(GOPATH)/pkg/mod/github.com/grpc-ecosystem/grpc-gateway@$(GRPC_GATEWAY_VERSION) \
-	  --doc_out=pkg/api/authorizationschema --doc_opt=markdown,docs.md \
-	  --plugin=protoc-gen-doc=$(PWD)/scripts/protoc-gen-doc
+	  --go_out=paths=source_relative:pkg/api/protomodel \
+	  --go-grpc_out=require_unimplemented_servers=false,paths=source_relative:pkg/api/protomodel \
+	  --grpc-gateway_out=logtostderr=true,paths=source_relative:pkg/api/protomodel \
+	  --doc_out=pkg/api/protomodel --doc_opt=markdown,docs.md \
+	  --openapiv2_out=allow_merge=true,simple_operation_ids=true,disable_service_tags=true:pkg/api/openapi \
```
arriqaaq, line 207: fails for me locally

embedded/document/engine.go
```diff
@@ -151,539 +52,780 @@ func NewEngine(store *store.ImmuStore, opts *sql.Options) (*Engine, error) {
...
+
+	sqlTx, err := e.sqlEngine.NewTx(ctx, opts)
+	if err != nil {
+		return mayTranslateError(err)
+	}
+	defer sqlTx.Cancel()
+
+	_, _, err = e.sqlEngine.ExecPreparedStmts(
 		ctx,
 		nil,
```
arriqaaq, line 320: you are not passin in this sqlTx to ExecPreparedStmts, it would create a new transaction underneath

embedded/document/engine.go
arriqaaq: I know there is a limit in the query, I thought it is upto the user to select one document based on their query and not restrict it to the first found through the query

jeroiraz: it will update one document, returning the id of the updated one. delete document having similar behaviour

PR #1641, opened 2023-04-27
chore: add DocumentFindOneAndUpdate api

arriqaaq:


embedded/document/engine.go
```diff
@@ -852,3 +858,67 @@ func (e *Engine) BulkInsertDocuments(ctx context.Context, collectionName string,
...
+
+	query := sql.NewSelectStmt(
+		[]sql.Selector{sql.NewColSelector(collectionName, DocumentBLOBField)},
+		collectionName,
+		exp,
+		nil,
+		nil,
+	)
+
+	tx, err := e.sqlEngine.NewTx(ctx, sql.DefaultTxOptions().WithExplicitClose(true))
```
arriqaaq, line 876: @jeroiraz what options should I enable for this to be a repeatable read?

embedded/document/engine.go
```diff
@@ -852,3 +884,70 @@ func (e *Engine) BulkInsertDocuments(ctx context.Context, collectionName string,
...
+
+	query := sql.NewSelectStmt(
+		[]sql.Selector{sql.NewColSelector(collectionName, DocumentBLOBField)},
+		collectionName,
+		exp,
+		nil,
+		nil,
+	)
+
+	tx, err := e.sqlEngine.NewTx(ctx, sql.DefaultTxOptions())
```
arriqaaq, line 902: @jeroiraz what options should I enable for this to be a repeatable read?

embedded/document/document.go
```diff
@@ -99,3 +100,9 @@ func (d *Document) Get(field string) interface{} {
 	}
 	return nil
 }
+
+// mergeDocumentStructs merges the given document struct with the given document struct to update
+func mergeDocumentStructs(doc *structpb.Struct, toUpdateDoc *structpb.Struct) error {
+	proto.Merge(doc, toUpdateDoc)
```
jeroiraz, line 106: taking a quick look at `proto.Merge` it seems to be that the document to be updated is the first parameter.

First though is that merging may not be the expected behaviour for an update, something to review

embedded/document/engine.go
```diff
@@ -322,34 +346,41 @@ func (e *Engine) DeleteCollection(ctx context.Context, collectionName string) er
...
 
 func (e *Engine) UpdateDocument(ctx context.Context, collectionName string, doc *structpb.Struct) (txID, rev uint64, err error) {
-	_, txID, rev, err = e.upsertDocument(ctx, collectionName, doc, false)
+	opt := newUpsertOption().withIsInsert(false)
+	_, txID, rev, err = e.upsertDocument(ctx, collectionName, doc, opt)
 	return
 }
 
-func (e *Engine) upsertDocument(ctx context.Context, collectionName string, doc *structpb.Struct, isInsert bool) (docID DocumentID, txID, rev uint64, err error) {
+func (e *Engine) upsertDocument(ctx context.Context, collectionName string, doc *structpb.Struct, opt *upsertOption) (docID DocumentID, txID, rev uint64, err error) {
```
jeroiraz, line 360: based on previous implementation of upsert in SQL layer, just `isInsert` option may be needed, so dynamic options may be too much, but it they are added it may be with a dynamic number and all the options applied e.g. for _,opt:= range options { opt(someObj) } ... but as far as I can see there is no need for it...

arriqaaq: There is also a need to pass in an existing transaction, and then it leads to too many args in the function, hence added an option struct for it

embedded/document/engine.go
arriqaaq: done

embedded/document/document.go
arriqaaq: ack, fixed it

PR #1643, opened 2023-05-01
chore: copy document catalogue when truncating db

arriqaaq:


embedded/document/engine.go
```diff
@@ -978,3 +978,8 @@ func (d *documentReader) Read(ctx context.Context, count int) ([]*structpb.Struc
 func (d *documentReader) Close() error {
 	return d.reader.Close()
 }
+
+// CopyCatalogToTx copies the current sql catalog to the ongoing transaction.
+func (e *Engine) CopyCatalogToTx(ctx context.Context, tx *store.OngoingTx) error {
+	return e.sqlEngine.CopyCatalogToTx(ctx, tx)
```
jeroiraz, line 984: isn't it covered with the sql catalog already?

note: is the sql catalog copied with the indexing up to date?

pkg/database/database.go
```diff
@@ -1656,6 +1656,17 @@ func logErr(log logger.Logger, formattedMessage string, err error) error {
...
-	// TODO: add document store support for truncation too
-	return d.sqlEngine.CopyCatalogToTx(ctx, tx)
+	// copy the sql catalog
+	err := d.sqlEngine.CopyCatalogToTx(ctx, tx)
+	if err != nil {
+		return err
+	}
+
+	// copy the document store catalog
+	err = d.documentEngine.CopyCatalogToTx(ctx, tx)
```
jeroiraz, line 1666: may not be needed... ?

embedded/document/engine.go
arriqaaq: document sql engine is instantiated with a different prefix

pkg/database/database.go
arriqaaq: document sql engine is instantiated with a different prefix

jeroiraz (approved):

PR #1676, opened 2023-05-18
chore: add lru cache for paginated readers

arriqaaq:


pkg/server/sessions/manager.go
```diff
@@ -105,7 +106,15 @@ func (sm *manager) NewSession(user *auth.User, db database.DB) (*Session, error)
 
 	sessionID := base64.URLEncoding.EncodeToString(randomBytes)
 
-	sm.sessions[sessionID] = NewSession(sessionID, user, db, sm.logger)
+	var maxActiveTxn int
+	if db != nil {
```
jeroiraz, line 110: are we currently supporting opening a session without a db?

pkg/server/sessions/session.go
```diff
@@ -53,12 +55,18 @@ type Session struct {
 	creationTime             time.Time
 	lastActivityTime         time.Time
 	transactions             map[string]transactions.Transaction
-	paginatedDocumentReaders map[string]*PaginatedDocumentReader // map from searchID to sql.RowReader objects
+	paginatedDocumentReaders *cache.LRUCache // track searchID to sql.RowReader objects
```
jeroiraz, line 58: `documentReaders *cache.LRUCache // track searchID to document.DocumentReader objects` ?

here is doesn't need to know it's a paginated reader

pkg/server/sessions/session.go
```diff
@@ -100,12 +108,18 @@ func (s *Session) removeTransaction(transactionID string) error {
 }
 
 func (s *Session) ClosePaginatedDocumentReaders() error {
```
jeroiraz, line 110: `CloseDocumentReaders` ?

pkg/server/sessions/manager.go
arriqaaq: There are tests for it, which were failing

pkg/server/sessions/session.go
```diff
@@ -53,12 +55,18 @@ type Session struct {
...
 }
 
-func NewSession(sessionID string, user *auth.User, db database.DB, log logger.Logger) *Session {
+func NewSession(sessionID string, user *auth.User, db database.DB, maxActiveSnapshots int, log logger.Logger) *Session {
+	if maxActiveSnapshots == 0 {
+		maxActiveSnapshots = store.DefaultMaxActiveTransactions
+	}
+
 	now := time.Now()
+	lruCache, _ := cache.NewLRUCache(maxActiveSnapshots)
```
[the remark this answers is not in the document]
arriqaaq: I see, I was assuming wrongly then. The problem is that it would mean there could only be maxActiveSnapshots number of read queries which can be open. Curtails it a bit, but no other option then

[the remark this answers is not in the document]
arriqaaq: cache being held by manager implies it would have to be maintained per database

pkg/server/sessions/session.go
```diff
@@ -46,28 +49,30 @@ type PaginatedDocumentReader struct {
...
-	paginatedDocumentReaders map[string]*PaginatedDocumentReader // map from searchID to sql.RowReader objects
-	log                      logger.Logger
+	mux              sync.RWMutex
+	id               string
+	user             *auth.User
+	database         database.DB
+	creationTime     time.Time
+	lastActivityTime time.Time
+	transactions     map[string]transactions.Transaction
+	documentReaders  *cache.LRUCache // track searchID to sql.RowReader objects
```
jeroiraz, line 59:
```suggestion
	documentReaders  *cache.LRUCache // track searchID to document.DocumentReader
```

pkg/server/sessions/session.go
```diff
@@ -34,6 +35,8 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
+const DefaultMaxDocumentReadersCacheSize = 5
```
jeroiraz, line 38:
```suggestion
const DefaultMaxDocumentReadersCacheSize = 1
```

jeroiraz (approved):

PR #1677, opened 2023-05-19
test(documents): add test to insert a document

marcosanchotene:


test/document_storage_tests/documents_tests/actions/insert_documents.go
```diff
@@ -0,0 +1,70 @@
...
+
+	payload := map[string]interface{}{
+		"documents": []interface{}{
+			document,
+		},
+	}
+
+	keys := reflect.ValueOf(document).MapKeys()
+	field := keys[1].String()
+	value := document[keys[1].String()]
```
marcosQuesada, line 38: You can get Keys and values by simple iteration, without using reflect as far as I see

marcosQuesada: Not a problem, but I would say that is not the usual way,  Reflection package is useful when access to the data is restricted or in non strict types...

marcosanchotene: Ok. I'll change it then. Thanks.

tomekkolo: there is no guarantee on order (also in reflect as you have map). Just pass to the function with which field/value you want to have search later.

test/document_storage_tests/documents_tests/actions/insert_documents.go
```diff
@@ -0,0 +1,70 @@
...
+
+func insertDocuments(expect *httpexpect.Expect, sessionID string, collectionName string, payload map[string]interface{}, field string, value interface{}) *httpexpect.Object {
+	expect.POST(fmt.Sprintf("/collection/%s/documents", collectionName)).
+		WithHeader("grpc-metadata-sessionid", sessionID).
+		WithJSON(payload).
+		Expect().
+		Status(http.StatusOK).JSON().Object().NotEmpty().
+		Keys().ContainsOnly("transactionId", "documentIds")
+
+	searchPayload := map[string]interface{}{
```
tomekkolo, line 51: much cleaner is just to have searchPayload as type struct and then marshal it to map[string]interface{} with json.Marshal.

test/document_storage_tests/documents_tests/insert_documents_test.go
```diff
@@ -0,0 +1,61 @@
...
+	collection *httpexpect.Object
+}
+
+func (s *InsertDocumentsTestSuite) SetupTest() {
+	s.expect, s.sessionID = actions.OpenSession(s.T())
+	s.collection = actions.CreateCollectionWithNameAndMultipleFields(s.expect, s.sessionID, uuid.New().String())
+}
+
+func (s *InsertDocumentsTestSuite) TestInsertOneDocumentWithMultipleFields() {
+	document := map[string]interface{}{
```
tomekkolo, line 41: this is not a real document, as eventually what you have is map[string]string (all the values will be considered as string). Make it a proper test structure and then json.Marshal

test/document_storage_tests/documents_tests/insert_documents_test.go
```diff
@@ -0,0 +1,61 @@
...
+		"first_name": "Bezalel",
+		"last_name":  "Simmel",
+		"gender":     "F",
+		"hire_date":  "1985-11-21",
+	}
+
+	documentFound := actions.InsertOneDocumentWithMultipleFields(s.expect, s.sessionID, s.collection, document)
+
+	documentFound.Keys().ContainsOnly("_id", "birth_date", "first_name", "last_name", "gender", "hire_date")
+	documentFound.Value("birth_date").IsEqual(document["birth_date"])
```
tomekkolo, line 52: you should expect also concrete types, now there are just strings

test/document_storage_tests/documents_tests/actions/insert_documents.go
marcosanchotene: How? I'll need the same variable to unmarshal it, isn't it?

tomekkolo: No, you can always unmarshal to map[string]interface. If you have structure, you can unmarshal to concrete type

Issue #1094, opened 2022-01-12
Support AWS role authentication


SimoneLazzaris:
**What would you like to be added or enhanced**
Support AWS role authentication

**Why is this needed**

The recommended way to interact with S3 is to assign a role to the EC2 instance hosting the application. Ephemeral credentials are then obtained using a well-known URL. At the moment this is not supported by immudb

**Additional context**

From AWS documentation

>When the application runs, it obtains temporary security credentials from Amazon EC2 instance metadata, as described in Retrieving Security Credentials from Instance Metadata. These are temporary security credentials that represent the role and are valid for a limited period of time.
> With some AWS SDKs, the developer can use a provider that manages the temporary security credentials transparently. (The documentation for individual AWS SDKs describes the features supported by that SDK for managing credentials.)
> Alternatively, the application can get the temporary credentials directly from the instance metadata of the EC2 instance. Credentials and related values are available from the iam/security-credentials/role-name category (in this case, iam/security-credentials/Get-pics) of the metadata. If the application gets the credentials from the instance metadata, it can cache the credentials.

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials

angelAtSequent:
Any chance to implement this feature?

ostafen:
Hey, @angelAtSequent, we already took this issue in consideration and plan to implement this soon.
Stay tuned!

vchaindz:
Closing — AWS role authentication is fully supported.

- **EC2 instance IAM role**: IMDSv2 token-based flow in `embedded/remotestorage/s3/s3.go` `requestCredentials()`. Enable via `--s3-role-enabled=true`, optionally specify `--s3-role=<name>`; otherwise discovered from the instance profile ARN. Credentials auto-refresh on a ticker.
- **ECS / Fargate container credentials**: added via PR #2076 (merged 2025-10). Enable via `--s3-use-fargate-credentials=true`; the driver reads `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` and pulls creds from `http://169.254.170.2`.

If you have a scenario that neither of these covers (e.g. IRSA on EKS with a web-identity token file), please open a new issue with the specific auth mode needed.

PR #1697, opened 2023-06-12
Feat/s3 role based auth

NikitaSmall:
This PR makes use of AWS roles to provide credentials
When the role is provided, the storage gets the credentials based on the security metadata. 
This change also includes the refresh of the credentials.

The original idea behind this PR is taken from here: https://github.com/codenotary/immudb/issues/1094


SergeyBoiko (approved):

embedded/remotestorage/s3/s3.go
```diff
@@ -118,7 +128,15 @@ func Open(
 				return http.ErrUseLastResponse
 			},
 		},
-	}, nil
+		s3CredentialsRefreshTicker: time.NewTicker(time.Minute),
```
SimoneLazzaris, line 131: It would be better to have this ticker created only if needed (i.e., when a s3 role is assigned)

embedded/remotestorage/s3/s3.go
```diff
@@ -691,4 +709,83 @@ func (s *Storage) scanObjectNames(ctx context.Context, prefix string, limit int)
...
+	var err error
+	s.accessKeyID, s.secretKey, err = s.requestCredentials()
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		select {
+		case _ = <-s.s3CredentialsRefreshTicker.C:
+			s.accessKeyID, s.secretKey, _ = s.requestCredentials()
```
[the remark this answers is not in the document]
SimoneLazzaris: The error should also be logged, otherwise it would be impossible for the sysadmin to understand what is going on.

NikitaSmall:
@SimoneLazzaris, thank you for your comments! Those are taken into account and fixed.

SimoneLazzaris (approved):
LGTM

NikitaSmall:
@SimoneLazzaris , @jeroiraz , could you please help me understand the pipeline: I saw some recent PRs in ImmuDB have some failed checks within workflows and those PRs were merged. As for this one, do I need to do anything else for it to be merged?

jeroiraz:
> @SimoneLazzaris , @jeroiraz , could you please help me understand the pipeline: I saw some recent PRs in ImmuDB have some failed checks within workflows and those PRs were merged. As for this one, do I need to do anything else for it to be merged?

thanks @NikitaSmall, we'll further test it early next week, if everything goes well it will be merged

README.md
```diff
@@ -172,6 +172,29 @@ export IMMUDB_S3_ENDPOINT="https://${IMMUDB_S3_BUCKET_NAME}.s3.${IMMUDB_S3_LOCAT
...
+export IMMUDB_S3_ROLE_ENABLED=true
+export IMMUDB_S3_BUCKET_NAME=<BUCKET NAME>
+export IMMUDB_S3_LOCATION=<AWS S3 REGION>
+export IMMUDB_S3_PATH_PREFIX=testing-001
+export IMMUDB_S3_ENDPOINT="https://${IMMUDB_S3_BUCKET_NAME}.s3.${IMMUDB_S3_LOCATION}.amazonaws.com"
+
+./immudb
+```
+
+Optionally, you can specify the exact role ImmuDB should be using with:
```
jeroiraz, line 190: quite minor, `immudb` all lower case, just to uniformity

embedded/remotestorage/s3/s3.go
```diff
@@ -73,12 +80,20 @@ var (
 	ErrInvalidResponseSubPathUnescape      = fmt.Errorf("%w: error un-escaping object name", ErrInvalidResponse)
 
 	ErrTooManyRedirects = errors.New("too many redirects")
+
+	arnRoleRegex = regexp.MustCompile(`arn:.*\/(.*)`)
```
jeroiraz, line 84: would this regex match all possible role definitions?

embedded/remotestorage/s3/s3.go
```diff
@@ -73,12 +80,20 @@ var (
...
 )
 
-const maxRedirects = 5
+const (
+	awsCredsRefreshPeriod  = time.Minute
+	awsInstanceMetadataURL = "http://169.254.169.254"
```
jeroiraz, line 89: this type of constant definitions should not be part of the code  `awsInstanceMetadataURL = "http://169.254.169.254"`

embedded/remotestorage/s3/s3.go
```diff
@@ -41,17 +43,22 @@ import (
...
+	endpoint                   string
+	S3RoleEnabled              bool
+	s3Role                     string
+	accessKeyID                string
+	secretKey                  string
+	bucket                     string
+	prefix                     string
+	location                   string
+	httpClient                 *http.Client
+	s3CredentialsRefreshTicker *time.Ticker
```
jeroiraz, line 55: `s3CredentialsRefreshTicker *time.Ticker` seems to be used in the scope of the method, it should not be part of the struct...

embedded/remotestorage/s3/s3.go
```diff
@@ -691,4 +715,127 @@ func (s *Storage) scanObjectNames(ctx context.Context, prefix string, limit int)
 	return entries, subPaths, nil
 }
 
+func (s *Storage) getRoleCredentials() error {
+	if !s.S3RoleEnabled {
+		return nil
+	}
+
+	if s.accessKeyID != "" || s.secretKey != "" {
```
jeroiraz, line 723: this validation could even be done before constructing the main remote storage struct, e.g. Validate method in options

embedded/remotestorage/s3/s3.go
jeroiraz: ah, it's for a long-running goroutine,  probably it's preferable to have the reference to the goroutine directly.

The goroutine should not hardly stop if it fails to fetch to renew the credentials but it may log the error and keep trying

embedded/remotestorage/s3/s3.go
NikitaSmall: This one is quite wide, so I believe that this regex will catch any role.

embedded/remotestorage/s3/s3.go
NikitaSmall: I did not find the `Validate` method for the regular `Options` or `RemoteStorageOptions`. Should I create one?
Additionally, this is the check that would only work in that niche case. Check beforehand might be a bit of a help but not by much.

jeroiraz: it could at least be early validated at `createRemoteStorageInstance` for instance. It may be simpler to validate options upfront and in one place. Also raising the error right from the start before giving any chance to do any further work that could end-up in an error being thrown while using it.

Ideally there could be a validate options method, we already have that level of uniformity under `embedded` package, but not in the server, at least not yet. But it does not need to be done right now. Having the validations upfront and in a single place would be enough

embedded/remotestorage/s3/s3.go
```diff
@@ -106,19 +119,30 @@ func Open(
...
+		bucket:        bucket,
+		location:      location,
+		prefix:        prefix,
 		httpClient: &http.Client{
 			CheckRedirect: func(req *http.Request, via []*http.Request) error {
 				return http.ErrUseLastResponse
 			},
 		},
-	}, nil
+		awsInstanceMetadataURL: "http://169.254.169.254",
```
jeroiraz, line 136: I wasn't clear with my previous comment, urls like this should not be part of the code, if an url different than the s3 host is needed then there should be another flag for that

embedded/remotestorage/s3/s3.go
```diff
@@ -691,4 +715,126 @@ func (s *Storage) scanObjectNames(ctx context.Context, prefix string, limit int)
...
+func (s *Storage) requestCredentials() (string, string, error) {
+	tokenReq, err := http.NewRequest("PUT", fmt.Sprintf("%s%s",
+		s.awsInstanceMetadataURL,
+		"/latest/api/token",
+	), nil)
+	if err != nil {
+		return "", "", errors.New("cannot form metadata token request")
+	}
+
+	tokenReq.Header.Set("X-aws-ec2-metadata-token-ttl-seconds", "21600")
```
jeroiraz, line 758: is this header also useful in most S3 storages? I mean, when not using aws

jeroiraz: same questions for `/latest/api/token` and `/latest/meta-data/iam/info`

NikitaSmall: This is specific to AWS and S3, yes, but this is the flow that is specifically applicable to AWS.
I checked Minio docs and references, I didn't find any way to use role-based auth. Essentially, it means that role-based auth and the whole `S3RoleEnabled` is AWS specific.

Should I refactor this? I could spend some time and make it _in theory_ portable while the reality is that there are no other applications for this code but AWS S3.

NikitaSmall:
@jeroiraz , thank you for your comments! I addressed most of them and responded to others.

embedded/remotestorage/s3/s3.go
SimoneLazzaris: actually, the 169.254.169.254 is a fixed ip address used by aws instances to get their runtime parameters and/or as api endpoint. So I think it makes sense to hardwire it in the code

NikitaSmall: @SimoneLazzaris you are right. As for now, I've made this into a default value for a parameter (flag or env var), so should be used anyway, just with a bit more flexibility than initially required.

Would you please push this a bit further to review or possibly merge this one?

NikitaSmall:
Hello!
Is there anything that can be done to push this PR forward? Or should we just forgo the change?

cmd/immudb/command/init.go
```diff
@@ -72,12 +72,15 @@ func (cl *Commandline) setupFlags(cmd *cobra.Command, options *server.Options) {
...
 	cmd.Flags().Bool("s3-storage", false, "enable or disable s3 storage")
+	cmd.Flags().Bool("s3-role-enabled", false, "enable role-based authentication for s3 storage")
 	cmd.Flags().String("s3-endpoint", "", "s3 endpoint")
+	cmd.Flags().String("s3-role", "", "role name for role-based authentication attempt for s3 storage")
 	cmd.Flags().String("s3-access-key-id", "", "s3 access key id")
 	cmd.Flags().String("s3-secret-key", "", "s3 secret access key")
 	cmd.Flags().String("s3-bucket-name", "", "s3 bucket name")
 	cmd.Flags().String("s3-location", "", "s3 location (region)")
 	cmd.Flags().String("s3-path-prefix", "", "s3 path prefix (multiple immudb instances can share the same bucket if they have different prefixes)")
+	cmd.Flags().String("s3-instance-metadata-url", "http://169.254.169.254", "s3 instance metadata url")
```
jeroiraz, line 83: no fixed url nor ip should be included

cmd/immudb/command/init.go
```diff
@@ -135,11 +138,14 @@ func setupDefaults(options *server.Options) {
...
 	viper.SetDefault("s3-storage", false)
 	viper.SetDefault("s3-endpoint", "")
+	viper.SetDefault("s3-role-enabled", false)
+	viper.SetDefault("s3-role", "")
 	viper.SetDefault("s3-access-key-id", "")
 	viper.SetDefault("s3-secret-key", "")
 	viper.SetDefault("s3-bucket-name", "")
 	viper.SetDefault("s3-location", "")
 	viper.SetDefault("s3-path-prefix", "")
+	viper.SetDefault("s3-instance-metadata-url", "http://169.254.169.254")
```
jeroiraz, line 148: no fixed url nor ip should be included

jeroiraz (approved):

jeroiraz:
Changes has been merged with PR #1876

PR #1712, opened 2023-06-19
feat: add flag for using external id as a main one

NikitaSmall:
when the remote storage is configured correctly, the identifier is taken from it IF POSSIBLE. That means that if the id is not present in the remote storage, Immudb works as before (creates a new id and then saves it in the remote).

The idea of this pull request originated from a last week discussion:

> we have troubles with IDENTIFIER FILES when we restart the tasks (pods) with configured S3 backend, we see remote storage does not match local files error. Looking in the code shows that the error is caused by the difference in the UUIDs stored in `immudb.identifier` files locally vs remotely. But when ImmuDB first initializes on the fresh disk (because it's an ephemeral disk attached to a task) it always creates a new `immudb.identifier`  file


pkg/server/remote_storage.go
```diff
@@ -80,7 +80,7 @@ func (s *ImmuServer) initializeRemoteStorage(storage remotestorage.Storage) erro
 			return err
 		}
 
-		if !fileExists(localIdentifierFile) {
+		if !fileExists(localIdentifierFile) || s.Options.RemoteStorageOptions.S3ExternalIdentifier {
```
jeroiraz, line 83: it may be problematic to overwrite an existent local identifier

it may be better to raise an error if the local identifier (if any) differs from the remote one

I'd suggest to extend the method `getOrSetUUID` in server.go file before initializing the remote storage with a parameter `createIfNotExists`, and set its value in the call with the value of  `!s.Options.RemoteStorageOptions.S3ExternalIdentifier` ...

README.md
```diff
@@ -172,6 +172,14 @@ export IMMUDB_S3_ENDPOINT="https://${IMMUDB_S3_BUCKET_NAME}.s3.${IMMUDB_S3_LOCAT
 ./immudb
 ```
 
+When working with the external storage, you can enable the option for the remote storage
+to be the primary source of identifier. This way, if ImmuDB is run using ephemeral disks,
```
jeroiraz, line 176: ultra minor, we prefer just lower-case name for immudb e.g. `This way, if immudb is run ...`

jeroiraz:
thanks a lot @NikitaSmall, overall changes look good, made a request for changes with regards to the local identifier handling

NikitaSmall:
@jeroiraz , thank you for your comments. I addressed them.

pkg/server/uuid.go
```diff
@@ -48,11 +48,16 @@ func NewUUIDContext(id xid.ID) uuidContext {
 	return uuidContext{id}
 }
 
-// getOrSetUUID either reads the identifier file or creates it if it doesn't exist.
+// getOrSetUUID either reads the identifier file or creates it if it doesn't exist
+// unless useExternalIdentifier is set to true, in which case the local identifier file is ignored.
```
jeroiraz, line 52: I think the local identifier has higher precedence. And if present, it should match the one fetched from remote storage.
If the local identifier is not present, it then could use the remote one if instructed.

NikitaSmall: makes perfect sense, now we use the local id if it's found even if the flag is on.

pkg/server/remote_storage.go
```diff
@@ -66,32 +75,40 @@ func (s *ImmuServer) initializeRemoteStorage(storage remotestorage.Storage) erro
...
 		if !fileExists(localIdentifierFile) {
-			err := ioutil.WriteFile(localIdentifierFile, remoteId, os.ModePerm)
+			err := ioutil.WriteFile(localIdentifierFile, remoteID, os.ModePerm)
 			if err != nil {
 				return err
 			}
+			s.UUID, err = xid.FromBytes(remoteID)
 		} else {
-			localId, err := ioutil.ReadFile(localIdentifierFile)
+			if s.Options.RemoteStorageOptions.S3ExternalIdentifier {
```
jeroiraz, line 102: I think it's already validated at line 78

NikitaSmall:
@jeroiraz thank you for your comments, I corrected the code to align with the vision.

cmd/immudb/command/init.go
```diff
@@ -78,6 +78,7 @@ func (cl *Commandline) setupFlags(cmd *cobra.Command, options *server.Options) {
 	cmd.Flags().String("s3-bucket-name", "", "s3 bucket name")
 	cmd.Flags().String("s3-location", "", "s3 location (region)")
 	cmd.Flags().String("s3-path-prefix", "", "s3 path prefix (multiple immudb instances can share the same bucket if they have different prefixes)")
+	cmd.Flags().Bool("s3-external-identifier", false, "use external identifier over the local")
```
jeroiraz, line 81:
```suggestion
	cmd.Flags().Bool("s3-external-identifier", false, "use the remote identifier if there is no local identifier")
```

jeroiraz:
LGTM

Just made a minor suggestion in the new flag description @NikitaSmall

pkg/server/remote_storage_test.go
```diff
@@ -229,6 +229,83 @@ func TestInitializeRemoteStorageDownloadIdentifier(t *testing.T) {
...
+
+	opts := DefaultOptions().WithDir(dir)
+	opts.RemoteStorageOptions.S3ExternalIdentifier = true
+
+	s := DefaultServer()
+
+	s.WithOptions(opts)
+
+	err := s.Initialize()
+	require.Error(t, ErrNoStorageForIdentifier, err)
```
jeroiraz:
```suggestion
	require.ErrorIs(t, err, ErrNoStorageForIdentifier)
```

pkg/server/remote_storage_test.go
```diff
@@ -229,6 +229,83 @@ func TestInitializeRemoteStorageDownloadIdentifier(t *testing.T) {
...
+
+	s := DefaultServer()
+
+	s.WithOptions(opts)
+
+	m := memory.Open()
+	s.remoteStorage = m
+
+	err := s.initializeRemoteStorage(m)
+	require.Error(t, ErrNoRemoteIdentifier, err)
```
jeroiraz:
```suggestion
	require.ErrorIs(t, err, ErrNoStorageForIdentifier)
```

pkg/server/remote_storage_test.go
NikitaSmall: This is already addressed, I believe

pkg/server/remote_storage_test.go
NikitaSmall: same here

NikitaSmall:
Thanks!
The suggestions are accepted where possible (the rest are done on the outdated code base).

jeroiraz:
> Thanks! The suggestions are accepted where possible (the rest are done on the outdated code base).

yes, just realized

pkg/server/remote_storage_test.go
```diff
@@ -229,6 +229,63 @@ func TestInitializeRemoteStorageDownloadIdentifier(t *testing.T) {
...
+
+	opts := DefaultOptions().WithDir(dir)
+	opts.RemoteStorageOptions.S3ExternalIdentifier = true
+
+	s := DefaultServer()
+
+	s.WithOptions(opts)
+
+	err := s.Initialize()
+	require.Error(t, err, ErrNoStorageForIdentifier)
```
jeroiraz, line 243:
```suggestion
	require.ErrorIs(t, err, ErrNoStorageForIdentifier)
```

Using `ErrorIs` to validate `err` is that specific type of error

pkg/server/remote_storage_test.go
```diff
@@ -229,6 +229,63 @@ func TestInitializeRemoteStorageDownloadIdentifier(t *testing.T) {
...
+
+	s := DefaultServer()
+
+	s.WithOptions(opts)
+
+	m := memory.Open()
+	s.remoteStorage = m
+
+	err := s.initializeRemoteStorage(m)
+	require.Error(t, err, ErrNoRemoteIdentifier)
```
jeroiraz, line 260:
```suggestion
	require.ErrorIs(t, err, ErrNoStorageForIdentifier)
```

Using `ErrorIs` to validate `err` is that specific type of error

PR #1732, opened 2023-07-05
chore: use copy instead of a loop

testwill:


jeroiraz:
many thanks @testwill

would you mind doing the same improvement in `colsBySelector` method as well?

jeroiraz:
would you please also sign your commit?

PR #1733, opened 2023-07-05
chore: unnecessary use of fmt.Sprintf

testwill:


pkg/pgsql/server/pgsql_integration_test.go
```diff
@@ -497,7 +497,7 @@ func TestPgsqlServer_VersionStatement(t *testing.T) {
 	require.NoError(t, err)
 
 	var version string
-	err = db.QueryRow(fmt.Sprintf("SELECT version()")).Scan(&version)
+	err = db.QueryRow(("SELECT version()").Scan(&version)
```
jeroiraz, line 500:
```suggestion
	err = db.QueryRow("SELECT version()").Scan(&version)
```

jeroiraz:
thanks, would you please also sign your commit?

PR #1761, opened 2023-08-02
test: use `t.Setenv` to set env vars in tests

Juneezee:
Starting from Go 1.17, we can use `t.Setenv` to set environment variable in test. The environment variable is automatically restored to its original value when the test and all its subtests complete. This ensures that each test does not start with leftover environment variables from previous completed tests.

Reference: https://pkg.go.dev/testing#T.Setenv

```go
func TestFoo(t *testing.T) {
	// before
	os.Setenv(key, "new value")
	defer os.Unsetenv(key)
	
	// after
	t.Setenv(key, "new value")
}
```

### DIFF
--- a/embedded/logger/simple_test.go
+++ b/embedded/logger/simple_test.go
@@ -20,15 +20,14 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
-	"os"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 )
 
 func TestSimpleLogger(t *testing.T) {
-	os.Setenv("LOG_LEVEL", "error")
-	defer os.Unsetenv("LOG_LEVEL")
+	t.Setenv("LOG_LEVEL", "error")
+
 	name := "test-simple-logger"
 	outputWriter := bytes.NewBufferString("")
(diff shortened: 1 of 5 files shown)


embedded/logger/simple_test.go
```diff
@@ -63,13 +62,24 @@ func TestSimpleLogger(t *testing.T) {
 func TestLogLevelFromEnvironment(t *testing.T) {
```
jeroiraz, line 62: this test seems to be failing

Juneezee: Ahh...that's because `make test` sets `LOG_LEVEL=error` :sweat_smile: And prior to this PR we have been doing `defer os.Unsetenv("LOG_LEVEL")` which unsets the `LOG_LEVEL=error` when the tests finished.

https://github.com/codenotary/immudb/blob/0ede732721dec111736051fbb0cc1e996e41b8a5/Makefile#L135-L138

image

Juneezee: Fixed in 1da96b07c2b49786f15477eabf1c9d87dc450bb5

jeroiraz (approved):

jeroiraz:
thanks @Juneezee !

PR #1763, opened 2023-08-04
fix: insertion ts for key-values should not be equal to the current root ts

arriqaaq:
This PR addresses an issue related to the insertion timestamp (ts) of key-value pairs within the tbtree. The problem is that the insertion ts for key-values was can be equal to the current root ts, which can cause unexpected behavior and inconsistencies in the data records.

I have added a test case to validate the correctness of this fix and to prevent any potential regressions in the future.

### DIFF
--- a/embedded/tbtree/tbtree.go
+++ b/embedded/tbtree/tbtree.go
@@ -1633,7 +1633,7 @@ func (t *TBtree) bulkInsert(kvts []*KVT) error {
 		if t == 0 {
 			// zero-valued timestamps are associated with current time plus one
 			t = currTs + 1
-		} else if kvt.T < currTs {
+		} else if kvt.T <= currTs { // insertion with a timestamp older or equal to the current timestamp should not be allowed
 			return fmt.Errorf("%w: specific timestamp is older than root's current timestamp", ErrIllegalArguments)
 		}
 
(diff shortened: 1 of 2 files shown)


jeroiraz:
thanks @arriqaaq.
the idea of allowing multiple insertions in a non-decreasing time was to not be restrictive, it shouldn't result in an inconsistent state as far as I though by then, were you able to identify a potential inconsistency?

arriqaaq:
Hey @jeroiraz 

I was just going through the code, and figured out that two separate insertions could happen on the same timestamp, hence overwriting the existing timestamp of the root. From what I read from the code, I understood that all the insert operations result in a non-decreasing time for the tree. Though currently I couldn't see a flaw in the code, but if there is ever a bug during indexing, it could be an issue. WDYT?

jeroiraz:
> Hey @jeroiraz
> 
> I was just going through the code, and figured out that two separate insertions could happen on the same timestamp, hence overwriting the existing timestamp of the root. From what I read from the code, I understood that all the insert operations result in a non-decreasing time for the tree. Though currently I couldn't see a flaw in the code, but if there is ever a bug during indexing, it could be an issue. WDYT?

the btree component was designed to be isolated from the core store itself, so it could potentially be used for a different use case or context. The increasing constraint could be an optional setting...

arriqaaq:
But if you feel it is not required as such, feel free to close this PR. In context of the append-only implementation, it was confusing hence the PR.

PR #1825, opened 2023-10-12
fix: standard syntax for drop index

SimoneLazzaris:


embedded/sql/sql_grammar.y
```diff
@@ -230,6 +230,11 @@ ddlstmt:
     {
         $$ = &DropIndexStmt{table: $4, cols: $6}
     }
+|
+    DROP INDEX IDENTIFIER DOT ids
```
jeroiraz, line 234: what about restricting to a single column with this syntax? 

DROP INDEX IDENTIFIER DOT IDENTIFIER

embedded/sql/sql_grammar.y
```diff
@@ -230,6 +230,11 @@ ddlstmt:
...
+    {
+        $$ = &DropIndexStmt{table: $3, cols: $5}
```
jeroiraz, line 236: $$ = &DropIndexStmt{table: $4, cols: []string{$6}}

Issue #1770, opened 2023-08-13
Support basic ASCII characters in database names like underscore `_`


edulix:
**What would you like to be added or enhanced**
Database names could be able to contain basic ASCII characters like underscore `_` - perhaps also some other chars to be defined.

**Why is this needed**
Right now I've found no easy way to visually differentiate words in database names. Adding underscore will help. Table names and columns already support underscore.

**Additional context**
When creating a database name with underscore, you get `punctuation marks and symbols are not allowed in database name`.

vchaindz:
Underscore `_` is already supported in database names on current `master` (and has been for some time). The validator in `pkg/server/server.go` (`isValidDBName`) explicitly allows it:

```go
case ch == '_':
```

Verified locally: `hello_world`, `abc_def_123` all pass validation.

Closing as already resolved. Please reopen if you can reproduce against v1.11.0.

PR #1863, opened 2023-11-07
feat(pkg/server): add support of underscore in db name

martinjirku:
This should solve the #1770. This change introduces the support of an underscore in the database name.  

- `pkg/server/server.go` - adds an additional case to checking the character validation.
- `pkg/integration/database_creation_test.go#159` - a new test for db creation with a name containing the `_`
- `pkg/server/server_test.go#1892` - unit test which tests the underscore in the isValidDBName

Signed-off-by: Martin Jirku <martin@jirku.sk>


jeroiraz (approved):
Thanks @martinjirku !

Issue #1755, opened 2023-07-28
Custom id in collection


SimoneLazzaris:
Now document collection works with an index field which has to be autogenerated. 
In other workds, idFieldName has to always be autogenerated.

As an enhancement, we should support also custom fields to work as primary index.

PR #1884, opened 2023-12-04
feat: add support for custom document id field

prog8:
With this iteration we keep consistency with current format of document identifiers. Auto-generated identifier is a hexadecimal string representation of binary identifier. In case of custom ID field we keep exactly the same format. This means a user has to make sure he sends document with custom ID field that follows this pattern. With the next iteration we can adjust the identifier so it doesn't require ID to be hex string.


embedded/document/engine.go
```diff
@@ -716,8 +716,8 @@ func (e *Engine) upsertDocuments(ctx context.Context, sqlTx *sql.SQLTx, collecti
 
 		provisionedDocID, docIDProvisioned := doc.Fields[docIDFieldName]
 		if docIDProvisioned {
-			if isInsert {
-				return 0, nil, fmt.Errorf("%w: field (%s) should NOT be specified when inserting a document", ErrIllegalArguments, docIDFieldName)
+			if isInsert && docIDFieldName == DefaultDocumentIDField {
```
tomekkolo, line 719: if I read correctly, if inserted document does not have custom id, then immudb will autogenerate one. I would block that, as it is doing something by the user behind the scenes, so something that I would not expect when the user explicitly specifies that document id is provided externally. I.e. even if ids are just hex encoded strings at the end, they might have more meaning from the user perspective.

embedded/document/engine_test.go
```diff
@@ -1283,6 +1283,101 @@ func TestDocumentUpdate(t *testing.T) {
 		require.Len(t, revisions, 1)
 	})
 }
+func TestCustomDocumentID(t *testing.T) {
```
tomekkolo, line 1286: add a negative tests, inserting document with the same id twice. Case of collisions of doc ids, if they are handled properly.

embedded/document/engine.go
prog8: Indeed. In case of missing custom ID provided by the user it will be auto-generated. This is similar strategy for example Elasticsearch uses. If user doesn't specify an identifier it is automatically generated. Each approach has pros and cons of course. In https://github.com/codenotary/immudb/issues/1755 I mentioned that maybe it's not worth giving a user flexibility to provide document DB with custom ID field just force the name of ID field but let user customize the value. In such a case it's even more important to have auto-generated `_id` if not provided or use the one from the user if available.

embedded/document/engine_test.go
prog8: good point

PR #1905, opened 2023-12-20
license update to BSL

SimoneLazzaris:
- update license file
- license update inside files

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,201 +1,89 @@
-                                 Apache License
-                           Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+License text copyright (c) 2020 MariaDB Corporation Ab, All Rights Reserved.
+"Business Source License" is a trademark of MariaDB Corporation Ab.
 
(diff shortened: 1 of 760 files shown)


tomekkolo (approved):

mitar:
Has this change been discussed with the community? Because I do not see any discussion in issues or GitHub discussions about this change? No announcement, nothing? Nothing in blog either. To make the project non-open source in the middle of holidays I think is really a shitty move. Such moves have been badly executed in the past, but this one probably tops all of them.

Also, I am not sure if such license change is really warranted for immudb? Do you feel like any of the large cloud providers is taking you on and profit from immudb offering, competing with you?

I was using immudb as embedded database in my Go project, but with this change I cannot anymore. It would mean my whole project would have an impossible licensing situation. I think BSL maybe works when database is used as a service, but for embedded database, what is the license of the resulting project?

If you had made a wider discussion on this, such issues with this licensing move could be discovered earlier. Now, I am not sure what to do? :-(

mitar:
Also, [README still states Apache license](https://github.com/codenotary/immudb#contributing).

PR #1917, opened 2024-01-28
chore(pkg/database): use atomic for waitingCount in instrumentedRWMutex

arturmelanchyk:
This change minimizes lock contention when working with `instrumentedRWMutex`
According to the attached benchmark `Lock` \ `Unlock` operations as well as `RLock` \ `RUnlock` gain ~50% performance boost while `State`'s performance remains on the same level

### DIFF
--- a/pkg/database/instrumented_rwmutex.go
+++ b/pkg/database/instrumented_rwmutex.go
@@ -18,6 +18,7 @@ package database
 
 import (
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
@@ -25,7 +26,7 @@ type instrumentedRWMutex struct {
 	rwmutex sync.RWMutex
 
 	trwmutex      sync.RWMutex
-	waitingCount  int
+	waitingCount  int64
 	lastReleaseAt time.Time
 }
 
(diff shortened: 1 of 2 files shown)


pkg/database/instrumented_rwmutex.go
```diff
@@ -58,15 +54,10 @@ func (imux *instrumentedRWMutex) Unlock() {
 }
 
 func (imux *instrumentedRWMutex) RLock() {
-	imux.trwmutex.Lock()
-	imux.waitingCount++
-	imux.trwmutex.Unlock()
+	atomic.AddInt64(&imux.waitingCount, 1)
+	defer atomic.AddInt64(&imux.waitingCount, -1)
```
tomekkolo, line 58: I do not think that it makes sense to use defer here, just call after rlock.

arturmelanchyk: fixed

tomekkolo (approved):

tomekkolo:
@arturmelanchyk , please sign commits with GPG

arturmelanchyk:
Just a kind reminder @tomekkolo @SimoneLazzaris , branch has been rebased onto latest master and all commits signed with GPG

tomekkolo:
> Just a kind reminder @tomekkolo @SimoneLazzaris , branch has been rebased onto latest master and all commits signed with GPG

Looked at it once more, why using 64 bit ? If you follow the path where it is used (DatabaseHealth), it is anyway cut to 32bits at the end. And generally, no need for such big counter here. Also not sure about real gain, this endpoint is not called so much.

arturmelanchyk:
@tomekkolo 
1. As to the real gain, database heavily rely on this `instrumentedRWMutex` type, its used everywhere instead of regular `sync.Mutex` and `sync.RWMutex`
https://github.com/codenotary/immudb/blob/c7265c67d1a9d52940822332da82e0d543a86547/pkg/database/database.go#L147-L153

see list of methods that use it:

Screenshot 2024-03-23 at 12 14 42

as a result, even tiny improvement in this type will lead to noticeable speed boost

2. As to the size, instance of this type is only created once, so comparing 32bit to 64 bit we would only save 32bits once. Also take into consideration that in previous version this type was `int` which would anyway compile to 64 bits on  most of the modern computers, so technically we do not add anything.

PR #1947, opened 2024-04-08
chore: set repository_owner dynamically in the push workflow.

seuros:
The GitHub workflow push.yml was updated to use the repository owner's Docker images instead of fixed ones. 

This change allows for more flexibility and control when using Docker images, and ensures that the correct images are used based on the repository owner.

This allow forked repos to not push in the immudb organization by default


ostafen (approved):

seuros:
@SimoneLazzaris , i was going to use secrets.REGISTRY_USER, but i endup using a environment variable.

btw REGISTRY_USER should be a variable not a secret.

SimoneLazzaris (approved):
Looks good. Thanks for your effort

PR #1950, opened 2024-04-11
Improve output of immuadmin status command

ostafen:
This PR enriches the output of the `immuadmin status` command as follows (previously only status was returned):
```golang
foo@bar:~$ immuadmin status
Status: Ok - server is reachable and responding to queries
Version: 1.9DOM.2
Start time: 11 Apr 24 15:12 CEST (up 7m3s)
Total databases: 1 (5.3 KB)
```


Razikus (approved):

pkg/api/schema/schema.proto
```diff
@@ -609,11 +609,14 @@ message ServerInfoResponse {
   // Unix timestamp (seconds) indicating when the server process has been started.
   int64 startedAt = 2;
 
+  // Total number of transactions across all databases.
+  int64 numTransactions = 3;
+
   // Total number of databases present.
-  int32 numDatabases = 3;
+  int32 numDatabases = 4;
```
Razikus, line 616: this is potential backward compability break
renumbering field will break existing clients or sdks, because these were working still on old proto

Razikus (approved):

pkg/api/schema/schema.proto
ostafen: The field has been introduced in this PR, I changed ordering in the context of different commits of this MR, so It should not represent a problem

Razikus: sorry, in review window I had only a change with renumbering index, not overall

Probably I had a view of changes after my last review

Razikus (approved):

Razikus (approved):

PR #1961, opened 2024-04-15
Allow order by on non-indexed fields

ostafen:
This merge request is meant to support the order by clause on non indexed fields.
If the query planner determines that the output is not already sorted, then an additional sorting step is added through a new type of reader: `sortReader`.
When the client attempts to read the first row on the sort reader, all rows gets fetched from the underlying reader and collected to a buffer. When the buffer contains at least 1024 rows, sorting is performed in memory and the sorted chunk is spilled to a temporary disk file. The process continues until all the result rows have been read. 
At this point, sorted disk chunks (if any) are merged into a single bigger sorted chunk and a reader to such a file is returned to the client. If the output contains a number of rows <= 1024, then a reader to the buffer is simply returned.

### DIFF
--- a/embedded/sql/options.go
+++ b/embedded/sql/options.go
@@ -22,10 +22,14 @@ import (
 	"github.com/codenotary/immudb/embedded/store"
 )
 
-var defaultDistinctLimit = 1 << 20 // ~ 1mi rows
+const (
+	defaultDistinctLimit  = 1 << 20 // ~ 1mi rows
+	defaultSortBufferSize = 1024
+)
 
 type Options struct {
 	prefix                        []byte
+	sortBufferSize                int
 	distinctLimit                 int
 	autocommit                    bool
 	lazyIndexConstraintValidation bool
@@ -35,7 +39,8 @@ type Options struct {
 
 func DefaultOptions() *Options {
 	return &Options{
-		distinctLimit: defaultDistinctLimit,
+		sortBufferSize: defaultSortBufferSize,
+		distinctLimit:  defaultDistinctLimit,
 	}
 }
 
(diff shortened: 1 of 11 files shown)


embedded/sql/sql_tx.go
```diff
@@ -130,6 +136,11 @@ func (sqlTx *SQLTx) Commit(ctx context.Context) error {
 		return err
 	}
 
+	err = sqlTx.removeTempFiles()
```
Razikus, line 139: we have deffered removal of temp files, is that explicit call needed?

embedded/sql/options.go
```diff
@@ -75,3 +80,8 @@ func (opts *Options) WithMultiDBHandler(multidbHandler MultiDBHandler) *Options
 	opts.multidbHandler = multidbHandler
 	return opts
 }
+
+func (opts *Options) WithSortBufferSize(size int) *Options {
+	opts.sortBufferSize = size
```
Razikus, line 85: we should document it - what it is, what will increase of this value provide, what if we will set to 0

embedded/sql/stmt.go
```diff
@@ -2654,16 +2630,19 @@ func (stmt *SelectStmt) genScanSpecs(tx *SQLTx, params map[string]interface{}) (
...
 					break
 				}
 			}
 		}
-
-		descOrder = stmt.orderBy[0].descOrder
 	}
 
 	if sortingIndex == nil {
-		return nil, ErrNoAvailableIndex
```
Razikus, line 2666: this could potentially affects current users code, should be communicated somehow

Razikus (commented):
do we have some small benchmark for this PR? just to know if there are implications for performance

Razikus (dismissed):

embedded/sql/options.go
ostafen: Added validity check and description

Razikus (dismissed):

embedded/sql/stmt.go
ostafen: Updated doc: https://github.com/codenotary/immudb.io/pull/510

Razikus (approved):

PR #1967, opened 2024-04-23
Support streaming response in SQLQuery/TxSQLQuery rpcs

ostafen:
This PR makes methods `SQLQuery/TxSQLQuery` return a streaming response , thus removing the present limitation on the maximum number of rows returned by a query.
From the postgres server side, streaming has been implemented by splittings the result in multiple messages, according to the [pg protocol spec](https://www.postgresql.org/docs/8.1/protocol-flow.html#:~:text=A%20simple%20query,later%20commands%20succeed.)).

To make this change backward compatible and ease the transition of existing grpc clients, a field has been added to the `SQLQueryRequest` message: `acceptStream` (such field can be deprecated and finally removed when all clients will be migrated).
Such field is used by the server to establish if streaming should be used or if an error should be returned to the client when the number of rows to return exceeds the limit. 
Since existing clients are not aware of the new field, it will default to `false` in their requests, so the behavior will be exactly the same as before.

Nothing is required for pg clients, since they are already designed to handle streaming.

NOTE: The method has been checked with the following SDK: `Python`, `Java` and `NodeJS`.


Makefile
```diff
@@ -170,10 +170,10 @@ build/codegen:
 	  -I$(GOPATH)/pkg/mod/github.com/grpc-ecosystem/grpc-gateway@$(GRPC_GATEWAY_VERSION) \
 	  -I$(GOPATH)/pkg/mod/github.com/grpc-ecosystem/grpc-gateway@$(GRPC_GATEWAY_VERSION)/third_party/googleapis \
 	  --go_out=paths=source_relative:pkg/api/schema \
-	  --go-grpc_out=require_unimplemented_servers=false,paths=source_relative:pkg/api/schema
-      --grpc-gateway_out=logtostderr=true,paths=source_relative:pkg/api/schema
-	  --doc_out=pkg/api/schema --doc_opt=markdown,docs.md
-	  --swagger_out=logtostderr=true:pkg/api/schema
+	  --go-grpc_out=require_unimplemented_servers=false,paths=source_relative:pkg/api/schema \
```
ostafen, line 173: Sounds like this command was executed without invoking grpc-gateway without the backslashes

pkg/api/schema/schema.proto
```diff
@@ -1639,7 +1642,7 @@ service ImmuService {
 
   rpc Delete(DeleteKeysRequest) returns (TxHeader) {
     option (google.api.http) = {
-      post: "/db/delete"
+      post: "/db/deletekey"
```
ostafen, line 1645: This endpoint was clashing with the one generated for the `DeleteDatabase()` rpc. 
Because of the fact that grpc-gateway was not invoked (see above), this didn't result in any error.

Razikus (dismissed):

PR #2002, opened 2024-07-16
PR to handle the AWS v4 signing by using IAM Role

lorepas:
Fixed a bug encountered when using assume role to connect immudb to an S3 bucket. In this PR I added the manage of the SessionToken and I used the `X-Amz-Security-Token` header as specified here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html

### DIFF
--- a/embedded/remotestorage/s3/s3.go
+++ b/embedded/remotestorage/s3/s3.go
@@ -52,7 +52,8 @@ type Storage struct {
 	prefix        string
 	location      string
 	httpClient    *http.Client
-
+	sessionToken  string
+	
 	awsInstanceMetadataURL string
 	awsCredsRefreshPeriod  time.Duration
 }
@@ -214,11 +215,11 @@ func (s *Storage) s3SignedRequestV4(
 
 	req, err := http.NewRequestWithContext(ctx, method, reqUrl, body)
 	if err != nil {
-		return nil, err
+			return nil, err
 	}
 	err = setupRequest(req)
 	if err != nil {
-		return nil, err
+			return nil, err
 	}
 
 	timeISO8601 := t.Format("20060102T150405Z")
@@ -227,13 +228,14 @@ func (s *Storage) s3SignedRequestV4(
 	credential := s.accessKeyID + "/" + scope
 
 	if contentSha256 == "" {
-		contentSha256 = unsignedPayload
+			contentSha256 = unsignedPayload
 	}
 
 	req.Header.Set("X-Amz-Date", timeISO8601)
 	req.Header.Set("X-Amz-Content-Sha256", contentSha256)
+	req.Header.Set("X-Amz-Security-Token", s.sessionToken)
 	if contentType != "" {
-		req.Header.Set("Content-Type", contentType)
+			req.Header.Set("Content-Type", contentType)
 	}
 
 	canonicalURI := req.URL.Path // TODO: This may require some encoding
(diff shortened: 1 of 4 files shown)


lorepas:
@jeroiraz have you take a look into this fix?

ostafen:
Hi, @lorepas, first of all thanks for the fix. Could you please align your branch with immudb's latest master commit?
Also, I've seen that your PR contains a lot of extra tabs (maybe due to your formatter config) that make difficult to isolate the exact changes of the PR. Thanks!

embedded/remotestorage/s3/s3.go
```diff
@@ -732,39 +735,39 @@ func (s *Storage) getRoleCredentials() error {
...
 	tokenReq, err := http.NewRequest("PUT", fmt.Sprintf("%s%s",
 		s.awsInstanceMetadataURL,
 		"/latest/api/token",
 	), nil)
 	if err != nil {
-		return "", "", errors.New("cannot form metadata token request")
+		return "", "", "", errors.New("cannot form metadata token request")
 	}
 
 	tokenReq.Header.Set("X-aws-ec2-metadata-token-ttl-seconds", "21600")
```
ostafen, line 759: I think we should make this configurable or at least move it to a `const`

lorepas:
> Hi, @lorepas, first of all thanks for the fix. Could you please align your branch with immudb's latest master commit? Also, I've seen that your PR contains a lot of extra tabs (maybe due to your formatter config) that make difficult to isolate the exact changes of the PR. Thanks!

Hi @ostafen and thank you for your reply! Now I refactored the code, so the changes you can see now are the ones made for the fix. For these changes I started from the latest opensource version, the 1.9DOM.2, for this reason I would put these changes into such version. Do you think is that possible?

ostafen:
@lorepas: latest version is now `1.9.4`. Any specific reason you are using `1.9DOM.2`?

lorepas:
> @lorepas: latest version is now `1.9.4`. Any specific reason you are using `1.9DOM.2`?

@ostafen because from the change of license made by immudb I understand that the latest opensource was 1.9DOM.2. All the ones after such version are licensed BSL. Is that right?

embedded/remotestorage/s3/s3.go
lorepas: Actually, in this case the access_key, secret_key and session_token are taken from the `GET /latest/api/token` made in that function. And such credentials are temporary, so I think that for the role we can't configure the session_token and it can't be a const.

ostafen: @lorepas: I'm talking about `X-aws-ec2-metadata-token-ttl-seconds`

ostafen:
@lorepas
The license change doesn't change the fact that it's open source or the project cannot be used by anyone. The only real restriction affects the commercial usage in a very limited range.
Thus, allowing PRs to outdated versions wouldn't make much sense and would confuse users and developers.

lorepas:
> @lorepas The license change doesn't change the fact that it's open source or the project cannot be used by anyone. The only real restriction affects the commercial usage in a very limited range. Thus, allowing PRs to outdated versions wouldn't make much sense and would confuse users and developers.

@ostafen yes, it's clear since the code is public all can contribute on that and for this reason it's opensource. I changed target branch to the master, however I've one last question: which are the limits on production use case about the use of immudb with BSL?

ostafen:
@lorepas: can you solve conflicts?

lorepas:
@ostafen conflicts solved!

ostafen:
It looks like code scanning complains about logging `storage` [here](https://github.com/codenotary/immudb/security/code-scanning/29), since it could reveal `secretKey`.
@lorepas, would you mind changing such line to:
```golang
log.Printf("Opening remote storage at %s", remotePath)
```

lorepas:
@ostafen I changed the line you suggest

ostafen:
Some related test is failing:

image

ostafen:
@lorepas: please, make sure all tests are passing: `TestSignatureV4` is still failing.

lorepas:
@ostafen Now I set the header x-amz-security-token only if role is enabled. I fixed also the test by eliminating such header since the test doesn't perform temporary authentication. I think now the test should work.

ostafen:
@lorepas: Thanks for fixing the test. Before merging, could you please squash your commits into one so to keep history clean?

lorepas:
@ostafen I tried to squash my commits but currently they increase in number by taking all the ones from my first and now. Can you do the squash while you're accepting the PR? Otherwise I accept hint to understand how to do such operation.

ostafen (approved):

ostafen:
@lorepas: I cannot merge the PR, because not all the commits are signed.
Regarding squashing your commits, read [this](https://stackoverflow.com/questions/30136558/how-to-squash-commits-which-have-merge-commit-in-between#:~:text=Assuming%20the%20feature%20branch%20is%20called%20feature%20and%20the%20main%20branch%20main%3A).

PR #2005, opened 2024-07-22
Prohibite changing transiency of keys

ostafen:


Razikus (approved):

felipelalli:
I couldn't understand exactly this correction. Could you explain it to me in more detail, please?

ostafen:
@felipelalli, the issue happened when trying to set a transaction key using `Set` which was previously set using `SetTransient`.

PR #2010, opened 2024-08-07
Fix AWS Signature V4 using IAM role

lorepas:
This PR is the same as the one discussed on #2002 . Only difference is that I used one signed commit.


lorepas:
@ostafen I created a new clean PR with the same changes discussed on #2002

ostafen (approved):

ostafen:
@lorepas: thanks for your effort! I merged the PR.

PR #2014, opened 2024-08-20
fix: time.Since should not be used in defer statement

druzn3k:
This commit fixes the complaining from `go vet` about the deferring of `time.Since`. This can cause wrong time evaluation of the elapsed time. This impact only the debug level, but still it should be addressed.

ref: https://github.com/golang/go/issues/60048

Signed-off-by: Maurizio Del Corno <druznek@gmail.com>

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -637,6 +637,10 @@ func NewImmuClient(options *Options) (*immuClient, error) {
 	return c, nil
 }
 
+func (c *immuClient) debugElapsedTime(method string, start time.Time) {
+	c.Logger.Debugf("method immuclient.%s took %s", method, time.Since(start))
+}
+
 func (c *immuClient) getServerIdentity() string {
 	// TODO: Allow customizing this value
 	return c.Options.Bind()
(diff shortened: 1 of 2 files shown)


ostafen:
@druzn3k: maybe adding a `debug` method to immuClient to wrap the call could simplify the code

druzn3k:
> @druzn3k: maybe adding a `debug` method to immuClient to wrap the call could simplify the code 

Why not :) I'm thinking about something in this ballpark:

```go
func (c *immuClient) debugWithTimeElapsed(msg string, start time.Time)
```

The name is on the ugly side though, but I can't come up with something better on the spot. Suggestions?

ostafen:
maybe:

```golang
func (c *immuClient) debugElapsedTime(method string, start *time.Time)
```
You could just pass the method name

druzn3k:
done @ostafen . Let me know if you like this version better.

ostafen (approved):

ostafen:
@druzn3k: looks good to me, thanks!

PR #2025, opened 2024-09-16
fix(pkg/dabase): return error when attempting to access deleted database

ostafen:


Razikus (approved):

pkg/database/db_manager.go
```diff
@@ -266,10 +266,15 @@ func (m *DBManager) Has(name string) bool {
...
+	db, exists := m.getDB(idx)
+	if !exists {
+		return false
+	}
+
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
 
-	return idx >= 0 && idx < len(m.databases)
+	return !db.deleted
```
Razikus, line 277: Shouldn't be here also check for idx?

Razikus (approved):

pkg/database/db_manager.go
ostafen: Alredy done inside function `db, exists := m.getDB(idx)`. It returns `false` if `idx` is not inside the range

PR #2109, opened 2026-07-07
fix(database): guard nil db in DBManager.CloseAll to avoid shutdown panic

### LINKED ISSUE #2108: panic: nil pointer dereference in DBManager.CloseAll on shutdown when a database failed to open
## What happened

On graceful shutdown (SIGTERM), immudb panics with a nil pointer dereference inside
`DBManager.CloseAll`, aborting `Stop()` before it completes. The process dies without
logging `shutdown completed`, so the shutdown sequence (final DB flush) is truncated.

This is reproducible whenever at least one database **failed to open** during the
lifetime of the server (in our case a replica database that could not fetch state from
its primary), and a shutdown is then triggered.

## Version

- immudb **1.9.7**, commit `6ac516b2236607f3afdfbd1cff7d0cb86f1cf1a9` (Built 2025-05-30)
- Deployment: official `codenotary/immudb:1.9.7` image on AWS ECS, S3 remote storage,
  primary + 2 replicas.
- The faulty code is **unchanged in the latest release (v1.11.1) and in `master`**
  (`pkg/database/db_manager.go` is byte-identical), so this affects current versions too.

## Stack trace

```
caught SIGTERM
stopping immudb: ...
... (replication stopped, defaultdb closed successfully) ...
database 'defaultdb' successfully closed
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0xbc186d]

goroutine 482 [running]:
github.com/codenotary/immudb/pkg/database.(*DBManager).CloseAll.func1({...}, {...})
        /src/pkg/database/db_manager.go:467
github.com/codenotary/immudb/embedded/cache.(*Cache).Apply(...)
        /src/embedded/cache/cache.go:301
github.com/codenotary/immudb/pkg/database.(*DBManager).CloseAll(...)
        /src/pkg/database/db_manager.go:459
github.com/codenotary/immudb/pkg/database.(*databaseList).CloseAll(...)
        /src/pkg/database/types.go:97
github.com/codenotary/immudb/pkg/server.(*ImmuServer).CloseDatabases(...)
        /src/pkg/server/server.go:778
github.com/codenotary/immudb/pkg/server.(*ImmuServer).Stop(...)
        /src/pkg/server/server.go:770
github.com/codenotary/immudb/pkg/server.(*ImmuServer).installShutdownHandler.func1()
        /src/pkg/server/server.go:917
```

## Root cause

In `DBManager.CloseAll` (`pkg/database/db_manager.go`), the closure passed to
`dbCache.Apply` calls `ref.db.Close()` **without checking whether `ref.db` is nil**:

```go
m.dbCache.Apply(func(_, value interface{}) error {
    ref := value.(*dbRef)

    if atomic.LoadUint32(&ref.count) > 0 {
        busyDBs++
        return nil
    }

    ref.db.Close()   // <-- panics if ref.db == nil
    return nil
})
```

This is the **only** close site in the file that does not guard `ref.db != nil`.
Every other site does:

- `createCache` / `OnEvict`: `if ref.db == nil { ...; return }`
- `Close`: `if ref != nil && ref.db != nil { ref.db.Close() }`
- `Release`: `if ref.db != nil { ref.db.Close() }`

A `dbRef` with `db == nil` is left in the cache by `Get` when `openDB` fails:

```go
func (m *DBManager) Get(idx int) (DB, error) {
    ...
    ref, err := m.allocDB(idx, db)   // Put()s a &dbRef{count: 1} (db is nil)
    ...
    if ref.db == nil {
        d, err := m.openDB(db.name, db.opts)
        if err != nil {
            m.Release(idx)           // count -> 0, but the ref stays in the cache
            return nil, err
        }
        ref.db = d
    }
    ...
}
```

`Release` only decrements the ref count; it never removes the entry from `dbCache`.
So after a failed open, the cache holds a `dbRef{count: 0, db: nil}`. On shutdown,
`CloseAll` iterates it (`count == 0`), calls `ref.db.Close()`, and panics.

This matches the observed logs: databases that opened successfully close cleanly, then
the panic hits on the database whose open had failed.

## How to reproduce

1. Configure a replica database whose primary state cannot be fetched (so `openDB`
   returns an error) — e.g. a replica pointing at a primary that does not have the
   database, or missing replication credentials.
2. Let the server run so `Get`/`openDB` fails for that database at least once, leaving a
   `dbRef{db: nil}` in the cache.
3. Send SIGTERM.
4. → panic in `CloseAll` instead of a clean `shutdown completed`.

## Impact

- Graceful shutdown never completes; the tx-log flush is truncated.
- On a **replica** this is recoverable (it re-syncs from the primary), but on a
  **primary** the truncated flush to remote (S3) storage can corrupt the transaction
  log on the next start.

## Suggested fix

Guard the nil case in `CloseAll`, mirroring the other close sites:

```go
m.dbCache.Apply(func(_, value interface{}) error {
    ref, _ := value.(*dbRef)
    if ref == nil {
        return nil
    }

    if atomic.LoadUint32(&ref.count) > 0 {
        busyDBs++
        return nil
    }

    if ref.db != nil {
        ref.db.Close()
    }
    return nil
})
```

Happy to open a PR with this change + a unit test covering the failed-open-then-shutdown
path if that helps.

Akhoris:
## What

Guard `ref.db != nil` in `DBManager.CloseAll` (`pkg/database/db_manager.go`) so a
db-less cache ref no longer causes a nil pointer dereference on shutdown.

Fixes #2108.

## Why

On `SIGTERM`, `Stop()` → `CloseDatabases()` → `DBManager.CloseAll` can panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38]
github.com/codenotary/immudb/pkg/database.(*DBManager).CloseAll.func1
        pkg/database/db_manager.go:467
...
github.com/codenotary/immudb/pkg/server.(*ImmuServer).Stop
github.com/codenotary/immudb/pkg/server.(*ImmuServer).installShutdownHandler.func1
```

The `dbCache.Apply` closure in `CloseAll` called `ref.db.Close()` without checking
`ref.db != nil`. It is the only close site in the file that didn't guard — `OnEvict`,
`Close` and `Release` all check `ref.db != nil`.

A db-less ref ends up in the cache whenever a database **fails to open**: `Get()` has
`allocDB` `Put()` a `&dbRef{count: 1}` (its `db` is still nil), then `openDB` fails and
`Get()` calls `Release()`, which only decrements the ref count — it never removes the
entry from the cache. On shutdown `CloseAll` iterates that `{count: 0, db: nil}` ref and
dereferences the nil `db`.

Observed in production on a replica whose replicated database could not fetch state from
its primary (`openDB` kept failing): successfully-opened databases closed cleanly, then
the panic hit on the failed one, so `Stop()` never completed and the graceful shutdown /
final flush was truncated.

## How

```go
m.dbCache.Apply(func(_, value interface{}) error {
    ref, _ := value.(*dbRef)
    if ref == nil {
        return nil
    }

    if atomic.LoadUint32(&ref.count) > 0 {
        busyDBs++
        return nil
    }

    // ref.db may be nil if the database was never successfully opened.
    if ref.db != nil {
        ref.db.Close()
    }
    return nil
})
```

## Test

Added `TestDBManagerCloseAllAfterFailedOpen`, which drives the failed-open-then-shutdown
path. It panics on the current code and passes with the fix:

- before: `--- FAIL … Panic value: runtime error: invalid memory address or nil pointer dereference`
- after: `ok  github.com/codenotary/immudb/pkg/database`

`go test ./pkg/database/ -run TestDBManager` passes.

### DIFF
--- a/pkg/database/db_manager.go
+++ b/pkg/database/db_manager.go
@@ -485,14 +485,22 @@ func (m *DBManager) CloseAll(ctx context.Context) error {
 
 		busyDBs := 0
 		m.dbCache.Apply(func(_, value interface{}) error {
-			ref := value.(*dbRef)
+			ref, _ := value.(*dbRef)
+			if ref == nil {
+				return nil
+			}
 
 			if atomic.LoadUint32(&ref.count) > 0 {
 				busyDBs++
 				return nil
 			}
 
-			ref.db.Close()
+			// ref.db may be nil if the database was never successfully opened
+			// (e.g. openDB failed in Get and left a db-less ref in the cache).
+			// Guard here as every other close site in this file does.
+			if ref.db != nil {
+				ref.db.Close()
+			}
 			return nil
 		})
 		tryClose = busyDBs > 0
--- a/pkg/database/db_manager_test.go
+++ b/pkg/database/db_manager_test.go
@@ -491,3 +491,32 @@ func TestDBManagerPutUpdate(t *testing.T) {
 	require.Equal(t, idx1, idx2, "re-putting same name should return same index")
 	require.Equal(t, 1, manager.Length(), "should not create duplicate entry")
 }
+
+func TestDBManagerCloseAllAfterFailedOpen(t *testing.T) {
+	// A database whose open fails leaves a db-less ref in the cache: Get() has
+	// allocDB Put() a &dbRef{count: 1}, openDB then fails and Release() only
+	// decrements the count without removing the entry. CloseAll must skip such a
+	// ref instead of dereferencing its nil db.
+	openErr := fmt.Errorf("open failed")
+	openDB := func(name string, opts *Options) (DB, error) {
+		return nil, openErr
+	}
+
+	manager := NewDBManager(openDB, 5, logger.NewMemoryLogger())
+	manager.Put("testdb", DefaultOptions(), false)
+
+	_, err := manager.Get(0)
+	require.ErrorIs(t, err, openErr)
+
+	// The failed open must have left a ref with a nil db in the cache.
+	v, err := manager.dbCache.Get(0)
+	require.NoError(t, err)
+	ref, _ := v.(*dbRef)
+	require.NotNil(t, ref)
+	require.Nil(t, ref.db)
+	require.Zero(t, atomic.LoadUint32(&ref.count))
+
+	require.NotPanics(t, func() {
+		require.NoError(t, manager.CloseAll(context.Background()))
+	})
+}


vchaindz:
Thanks a lot for the thorough write-up and the fix, @Akhoris — the root-cause analysis was spot on and the regression test is much appreciated. Merging now. 🙏

vchaindz (approved):
Reviewed the fix and the regression test — correct, minimal, and consistent with the nil-guards at every other close site in the file. LGTM.

vchaindz:
One last thing before this goes in: `master` has "require signed commits" enabled, and this commit is unsigned, so the merge is blocked on that. Could you re-push it signed? Something like:

```
git commit --amend -S --no-edit
git push --force-with-lease
```

(assuming your signing key is set up via `git config user.signingkey` / `commit.gpgsign`). Once it shows as "Verified" I'll merge it right away. Thanks again!

Akhoris:
Done, thanks @vchaindz! 🙏

I re-signed the commit with GPG and force-pushed — it now shows as **Verified** (`a13fcad`). The DCO sign-off is preserved and the diff is unchanged from what you reviewed; only the signature was added.

Should be unblocked on the signed-commits requirement now. Let me know if the workflows need another "Approve and run" from your side after the force-push, or if there's anything else you need from me.

vchaindz:
Signed commit came through and verified — merged into master. Thanks again for the clean fix and the regression test, @Akhoris! 🎉

alexandru-tone:
Hi,
Would it be possbile for a v1.11.2 tag to include this PR?
Thank you!

vchaindz:
Hi @alexandru-tone we had some other open tasks, but the PR is now included in https://github.com/codenotary/immudb/releases/tag/v1.11.2

Issue #2100, opened 2026-06-25
Title: immudb 1.11.0 SQL engine panic (nil pointer dereference / index out of range) on valid queries


7070500:
**What happened**
After upgrading from immudb 1.10.0 to 1.11.0, valid SQL queries started causing immudb to panic.
Downgrading back to 1.10.0 immidiately resolves the issue without any code or data changes.

Panic 1:
Panic: runtime error: invalid memory address or nil pointer dereference

Stack trace:
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0xb4c540]
goroutine 80357 [running]:
github.com/codenotary/immudb/embedded/sql.encodeRow(0xc0511bf460)
/src/embedded/sql/file_sort.go:455 +0xa0
github.com/codenotary/immudb/embedded/sql.(*fileSorter).flushBuffer(0xc035569c40)
/src/embedded/sql/file_sort.go:415 +0xa8
github.com/codenotary/immudb/embedded/sql.(*fileSorter).sortAndFlushBuffer(0xc035569c40)
/src/embedded/sql/file_sort.go:387 +0x2d
github.com/codenotary/immudb/embedded/sql.(*fileSorter).update(0xc035569c40, 0xc032da41c0)
/src/embedded/sql/file_sort.go:60 +0x2b
github.com/codenotary/immudb/embedded/sql.(*sortRowReader).readAll(0xc035569c00, {0x2ba91c8, 0xc03a79fda0})
/src/embedded/sql/sort_reader.go:397 +0x99
github.com/codenotary/immudb/embedded/sql.(*sortRowReader).readAndSort(0xc035569c00, {0x2ba91c8?, 0xc03a79fda0?})
/src/embedded/sql/sort_reader.go:298 +0x2c
github.com/codenotary/immudb/embedded/sql.(*sortRowReader).Read(0xc035569c00, {0x2ba91c8?, 0xc03a79fda0?})
/src/embedded/sql/sort_reader.go:285 +0x30
github.com/codenotary/immudb/embedded/sql.(*projectedRowReader).Read(0xc03ec3cf00, {0x2ba91c8?, 0xc03a79fda0?})
/src/embedded/sql/proj_row_reader.go:186 +0x42
github.com/codenotary/immudb/embedded/sql.(*offsetRowReader).Read(0xc0511bf1c0, {0x2ba91c8, 0xc03a79fda0})
[...]

PR #2110, opened 2026-07-12
fix(pgsql): guard extended-query Parse against zero parsed statements + nil row-value-slot hardening

vchaindz:
## Summary

Follow-up to #2100: a third panic of the same regression family was reported [in a comment](https://github.com/codenotary/immudb/issues/2100#issuecomment-4950023827) after the original fix was merged — this time in the pgsql wire-protocol server:

```
panic: runtime error: index out of range [0] with length 0
github.com/codenotary/immudb/pkg/pgsql/server.(*session).QueryMachine
    pkg/pgsql/server/query_machine.go:177
```

### Commit 1 — `fix(pgsql)`: extended-query Parse panic

Since 1.11.0 the SQL grammar accepts empty/comment-only input, so `ParseSQL` returns an empty statement slice with no error. The extended-query `Parse` handler guarded `len(stmts) > 1` but then indexed `stmts[0]` unguarded — any Parse message carrying such SQL (e.g. a driver liveness probe like `-- ping`) crashed the whole server.

Parse of an empty query string is valid in the extended protocol: the handler now skips parameter/result-column inference and registers the statement; a later Execute re-parses it and answers with `EmptyQueryResponse`, matching PostgreSQL behavior.

### Commit 2 — `chore(sql)`: nil row-value-slot hardening

A repo-wide audit for both #2100 bug families found:

- **Empty-statement-slice class: fully closed.** Every `ParseSQL`/`ParseSQLString` call site is now length-guarded, range-based, or parses hardcoded non-empty input. The Parse handler fixed here was the last reachable one.
- **Nil `ValuesByPosition` slot class (projection pushdown): no live panic left**, but five consumers dereference every positional slot and stay safe only because projection currently repopulates rows before they run — the same fragile invariant whose violation caused the `fileSorter.encodeRow` panic. Guards added (nil slot ⇒ SQL NULL, or `ErrUnexpectedValue` for the document readers where the slots are always explicitly projected):
  - CTE materialization (`materializeCTE`/`materializeRecursiveCTE`) — was a hard type-assertion panic on a nil interface
  - `Row.digest` (DISTINCT, diff)
  - `setOpRowReader.rowDigest` (EXCEPT/INTERSECT)
  - `sqlRowsToProto` (gRPC row serialization)
  - `embedded/document` readers (`_id`/`_doc` extraction, count)

## Tests

- Unit regression test in `TestSession_QueriesMachine` reproducing the exact panic at `query_machine.go:177` (fails with the guard reverted).
- End-to-end test driving a real server through lib/pq's extended protocol (`Prepare("-- ping")` → Parse/Describe/Sync → Execute), asserting the session survives and stays usable.
- Unit tests encoding the nil-slot-equals-NULL contract for `Row.digest`, set-op digest, CTE materialization, and gRPC serialization.

`go build ./...`, `go vet`, and full test suites for `embedded/sql`, `embedded/document`, `pkg/pgsql/...`, `pkg/database`, and `pkg/server` all pass.

Fixes the follow-up panic reported in #2100.

### DIFF
--- a/pkg/server/sql.go
+++ b/pkg/server/sql.go
@@ -200,8 +200,9 @@ func sqlRowsToProto(descriptors []sql.ColDescriptor, rows []*sql.Row, outRows []
 			row.Columns[i] = descriptors[i].Selector()
 
 			v := sqlRow.ValuesByPosition[i]
+			// a nil slot (column skipped by projection pushdown) serializes as NULL
 			_, isNull := v.(*sql.NullValue)
-			if isNull {
+			if v == nil || isNull {
 				row.Values[i] = &schema.SQLValue{Value: &schema.SQLValue_Null{}}
 			} else {
 				row.Values[i] = schema.TypedValueToRowValue(v)
--- a/embedded/document/document_reader.go
+++ b/embedded/document/document_reader.go
@@ -67,8 +67,10 @@ func (r *documentReader) ReadN(ctx context.Context, count int) ([]*protomodel.Do
 			return nil, mayTranslateError(err)
 		}
 
-		docIDBytes := row.ValuesByPosition[0].RawValue().([]byte)
-		docBytes := row.ValuesByPosition[1].RawValue().([]byte)
+		docIDBytes, docBytes, err := docRowBytes(row)
+		if err != nil {
+			return nil, err
+		}
 
 		doc := &structpb.Struct{}
 		err = proto.Unmarshal(docBytes, doc)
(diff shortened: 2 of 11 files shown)


pkg/server/sql.go
```diff
@@ -200,8 +200,9 @@ func sqlRowsToProto(descriptors []sql.ColDescriptor, rows []*sql.Row, outRows []
 			row.Columns[i] = descriptors[i].Selector()
 
 			v := sqlRow.ValuesByPosition[i]
+			// a nil slot (column skipped by projection pushdown) serializes as NULL
 			_, isNull := v.(*sql.NullValue)
-			if isNull {
+			if v == nil || isNull {
```
SimoneLazzaris, line 205: I'm not a fan of checking `v == nil` *after* casting it, but it works

SimoneLazzaris (approved):

PR #2111, opened 2026-07-15
chore(deps): bump x/crypto, x/net and grpc in linear-fake examples

vchaindz:
## Summary

Clears all open Dependabot alerts on `master`, which point at the two example modules under `docs/security/vulnerabilities/linear-fake/` (the root `go.mod` is already patched):

- **golang.org/x/crypto** v0.48.0 → v0.52.0 in both modules — clears 13 CVEs, including the critical SSH issues (CVE-2026-42508 auth bypass via unenforced `@revoked`, CVE-2026-46595 `VerifiedPublicKeyCallback` permission skip, CVE-2026-39831 FIDO/U2F presence-check bypass, CVE-2026-39832/39833 key-constraint enforcement, CVE-2026-39830/39834 DoS).
- **golang.org/x/net** v0.50.0 → v0.55.0 in both modules — CVE-2026-25680 (HTML parser DoS).
- **google.golang.org/grpc** v1.58.3 → v1.79.3 in the server example — CVE-2026-33186 (`:path` authorization bypass), plus matching `genproto`/`protobuf` updates.

## Build fix

The server example did not compile even before the bumps: `dbWrapper` in `server.go` had drifted from the `database.DB` interface of the pinned immudb v1.4.1 (masked by an unrelated protobuf build error with the old dependency set). This PR aligns the two mismatched stubs:

- `ExportTxByID` now uses the four-value return signature
- `WaitForTx` gains the `allowPrecommitted bool` parameter

## Verification

- `go build ./...` and `go vet ./...` pass in both `linear-fake/go` and `linear-fake/server`
- Both go.mod files confirmed at crypto ≥ 0.52.0, net ≥ 0.55.0, grpc ≥ 1.79.3

The four remaining alerts referencing the removed `tools/stream` module were dismissed as `not_used`; the rest should auto-resolve once this merges.


SimoneLazzaris (approved):
LGTM

